Issue #3, opened 2018-01-13
`make coverage` segfaults on Linux


dfellis:
Hey @isaacbrodsky :) The `testCompact` test in code coverage mode currently segfaults in Linux (Kubuntu 17.04) on both gcc and clang.

I'll try to figure it out and fix it if I can, but you'll probably run into the same thing when you're getting coveralls set up, so I thought I'd give you a heads up.

```
damocles@Talyn:~/oss/h3(master)$ make coverage
[  1%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/algos.c.o
[  1%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/baseCells.c.o
[  1%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/bbox.c.o
[  2%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/coordijk.c.o
[  2%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/faceijk.c.o
[  2%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/geoCoord.c.o
[  3%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/h3Index.c.o
[  3%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/h3IndexFat.c.o
[  3%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/h3UniEdge.c.o
[  5%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/h3api.c.o
[  5%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/linkedGeo.c.o
[  5%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/mathExtensions.c.o
[  6%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/vec2d.c.o
[  6%] Building C object CMakeFiles/h3.1.dir/src/h3lib/lib/vertexGraph.c.o
[  6%] Linking C shared library lib/libh3.1.so
[  6%] Built target h3.1
[  6%] Building C object CMakeFiles/testH3ToGeoBoundary.dir/src/apps/testapps/testH3ToGeoBoundary.c.o
[  7%] Building C object CMakeFiles/testH3ToGeoBoundary.dir/src/apps/applib/lib/kml.c.o
[...]

PR #4, opened 2018-01-13
Fix typo in compact test

dfellis:
This resolves issue #3 

The correct amount of memory was not being allocated for the uncompacted hexagons, so the test was writing to memory it shouldn't have been, and was caught on my laptop only when checking code coverage. I'm not sure why, though. :man_shrugging:

### DIFF
--- a/src/apps/testapps/testCompact.c
+++ b/src/apps/testapps/testCompact.c
@@ -85,7 +85,7 @@ TEST(res0) {
     }
 
     H3Index* decompressed = calloc(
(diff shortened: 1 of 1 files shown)


isaacbrodsky:
Thanks for fixing this. I assume the `calloc` call failed due to trying to allocate gigabytes of memory.

dfellis:
@isaacbrodsky you'll have to merge the change in, I don't have permission to do so. :)

PR #8, opened 2018-01-18
Remove all (direct) usage of H3IndexFat from the algos package

dfellis:
Hi @isaacbrodsky @nrabinowitz @joegilley and friends,

Since we weren't able to remove the H3IndexFat struct from the codebase before it became open sourced, I'm going to help get us there now that it is. :)

This diff removes all direct usage of H3IndexFat from the `algos.c` file. Some of the functions it calls from other files (especially `h3api.c`) still use it, but direct usage now calls the equivalent H3Index functions (a few of which I wrote).

The performance appears to be slightly better (approximately 6-7%) now than it was before, so that's also a plus (and a nice indicator that our hunch was correct, eliminating H3IndexFat should reduce memory usage and more importantly memory allocation during tight loops on the core algorithm like in the polyfill function).

```
damocles@Talyn:~/oss/h3(slim-algos)$ make benchmarks
[  0%] Formatting sources
[  0%] Built target format
[ 62%] Built target h3.1
[ 75%] Built target benchmarkH3Api
Scanning dependencies of target bench_benchmarkH3Api
	-- geoToH3: 0.004300 milliseconds per iteration (10000 iterations)
	-- h3ToGeo: 0.001100 milliseconds per iteration (10000 iterations)
	-- h3ToGeoBoundary: 0.004100 milliseconds per iteration (10000 iterations)
[ 75%] Built target bench_benchmarkH3Api
[100%] Built target benchmarkPolyfill
Scanning dependencies of target bench_benchmarkPolyfill
	-- polyfillSF: 4.400000 milliseconds per iteration (500 iterations)
	-- polyfillAlameda: 7.430000 milliseconds per iteration (500 iterations)
	-- polyfillSouthernExpansion: 328.200000 milliseconds per iteration (10 iterations)
[100%] Built target bench_benchmarkPolyfill
Scanning dependencies of target benchmarks
[100%] Built target benchmarks
damocles@Talyn:~/oss/h3(slim-algos)$ git checkout master
Switched to branch 'master'
damocles@Talyn:~/oss/h3(master)$ make clean && make > /dev/null
damocles@Talyn:~/oss/h3(master)$ make benchmarks
[  0%] Formatting sources
[  0%] Built target format
[ 62%] Built target h3.1
[ 75%] Built target benchmarkH3Api
	-- geoToH3: 0.004300 milliseconds per iteration (10000 iterations)
	-- h3ToGeo: 0.001100 milliseconds per iteration (10000 iterations)
	-- h3ToGeoBoundary: 0.003900 milliseconds per iteration (10000 iterations)
[ 75%] Built target bench_benchmarkH3Api
[100%] Built target benchmarkPolyfill
	-- polyfillSF: 4.630000 milliseconds per iteration (500 iterations)
	-- polyfillAlameda: 7.918000 milliseconds per iteration (500 iterations)
	-- polyfillSouthernExpansion: 351.500000 milliseconds per iteration (10 iterations)
[100%] Built target bench_benchmarkPolyfill
[100%] Built target benchmarks
```

Code coverage is still at 100%. Also I added extra flags to the debug build to include macro expansion support to gdb when debugging what's going on (I got tripped up hard by the 1-indexing `H3_GET/SET_INDEX_DIGIT` versus H3IndexFat's 0-indexing and it took a while to figure that out).


dfellis:
Oh, and I can't wait to put up the diff that actually removes H3IndexFat :D (code contribution net negative, here I come :) )

dfellis:
Ugh. The `clang-format` on my laptop *introduced* that change. Will change it back and see if I just have an ancient version of `clang-format`. @isaacbrodsky would you mind letting me know which version you're using?

dfellis:
My clang-format version, btw:

```
damocles@Talyn:~/oss/h3(slim-algos)$ clang-format --version
clang-format version 4.0.1-6 (tags/RELEASE_401/final)
```

dfellis:
Awesome, fix the formatting. :)

src/h3lib/lib/algos.c
```diff
@@ -205,26 +203,27 @@ H3Index h3NeighborRotations(H3Index origin, const CoordIJK* translationIjk,
     int dir = _unitIjkToDigit(&adjustment);
 
     int newRotations = 0;
-    int oldBaseCell = hf.baseCell;
-    int oldLeadingDigit = _leadingNonZeroDigit(&hf);
```
joegilley: i assume these functions will be removed when you follow up with the `H3IndexFat` removal?

src/h3lib/lib/baseCells.c
```diff
@@ -31,14 +31,14 @@ typedef struct {
 /** @brief Neighboring base cell ID in each IJK direction.
  *
  * For each base cell, for each direction, the neighboring base
- * cell ID is given. -1 indicates there is no neighbor in that direction.
+ * cell ID is given. 127 indicates there is no neighbor in that direction.
```
isaacbrodsky: Just wondering, why is this changed from -1 to 127? Also note that this won't match the generateBaseCellNeighbors program.

dfellis: When you bitmask in -1 for the base cell, it doesn't truncate it to the 8 bits we gave it, and it blows away the hex mode and reserved bits in the upper space of the 64-bit integer.

We could make the `H3_SET_BASE_CELL` macro constrain inputs to the 8-bit range (in which case it would set 255 for a -1 input), but that would slow it down when we could just make it not shove bad data in in the first place.

I chose 127 instead of 255 because I noticed that we're actually only using 7 bits of the 8 bits allocated (I think we allocated 8 bits to keep byte offsets simpler, which makes these bitmasking operations use fewer cycles for most things), and this way if we ever wanted to do anything crazy like mark the upper half of the base cell space for hexagons on mars after Mr. Musk gets us there, then we can. :)

Do you want me to update the `generateBaseCellNeighbors` program to output 127 instead of -1? I was under the impression that output still had to be hand corrected. Did that change?

dfellis: Also, only the neighbor rotation logic used this data, so I determined making changes to it to make the code easier to read and faster was fine.

isaacbrodsky: `generateBaseCellNeighbors` does match the contents in baseCells.c exactly, so I think the generator should be updated if we make this change.

One thing I think is nice about -1 is that it's more obviously a sentinel value, while 127 seems like it could be confused with a real base cell ID. If the only use is in `h3NeighborRotations`, would it make sense to simply modify that function to not set an invalid base cell?
```
            int newBaseCell = baseCellNeighbors[oldBaseCell][dir];
            // ...

            if (newBaseCell == -1) {
                // ...
                newBaseCell = // ...;
            }
            H3_SET_BASE_CELL(out, newBaseCell);
```

dfellis: Well, then we introduce runtime latency for no real gain.

I created an `INVALID_BASE_CELL` constant in the codebase, we could use it in this file and fix the generateBaseCellNeighbors to emit that constant and get improved legibility (it is definitely more explicit than -1) with the better perf, too.

isaacbrodsky: I'm unclear on there being a performance difference, since the `if (... == INVALID_BASE_CELL` check needs to happen either way (algos.c line 218 in master). Maybe there's something I'm not noticing?

I do agree it makes sense to reuse the INVALID_BASE_CELL constant here and in the generator.

dfellis: I'm not sure about the performance difference, but there would be more instructions needed with this approach for the extra temporary variable and comparison versus reusing the already existing 64-bit register that `out` is stored in and it may or may not require copying a different register out to L1 cache and back in later.

It's not that big of a deal, but I think using the constant variable name is more legible. :)

So I'll update the `baseCells.c` and `generateBaseCellNeighbors` to use it.

src/apps/miscapps/generateBaseCellNeighbors.c
```diff
@@ -21,7 +21,8 @@
  *  The program generates a table of neighbors of base cells, and the
  *  60 degree ccw rotations into the coordinate system of that base cell.
  *
- *  -1 is generated for neighbors that are deleted (the deleted neighbor
+ *  INVALID_BASE_CELL is generated for neighbors that are deleted (the deleted
+ * neighbor
```
isaacbrodsky: nit: please reflow the comments here

src/apps/miscapps/generateBaseCellNeighbors.c
```diff
@@ -79,6 +80,17 @@ void auditBaseCellNeighbors(int baseCellNeighbors[NUM_BASE_CELLS][7],
     }
 }
 
+char* getBaseCellOutput(baseCell) {
```
isaacbrodsky: Please add an explicit type for `baseCell`

src/apps/miscapps/generateBaseCellNeighbors.c
```diff
@@ -223,11 +237,14 @@ void generate() {
 
     printf("const int baseCellNeighbors[NUM_BASE_CELLS][7] = {\n");
     for (int i = 0; i < NUM_BASE_CELLS; i++) {
-        printf("{%d, %d, %d, %d, %d, %d, %d}, // base cell %d%s\n",
-               baseCellNeighbors[i][0], baseCellNeighbors[i][1],
-               baseCellNeighbors[i][2], baseCellNeighbors[i][3],
-               baseCellNeighbors[i][4], baseCellNeighbors[i][5],
-               baseCellNeighbors[i][6], i,
+        printf("{%s, %s, %s, %s, %s, %s, %s}, // base cell %d%s\n",
```
isaacbrodsky: I would prefer we didn't leak the strings allocated on line 85. Could this be simply changed as:

```
char[7]* neighborStrings;
getBaseCellOutput(baseCellNeighbors[i], neighborStrings);
// printf
destroyBaseCellOutput(neighborStrings);
```

src/apps/miscapps/generateBaseCellNeighbors.c
dfellis: Oh wow, I didn't even notice that and I'm amazed that compiled.

src/apps/miscapps/generateBaseCellNeighbors.c
dfellis: Sure, will make sure i486SX machines won't have any issues running this application. ;)

src/apps/miscapps/generateBaseCellNeighbors.c
dfellis: I didn't do it. :) (clang-format did)

PR #12, opened 2018-01-19
Remove H3IndexFat from geoToH3

dfellis:
Another diff to remove another piece of H3IndexFat. This one focuses on the `geoToH3` function.

I only did one function from the API because it was complex enough on its own, and I also didn't want to complete the refactoring without input from @isaacbrodsky and crew.

Since this implementation needed internals of H3Index, I put it into the `h3Index.c` file and removed `geoToH3` from the `h3api.c` file entirely. I'm planning on completely removing the `h3api.c` file along with the `h3IndexFat` files, unless there are objections. In this way `h3api.h` is just a header for exported functions, whereever they live, and that seems fine to me.

I also removed the dead code from `h3IndexFat.c` that the code coverage told me was unused. I figured that'd be better than a temporary dip on coverage percentage?

### DIFF
--- a/src/h3lib/lib/h3Index.c
+++ b/src/h3lib/lib/h3Index.c
@@ -19,8 +19,10 @@
  */
 #include "h3Index.h"
 #include <inttypes.h>
+#include <math.h>
 #include <stdlib.h>
 #include <string.h>
+#include "baseCells.h"
 #include "faceijk.h"
 #include "h3IndexFat.h"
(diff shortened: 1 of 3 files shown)


isaacbrodsky:
Eventually removing h3api.c makes sense to me. The alternative would be to move all the public functions to h3api.c, which seems less reasonable. I'll leave this open just a little longer in case others have thoughts on this, but looks good to me.

PR #13, opened 2018-01-21
Remove H3IndexFat

dfellis:
Made the change to remove H3IndexFat completely from the H3!

This change depends on the work in #12 so that has to be landed first, then I can rebase this on latest master (it should rebase perfectly). This also eliminates `h3Api.c` and appears to speed up H3 by 9% according to the benchmarks. (I haven't even gotten started on all of the perf stuff I want to do. I think the IJK rotation logic could be reduced from a loop to an O(1) algo and speed many things up.)


src/apps/testapps/testH3Index.c
```diff
@@ -47,6 +47,20 @@ TEST(h3IsValidDigits) {
              "h3IsValid failed on invalid unused digits");
 }
 
+TEST(h3IsValidBaseCell) {
```
isaacbrodsky: I think we're losing a test that an index with an invalid base cell (such as 127) is invalid.

src/h3lib/include/h3api.h
```diff
@@ -43,6 +43,9 @@
 /* For size_t */
 #include <stdlib.h>
 
+/* For constants used throughout H3 */
+#include "constants.h"
```
isaacbrodsky: Why is this include added?

dfellis: Many parts of the codebase depended on the constants being defined that they were getting indirectly through inclusion of `H3IndexFat.h`. It was easier to add it in here than in so many places across the codebase, and in some ways it makes sense -- if you can't `#include "h3api.h"` and have enough defined to be able to actually compile the requisite methods, then what kind of API is that?

src/apps/testapps/testH3Index.c
dfellis: Oh? I'll look for it and include it in here. I satisfied myself with reaching 100% code coverage last night.

src/h3lib/include/h3api.h
isaacbrodsky: I think `h3api.h` exists for consumers of the library, not for the implementation, and consumers would not need these internal constants. I would prefer to replace `#include "h3IndexFat.h"` with `#include "constants.h"` if that's needed.

dfellis: Will do.

src/apps/testapps/testH3Index.c
dfellis: Done.

src/h3lib/include/h3api.h
dfellis: Done.

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-    H3IndexFat hfLargeRes;
-    setH3IndexFat(&hfLargeRes, 15, 0, 0);
-    // Can't set resolution to 16 using setH3IndexFat
-    // since it would corrupt the stack.
-    hfLargeRes.res = 16;
-    t_assert(!h3FatIsValid(&hfLargeRes),
-             "h3FatIsValid failed on large resolution");
-}
-
-TEST(h3FatUnusedDigitInvalid) {
```
isaacbrodsky: I think this is covered by h3IsValidDigits although not as comprehensively

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-}
-
-TEST(h3FatNegativeResInvalid) {
-    H3IndexFat hfNegativeRes;
-    setH3IndexFat(&hfNegativeRes, -1, 0, 0);
-    t_assert(!h3FatIsValid(&hfNegativeRes),
-             "h3FatIsValid failed on negative resolution");
-}
-
-TEST(h3FatLargeResolutionInvalid) {
```
isaacbrodsky: This one can't be tested on H3Index because there's only 4 bits for the resolution.

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-        H3IndexFat hf;
-        setH3IndexFat(&hf, 0, 0, 0);
-        hf.mode = i;
-        char failureMessage[BUFF_SIZE];
-        sprintf(failureMessage, "h3FatIsValid failed on mode %d", i);
-        t_assert(!h3FatIsValid(&hf) || i == 1, failureMessage);
-    }
-}
-
-TEST(h3FatNegativeResInvalid) {
```
isaacbrodsky: This one can't be tested on H3Index because there's only 4 bits for the resolution.

src/apps/testapps/testH3Index.c
```diff
@@ -47,6 +48,28 @@ TEST(h3IsValidDigits) {
...
+        char failureMessage[BUFF_SIZE];
+        sprintf(failureMessage, "h3IsValid failed on base cell %d", i);
+        t_assert(h3IsValid(h), failureMessage);
+
+        t_assert(H3_EXPORT(h3GetBaseCell)(h) == i,
+                 "failed to recover base cell");
+    }
+}
+
+TEST(h3FatIsValidBaseCellInvalid) {
```
isaacbrodsky: nit: h3Fat -> h3

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
- */
-
-#include "h3Index.h"
-#include "h3IndexFat.h"
-#include "test.h"
-#include "utility.h"
-
-BEGIN_TESTS(h3IndexFat);
-
-TEST(h3FatIsValidWithMode) {
```
isaacbrodsky: I don't think we have a corresponding test for this

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-             "h3FatIsValid failed on invalid index digit (14)");
-
-    hf.res = 1;
-    hf.index[14] = 7;
-    hf.index[1] = 1;
-    t_assert(!h3FatIsValid(&hf),
-             "h3FatIsValid failed on invalid index digit (1)");
-}
-
-TEST(h3FatBadDigitInvalid) {
```
isaacbrodsky: I don't think this is tested (would look like res=1 but the digits are all 7)

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-        sprintf(failureMessage, "h3FatIsValid failed on base cell %d", i);
-        t_assert(h3FatIsValid(&hf), failureMessage);
-
-        H3Index h = h3FatToH3(&hf);
-        t_assert(H3_EXPORT(h3GetBaseCell)(h) == i,
-                 "failed to recover base cell");
-    }
-}
-
-TEST(h3FatIsValidNegativeBaseCellInvalid) {
```
isaacbrodsky: This can't be expressed in H3Index

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-    setH3IndexFat(&hfNegative, 1, 0, -1);
-    t_assert(!h3FatIsValid(&hfNegative),
-             "h3FatIsValid failed on negative digit");
-
-    H3IndexFat hfLarge;
-    setH3IndexFat(&hfLarge, 1, 0, 7);
-    t_assert(!h3FatIsValid(&hfLarge), "h3FatIsValid failed on too large digit");
-}
-
-TEST(h3FatIsValidBaseCell) {
```
isaacbrodsky: This test is ported over.

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-
-TEST(h3FatIsValidNegativeBaseCellInvalid) {
-    H3IndexFat hfNegativeBaseCell;
-    initH3IndexFat(&hfNegativeBaseCell, 0);
-    hfNegativeBaseCell.baseCell = -1;
-    t_assert(!h3FatIsValid(&hfNegativeBaseCell),
-             "h3FatIsValid failed on negative base cell");
-}
-
-TEST(h3FatIsValidBaseCellInvalud) {
```
isaacbrodsky: This test is ported over.

src/apps/testapps/testH3IndexFat.c
```diff
@@ -1,148 +0,0 @@
...
-    H3IndexFat hfWrongBaseCell;
-    initH3IndexFat(&hfWrongBaseCell, 0);
-    hfWrongBaseCell.baseCell = NUM_BASE_CELLS;
-    t_assert(!h3FatIsValid(&hfWrongBaseCell),
-             "h3FatIsValid failed on invalid base cell");
-}
-
-TEST(setH3IndexFat) {
```
isaacbrodsky: This is already tested on H3Index (setH3Index)

src/apps/testapps/testH3Index.c
dfellis: Sorry I missed that.

src/apps/testapps/testH3IndexFat.c
dfellis: Adding.

src/apps/testapps/testH3IndexFat.c
dfellis: Skipping the negative test (that can't exist, now) and porting the digit 7 test.

src/h3lib/lib/h3Index.c
```diff
@@ -663,3 +645,118 @@ H3Index H3_EXPORT(geoToH3)(const GeoCoord* g, int res) {
...
 }
+
+/**
+ * Convert an H3Index to the FaceIJK address on a specified icosahedral face.
+ * @param h The H3Index.
+ * @param fijk The FaceIJK address, initialized with the desired face
+ *        and normalized base cell coordinates.
+ * @return Returns 1 if the possibility of overage exists, otherwise 0.
+ */
+int _h3ToFaceIjkWithInitializedFijk(H3Index h, FaceIJK* fijk) {
```
joegilley: nit: `s/_h3ToFaceIjkWithInitializedFijk/_h3ToFaceIJKWithInitializedFIJK` and same for others

dfellis: So that's how the original versions of these functions were named in H3IndexFat, and it makes a kind of sense to camelCase that way since it would be obvious to split as "h3 to face ijk with initialized fijk" whereas the other way would split as "h3 to face i j k with initialized f i j k".

dfellis:
Wasn't sure if you guys wanted to manage the CHANGELOG yourselves or allow contributors to (since my other diffs didn't get entries). Since they're all focused on eliminating H3IndexFat, that should still be fine. :)

src/h3lib/lib/h3Index.c
joegilley: If that’s how it was, no need to bunch unrelated changes into this.  Comment rescinded!

I don’t buy the `Ijk` makes more sense than `IJK` argument, though :P

PR #14, opened 2018-01-21
Change _hashVertex to use fmod.

isaacbrodsky:
@nrabinowitz This hash function encounters undefined behavior (C99 6.3.1.4.1, casting a float where the integral part cannot be represented by the integer type is undefined), which causes the test suite to fail on armv7. The change here is to put the integral part in range of uint32_t before casting.

It looks to me like the standard way of writing this function would be something like `lat * 37 + lon`, to avoid `lat=1,lon=2` having the same hash as `lat=2,lon=1`. It also looks like this could be done with `union`s of `double` and `uint64_t`, but not sure we want to do that since it would be undefined behavior as I understand.

### DIFF
--- a/src/h3lib/lib/vertexGraph.c
+++ b/src/h3lib/lib/vertexGraph.c
@@ -64,8 +64,8 @@ void destroyVertexGraph(VertexGraph* graph) {
 uint32_t _hashVertex(const GeoCoord* vertex, int res, int numBuckets) {
     // Simple hash: Take the sum of the lat and lon with a precision level
     // determined by the resolution, converted to int, modulo bucket count.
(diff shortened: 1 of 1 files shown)


nrabinowitz:
> It looks to me like the standard way of writing this function would be something like lat * 37 + lon, to avoid lat=1,lon=2 having the same hash as lat=2,lon=1

I'd be fine updating it - the goal here was to have a simple hash function that was very fast and worked across FPE issues (i.e. correctly recognized vertices from adjacent hexagons as identical). Because it's only ever dealing with a limited set of cases I didn't overthink it, but certainly differentiating lat and lon seems like a good idea. Why `37` though?

isaacbrodsky:
> Why `37` though?

I suggested it arbitrarily. The guidance on writing hash functions is usually to use multiply by a prime.

PR #15, opened 2018-01-21
Don't require BUILD_STATIC for Windows tests.

isaacbrodsky:
Makes the tests and misc. apps compatible with BUILD_STATIC=OFF on Windows. Previously, they didn't work because the base cell tables were not exported. It also changes the build to place the DLL in the `bin` directory, only on Windows.

### DIFF
--- a/src/h3lib/include/baseCells.h
+++ b/src/h3lib/include/baseCells.h
@@ -46,5 +46,8 @@ extern const BaseCellData baseCellData[NUM_BASE_CELLS];
 int _isBaseCellPentagon(int baseCell);
 int _faceIjkToBaseCell(const FaceIJK* h);
 int _faceIjkToBaseCellCCWrot60(const FaceIJK* h);
+void _baseCellToFaceIjk(int baseCell, FaceIJK* h);
+bool _baseCellIsCwOffset(int origin, int test);
+int _getBaseCellNeighbor(int baseCell, int dir);
 
 #endif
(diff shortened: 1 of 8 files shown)


src/h3lib/include/baseCells.h
```diff
@@ -46,5 +46,8 @@ extern const BaseCellData baseCellData[NUM_BASE_CELLS];
 int _isBaseCellPentagon(int baseCell);
 int _faceIjkToBaseCell(const FaceIJK* h);
 int _faceIjkToBaseCellCCWrot60(const FaceIJK* h);
+void _baseCellToFaceIjk(int baseCell, FaceIJK* h);
+bool _baseCellIsCwOffset(int origin, int test);
+int _getBaseCellNeighbor(int baseCell, int dir);
```
nrabinowitz: I thought the convention we were using was that `_`-prefixed functions were local to the module? We never went through and did a clean sweep here, but it might be worthwhile.

src/h3lib/lib/baseCells.c
```diff
@@ -848,3 +848,22 @@ int _faceIjkToBaseCellCCWrot60(const FaceIJK* h) {
...
+
+/** @brief Find the FaceIJK given a base cell.
+ */
+void _baseCellToFaceIjk(int baseCell, FaceIJK* h) {
+    *h = baseCellData[baseCell].homeFijk;
+}
+
+/** @brief Return whether or not the tested face is a cw offset face.
+ */
+bool _baseCellIsCwOffset(int origin, int test) {
```
nrabinowitz: Nit: `test` is pretty generic, maybe `testFace`?

src/h3lib/include/baseCells.h
isaacbrodsky: We should also add some developer documentation that explains the difference between H3_EXPORT, unprefixed, and prefixed functions.

Issue #18, opened 2018-01-25
Distance calculation


kaspermarstal:
Is it possible to calculate the distance in kilometers between the centers of two arbitrary hexagons using h3?

dfellis:
You can use `h3ToGeo` to get the center coordinates of each hexagon and then use [the Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) or [Vincenty's Formula](https://en.m.wikipedia.org/wiki/Vincenty%27s_formulae) to calculate the distance between them, depending on the speed vs accuracy trade-off you wish to make.

kaspermarstal:
Wow, thanks for the example and for opening #23! We would be happy to contribute a pull request with this functionality (under your specifications of course).

PR #22, opened 2018-01-29
add haversine distance calc example

joegilley:
adds an example for referencing in issue #18

### DIFF
--- a/examples/distance.c
+++ b/examples/distance.c
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * Example program that calculates the distance in kilometers between two
+ * hexagon indices.
+ */
+
+// mean Earth radius
+#define R 6371.0088
+
+#include <h3/h3api.h>
+#include <math.h>
+#include <stdio.h>
+
+double dist(double th1, double ph1, double th2, double ph2) {
+    double dx, dy, dz;
+    ph1 -= ph2;
+
+    dz = sin(th1) - sin(th2);
+    dx = cos(ph1) * cos(th1) - cos(th2);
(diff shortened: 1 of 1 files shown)


examples/distance.c
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+// mean Earth radius
+#define R 6371.0088
+
+#include <h3/h3api.h>
+#include <math.h>
+#include <stdio.h>
+
+double dist(double th1, double ph1, double th2, double ph2) {
```
nrabinowitz: Since this is an example, maybe include a comment on the algo here with a link to some reference?

examples/distance.c
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2017 Uber Technologies, Inc.
```
isaacbrodsky: 2018

examples/distance.c
nrabinowitz: Also, would it be better to just expose `_geoDistRads` or `_geoDistKm` from geoCoord?

isaacbrodsky: Add a comment for this function? What are the parameters, is it Haversine distance, etc?

joegilley: could do!  i **expect** cutting out the middle step and creating `h3DistKm` would be more valuable and appropriate for this lib, since that's what I'm doing here anyway, but cross that bridge if/when we find the need?

joegilley: done!

examples/distance.c
joegilley: is it 2018 already?!

examples/distance.c
joegilley: > maybe include a comment on the algo here

done!

examples/distance.c
```diff
@@ -0,0 +1,72 @@
...
+    return asin(sqrt(dx * dx + dy * dy + dz * dz) / 2) * 2 * R;
+}
+
+int main(int argc, char *argv[]) {
+    // 1455 Market St @ resolution 15
+    H3Index h3HQ1 = stringToH3("8f2830828052d25");
+    // 555 Market St @ resolution 15
+    H3Index h3HQ2 = stringToH3("8f283082a30e623");
+
+    double distance = 0;
```
isaacbrodsky: Looks like `double distance` is not used.

joegilley: uh oh, possible to configure linter to catch?  compiling with `-Wall` catches this...

joegilley: > publicizing geoDistRads/geoDistKm

I suggest we put out a minor version feature bump with these two and `h3DistRads/h3DistKm`; doing that is out of scope for this addition. like I mentioned to @nrabinowitz , I expect the h3 distance functions will be more useful and appropriate than yet another set of geo distance functions.

dfellis:
Making `geoDistKm` a first-class citizen really does sound like a useful change (and worthy of 3.1.0). You can add my vote on that change, too. :)

joegilley:
created #23 for tracking

PR #25, opened 2018-02-04
Automatically run tests under Valgrind

isaacbrodsky:
Adds an option for running all tests under Valgrind, `WRAP_VALGRIND`, or another wrapper using `TEST_WRAPPER`.

Another possible improvement would be to automatically run Valgrind as part of CI.

### DIFF
--- a/src/apps/testapps/testVertexGraph.c
+++ b/src/apps/testapps/testVertexGraph.c
@@ -235,10 +235,7 @@ TEST(removeVertexNode) {
     t_assert(graph.size == 2, "Graph size decremented");
 
     // Remove non-existent node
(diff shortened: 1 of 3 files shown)


src/apps/testapps/testVertexGraph.c
```diff
@@ -236,6 +236,9 @@ TEST(removeVertexNode) {
 
     // Remove non-existent node
     node = malloc(sizeof(VertexNode));
+    node->next = NULL;
```
nrabinowitz: Easier just to use `calloc` instead of `malloc` here?

isaacbrodsky: It would be easier; done.

PR #26, opened 2018-02-04
Add check for out of range coordinates when indexing.

isaacbrodsky:
Increase the library's resilience to out of range and defective inputs.

### DIFF
--- a/src/h3lib/lib/h3Index.c
+++ b/src/h3lib/lib/h3Index.c
@@ -46,7 +46,7 @@ int H3_EXPORT(h3GetBaseCell)(H3Index h) { return H3_GET_BASE_CELL(h); }
  * @return The H3 index corresponding to the string argument, or 0 if invalid.
  */
 H3Index H3_EXPORT(stringToH3)(const char* str) {
-    H3Index h = 0;
+    H3Index h = H3_INVALID_INDEX;
     // If failed, h will be unmodified and we should return 0 anyways.
     sscanf(str, "%" PRIx64, &h);
     return h;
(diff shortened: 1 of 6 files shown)


isaacbrodsky:
I haven't yet debugged why the tests work locally but not in CI, I would assume the problem is we're running into undefined behavior. I might add a test that calls `_faceIjkToH3` directly. A test could also be added to `geoToH3` for out of range coordinates.

dfellis:
It's failing in the test H3Index test. I don't have my laptop this weekend, so I can't check any further why.

src/h3lib/lib/h3Index.c
```diff
@@ -557,6 +557,11 @@ H3Index _faceIjkToH3(const FaceIJK* fijk, int res) {
 
     // check for res 0/base cell
     if (res == 0) {
+        if (fijk->coord.i > 2 || fijk->coord.j > 2 || fijk->coord.k > 2) {
```
nrabinowitz: Can we have a well-named constant for `2` here?

joegilley: +1 well-named constant and maybe a comment on why > 2 for any of the IJK is bad too :D ?

src/h3lib/lib/h3Index.c
```diff
@@ -557,6 +557,11 @@ H3Index _faceIjkToH3(const FaceIJK* fijk, int res) {
...
+            // out of range input
+            return 0;
```
joegilley: is this a new sentinel value? if it is, can we add it to a documentation? if it isn't new, never mind :)

isaacbrodsky: This is the same sentinel used in geoToH3, which is why I used it here.

isaacbrodsky:
I tracked the failure to a difference in `_geoDistRads` when compiling under GCC 4.8, causing a difference in which face is selected for the out of bounds coordinates. I'll determine why that is different next, or change the test to not depend on it.

PR #29, opened 2018-03-12
Improve CMake build and install

isaachier:


CMakeLists.txt
```diff
@@ -345,19 +427,78 @@ if(ENABLE_TESTS)
...
+# Installation (https://github.com/forexample/package-example)
+
+# Layout. This works for all platforms:
+#   * <prefix>/lib/cmake/<PROJECT-NAME>
+#   * <prefix>/lib/
+#   * <prefix>/include/
+set(config_install_dir "lib/cmake/${PROJECT_NAME}")
+set(include_install_dir "include")
+
+set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")
```
isaacbrodsky: I think generated/ should be added to .gitignore in case the build is being done in source.

dfellis:
This change removes the ability for developers to choose which C compiler they wish to use, and breaks building via gcc according to Travis CI. Both of those issues ought to be addressed before landing this.

isaachier:
No they just force you to use the default options instead of rolling your own. Use CMAKE_C_COMPILER instead. Same applies for BUILD_SHARED_LIBS.

isaachier:
And yes I can fix Travis.

CMakeLists.txt
```diff
@@ -61,40 +49,136 @@ endif()
...
 
-include_directories(src/h3lib/include)
+include(CMakeDependentOption)
+include(CTest)
 
-file(GLOB_RECURSE ALL_SOURCE_FILES src/*.c src/*.h)
-file(GLOB_RECURSE EXAMPLE_SOURCE_FILES examples/*.c)
+find_library(M_LIB m)
+
+set(LIB_SOURCE_FILES
```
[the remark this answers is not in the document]
isaachier: I think it helps make sure you actually build the files in question. Otherwise, make might not have info on a new file. And sure, my vim is just set to two spaces.

isaachier:
@dfellis Travis is fine. What you are seeing are clang-tidy checks that failed. There are worth fixing anyway.

isaachier:
And the error stems from the coverage being broken, not the emitted warnings. See the end of the build log.

```
Capturing coverage data from /home/travis/build/uber/h3
Found gcov version: 4.8.4
Scanning /home/travis/build/uber/h3 for .gcda files ...
geninfo: ERROR: no .gcda files found in /home/travis/build/uber/h3!
```

IDK how that was introduced but I didn't touch the coverage stuff.

dfellis:
You'll note that the other PR up right now did not have code coverage failures for GCC. Despite your claim, this change does currently break the coverage reporting.

isaachier:
OK duly noted. Let me research this issue.

.ycm_extra_conf.py
```diff
@@ -0,0 +1,98 @@
+# Copyright (c) 2018, Uber Technologies, Inc
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
```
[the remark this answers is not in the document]
isaachier: This might be copied from my Jaeger client that went through an MIT to Apache 2 license migration. Let me fix it.

CMakeLists.txt
```diff
@@ -136,17 +213,13 @@ macro(add_h3_executable name)
     # invoke built-in add_executable
     add_executable(${ARGV})
     if(TARGET ${name})
-        target_link_libraries(${name} h3.1)
-        if (M_LIB)
-            target_link_libraries(${name} ${M_LIB})
-        endif()
-        target_include_directories(${name} PUBLIC src/apps/applib/include)
+        target_link_libraries(${name} PUBLIC h3)
+        target_link_libraries(${name} PUBLIC ${M_LIB})
```
isaacbrodsky: I think there needs to be an `if (M_LIB)` guard due to platforms that don't have libm (WIndows).

isaachier: At first I though that would be the case. Then I thought, M_LIB should expand to be empty if not found so it will be a no-op in that case and fine. But I thought I finally reverted back to the if guard after realizing M_LIB actually gets set to libm-NOTFOUND in that case (only in CMake would that be a falsy if value, check the docs for more on that). So if it isn't there, I'll definitely put the if back where it was.

isaacbrodsky:
Thanks for updating with the README and CHANGELOG updates.

PR #32, opened 2018-03-19
Allocation improvements

isaachier:


src/apps/benchmarks/benchmarkPolyfill.c
```diff
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <string.h>
```
dfellis: I don't see any code in here using `string.h`

src/apps/filters/hexRange.c
```diff
@@ -30,13 +30,15 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
```
dfellis: Same here. `printf` is in `stdio`, I don't see why `string` was included.

src/apps/filters/kRing.c
```diff
@@ -25,22 +25,22 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
```
dfellis: Please remove this.

src/apps/miscapps/generateBaseCellNeighbors.c
```diff
@@ -237,18 +226,19 @@ void generate() {
 
     printf("const int baseCellNeighbors[NUM_BASE_CELLS][7] = {\n");
     for (int i = 0; i < NUM_BASE_CELLS; i++) {
-        char* neighborStrings[7] = {0, 0, 0, 0, 0, 0, 0};
+        printf("{");
```
dfellis: Please undo the changes you made here. This approach will print garbage around the `INVALID_BASE_CELL` printout, instead of only printing that message when reaching that condition, so it will not generate the base cell neighbors code correctly, anymore.

src/h3lib/lib/algos.c
```diff
@@ -22,6 +22,7 @@
 #include <math.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <string.h>
```
dfellis: Remove this.

src/apps/benchmarks/benchmarkPolyfill.c
isaachier: `memset` for all of these comments.

src/apps/miscapps/generateBaseCellNeighbors.c
isaachier: Now I'm very confused. The difference between the old code and this is the removal of the `generateBaseCellOutput`, which simply formatted strings that are printed here. Instead, I used identical code here to write the digits/macro directly to stdout.

src/h3lib/lib/algos.c
```diff
@@ -90,9 +91,8 @@ int H3_EXPORT(maxKringSize)(int k) {
  */
 void H3_EXPORT(kRing)(H3Index origin, int k, H3Index* out) {
     int maxIdx = H3_EXPORT(maxKringSize)(k);
-    int* distances = calloc(maxIdx, sizeof(int));
+    int distances[maxIdx];
```
isaacbrodsky: I was a little concerned this does not initialize `distances`. It might be that `distances` cannot be used before initializing. Not sure if you want to add a memset here for clarity.

src/h3lib/lib/h3Index.c
```diff
@@ -340,17 +337,11 @@ int H3_EXPORT(compact)(const H3Index* h3Set, H3Index* compactedSet,
...
+        memset(hashSetArray, 0, sizeof(hashSetArray));
         compactedSetOffset += uncompactableCount;
-        remainingHexes = compactableHexes;
+        memcpy(remainingHexes, compactableHexes, sizeof(compactableHexes));
         numRemainingHexes = compactableCount;
     }
-    // Final cleanup
-    free(remainingHexes);
-    free(hashSetArray);
     return 0;
```
isaacbrodsky: It looks like this line is no longer reached in tests, perhaps because of the change around line 276. Could you change to preserve 100% coverage?

src/apps/miscapps/generateBaseCellNeighbors.c
isaacbrodsky: @dfellis Could you paste the output you are seeing or expecting? I ran this branch and got the same output as before.

dfellis: My mistake on the reading of his changes. I did not run his branch.

src/h3lib/lib/algos.c
isaachier: Good catch. Was almost certain I followed them all with a `memset`.

isaachier: Fixed.

src/h3lib/lib/algos.c
```diff
@@ -741,7 +742,8 @@ void H3_EXPORT(polyfill)(const GeoPolygon* geoPolygon, int res, H3Index* out) {
     // This first part is identical to the maxPolyfillSize above.
 
     // Get the bounding boxes for the polygon and any holes
-    BBox* bboxes = calloc(sizeof(BBox), geoPolygon->numHoles + 1);
+    BBox bboxes[geoPolygon->numHoles + 1];
+    memset(bboxes, 0, sizeof(BBox));
```
isaachier: This is a mistake. Let me fix now.

src/h3lib/lib/algos.c
```diff
@@ -741,7 +742,8 @@ void H3_EXPORT(polyfill)(const GeoPolygon* geoPolygon, int res, H3Index* out) {
     // This first part is identical to the maxPolyfillSize above.
 
     // Get the bounding boxes for the polygon and any holes
-    BBox* bboxes = calloc(sizeof(BBox), geoPolygon->numHoles + 1);
+    BBox bboxes[geoPolygon->numHoles + 1];
+    memset(bboxes, 0, sizeof(bboxes));
```
isaachier: Fixed.

src/h3lib/lib/h3Index.c
isaachier: Sure. Will push fix for that now.

isaachier:
Think I addressed all the comments.

isaacbrodsky:
I tested this on MSVC (Visual Studio 2017) and it seems it does not support variable length arrays. I would prefer the repo is limited to the subset of C99/C11 that MSVC supports to maximize how broadly it can be used.

src/h3lib/lib/linkedGeo.c
```diff
@@ -103,35 +106,27 @@ LinkedGeoCoord* addLinkedCoord(LinkedGeoLoop* loop, const GeoCoord* vertex) {
...
-    LinkedGeoCoord* nextCoord;
     // flag to skip the input polygon
-    int skip = 1;
-    while (currentPolygon != NULL) {
-        currentLoop = currentPolygon->first;
-        while (currentLoop != NULL) {
-            currentCoord = currentLoop->first;
-            while (currentCoord != NULL) {
+    bool skip = true;
+    for (LinkedGeoPolygon *currentPolygon = polygon, *nextPolygon;
```
nrabinowitz: Nit: I believe our style in this project is `LinkedGeoPolygon* currentPolygon`, can we use that here? Or would that require `LinkedGeoPolygon* currentPolygon, LinkedGeoPolygon* nextPolygon`?

isaachier: You mean the pointer alignment? Ya it's a problem. `clang-format` isn't even fixing it. The only alternative is `LinkedGeoPolygon* nextPolygon;` followed by `for (LinkedGeoPolygon* currentPolygon = ...)`. I once answered a question about this on StackOverflow here: https://stackoverflow.com/a/16490876/1930331.

cmake/CheckAlloca.cmake
```diff
@@ -0,0 +1,21 @@
+if(__check_alloca)
```
isaacbrodsky: Please include the license header in all source files.

src/h3lib/include/stackAlloc.h
```diff
@@ -0,0 +1,44 @@
+#ifndef STACKALLOC_H
```
isaacbrodsky: Please also include a `@file`/`@brief` comment describing the purpose of this file.

src/apps/miscapps/generateBaseCellNeighbors.c
```diff
@@ -237,24 +223,25 @@ void generate() {
...
+                printf("%d", baseCellNeighbors[i][j]);
+            } else {
+                printf("INVALID_BASE_CELL");
+            }
         }
-        printf("{%s, %s, %s, %s, %s, %s, %s}, // base cell %d%s\n",
-               neighborStrings[0], neighborStrings[1], neighborStrings[2],
-               neighborStrings[3], neighborStrings[4], neighborStrings[5],
-               neighborStrings[6], i,
+        printf("}, /* base cell %d%s */\n", i,
```
isaacbrodsky: Since the formatting of the comments changes, could you rerun generateBaseCellNeighbors and update baseCells.c?

isaacbrodsky:
I successfully built and tested this on MSVC :) - #34 will make that automated.

src/apps/miscapps/generateBaseCellNeighbors.c
isaachier: Sure thing.

cmake/CheckAlloca.cmake
isaachier: OK.

src/h3lib/include/stackAlloc.h
isaachier: Good idea

isaachier:
Luckily even MSVC has the alloca function. It's a bit ugly but this should definitely allow some form of VLA on all popular platforms.

src/apps/miscapps/generateBaseCellNeighbors.c
```diff
@@ -221,9 +243,52 @@ static void generate() {
 
     auditBaseCellNeighbors(baseCellNeighbors, baseCellRotations);
 
+    time_t rawTime;
+    time(&rawTime);
+    struct tm tmStruct;
+    reentrant_gmtime(&tmStruct, &rawTime);
+    const char licenseFormat[] =
+        "/*\n"
+        " * Copyright 2016-%d Uber Technologies, Inc.\n"
```
isaacbrodsky: This seems like a lot of work just to automatically format the date. I don't think the parts of the baseCells.c file unrelated to the baseCellNeighbors/baseCellNeighbor60CCWRots should be generated. (Those two were the only parts of the file that were generated from this - maybe those two could be split off to their own, generated file.)

Could you revert this change, and we can revisit it in another issue or PR?

Issue #34, opened 2018-03-22
Add Windows and Mac OS X CI Support


dfellis:
A few pull requests (especially those by @isaachier ) have been blocked because the change breaks compiling on Windows, but there's no obvious way to be sure of this right now, [since Travis CI doesn't support Windows](https://docs.travis-ci.com/user/multi-os/) (but does support OS X).

It appears that it might make sense to use [Microsoft's free Visual Studio Team Services Continuous Integration](https://www.visualstudio.com/team-services/continuous-integration/) for Windows support. There are instructions for setting up both [gcc-based](https://docs.microsoft.com/en-us/vsts/build-release/apps/c-cpp/gcc?view=vsts&tabs=github) and [visual studio-based](https://docs.microsoft.com/en-us/vsts/build-release/apps/windows/cpp?view=vsts&tabs=github) testing.

Microsoft's CI solution supposedly also supports Linux testing, but not OS X testing, and I also couldn't find anything about code coverage reporting, so I think it would still be required to keep Travis CI (and turn on OS X testing, ideally), but [a second testing integration and badge](https://docs.microsoft.com/en-us/vsts/build-release/actions/ci-build-github?view=vsts#create-a-vsts-build-status-with-a-github-readme-file) could be added for the runs on Windows.

Again, this is something that I would like to help on, but most likely Uber would want to own the Microsoft CI account.

I think I can do the OS X integration, however, if we want to split this into two tasks.

isaachier:
I have a fix for Windows. You can see the appveyor.yml here: https://github.com/isaachier/h3/blob/appveyor/appveyor.yml. The associated builds are here: https://ci.appveyor.com/project/isaachier/h3.

isaacbrodsky:
I'll let @isaachier submit a PR to add Windows CI. Let me know if you'd like me to help, or to work on that.

I think the OSX test part is pretty simple, so I can make a quick PR for that.

isaachier:
Yes I'd love to. Probably early tomorrow. Just wanted to keep the allocation PR clean.

PR #35, opened 2018-03-22
Add OSX build in Travis, add compatibility note.

isaacbrodsky:
This PR also adds a few requests to CONTRIBUTING.md about compiler compatibility and coverage.

Maybe we should mention in CONTRIBUTING.md that the code is C11?

Edit: #34

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,8 @@ matrix:
         - make && make coverage
       after_success:
         - coveralls --lcov-file coverage.cleaned.info --verbose
+    - env: NAME="Mac OSX (Xcode 8)"
+      os: osx
 
 # Configure the build script, out of source.
 before_script:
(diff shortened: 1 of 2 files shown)


dfellis:
While it's true that [the OS X test ran as "gcc" via clang's gcc-compatibility mode](https://travis-ci.org/uber/h3/jobs/356692356) perhaps it would also be worthwhile to explicitly test on OS X with "normal" clang?

Not a blocker, just a suggestion.

isaacbrodsky:
@dfellis I'm not sure what the difference would be, so I think I'll leave that for later.

dfellis:
I could have sworn I'd read an article a few years ago about clang detecting what executable name it was running under and being more/less aggressive on gcc-isms, but I can't find the article or any support in the clang documentation about that.

PR #37, opened 2018-03-22
Fix references lacking `H3_EXPORT`

isaachier:
Out of curiosity, tried building with `-DH3_PREFIX=h3`. Got compilation and linking errors. These are the fixes I made locally.


isaachier:
As I mentioned I just set the prefix. Seeing as there is 100% coverage, nothing should be left.

Issue #45, opened 2018-04-26
h3ToGeoBoundary returns duplicate points for some hexagons


nrabinowitz:
Certain hexagons (including at least some on the edge of the icosahedron face) have duplicate points generated by `h3ToGeoBoundary`:

```
$ ./bin/h3ToGeoBoundary
894cc536537ffff
894cc536537ffff
{
   18.043333154 293.721634765
   18.042238363 293.720709372
   18.040818259 293.721458061
   18.040492975 293.723132133
   18.041040385 293.723594817
   18.041757122 293.724032885
   18.043007860 293.723308818
   18.043007860 293.723308818
}
```
The last two vertices here are duplicates. The repeated vertex causes problems in the generation of GeoJSON boundaries.

Will update this issue when I find more details on which hexagons are causing the problem.

nrabinowitz:
Pretty sure this is the result of some Class III hexagons having vertexes directly on the icosahedron edge - those vertices are duplicated.

isaacbrodsky:
We noticed that these duplicated vertices are shared with the parent class II hexagon, which might be useful when excluding them from output.

nrabinowitz:
Closed in #46 .

PR #46, opened 2018-04-27
Fix duplicate vertex in h3ToGeoBoundary for certain class III hexagons

nrabinowitz:
Addresses #45. Class III hexagons with a vertex on the icosahedron edge were getting an extra, duplicate vertex. Fixing this revealed hundreds of instances of this issue in our test data, which is also fixed here.


src/apps/applib/lib/utility.c
```diff
@@ -163,3 +164,16 @@ int readBoundary(FILE* f, GeoBoundary* b) {
 
     return 0;
 }
+
+void assertExpectedBoundary(H3Index h3, const GeoBoundary* b1) {
```
nrabinowitz: Moved this to `utility.c` to share - let me know if you think it should go into `test.c` instead.

isaacbrodsky: Is this used by non-tests? Seems like it should be in test.c.

src/h3lib/lib/vec2d.c
```diff
@@ -52,3 +52,7 @@ void _v2dIntersect(const Vec2d* p0, const Vec2d* p1, const Vec2d* p2,
     inter->x = p0->x + (t * s1.x);
     inter->y = p0->y + (t * s1.y);
 }
+
+bool _v2dEquals(const Vec2d* p0, const Vec2d* p1) {
```
isaacbrodsky: Please add a function comment

src/apps/applib/lib/utility.c
dfellis: Agreed.

.gitignore
```diff
@@ -67,6 +67,10 @@ cmake-build-debug/
 generated/
 h3Config.cmake
 h3ConfigVersion.cmake
+CMakeDoxyfile.in
+CMakeDoxygenDefaults.cmake
+DartConfiguration.tcl
```
dfellis: This is a strange one to include in the gitignore.

src/h3lib/lib/faceijk.c
```diff
@@ -748,9 +748,14 @@ void _faceIjkToGeoBoundary(const FaceIJK* h, int res, int isPentagon,
             // find the intersection and add the lat/lon point to the result
             Vec2d inter;
             _v2dIntersect(&orig2d0, &orig2d1, edge0, edge1, &inter);
-            _hex2dToGeo(&inter, centerIJK.face, adjRes, 1,
-                        &g->verts[g->numVerts]);
-            g->numVerts++;
+            // if the intersection is the same as one of the edge endpoints,
+            // the additional vertex is on the edge itself and unnecessary
```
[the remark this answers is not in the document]
dfellis: Not sure if that's actually better. How about: "When a hexagon crosses over an edge, there is shape distortion similar to cutting two differently-stretched hexagons in half and then gluing them together. Most of the time this will add new vertices to the resulting 'hexagon', but if the cut goes perfectly through an existing vertex, it won't, so guard for that situation."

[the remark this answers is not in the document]
sahrk: This bug sounds like my bad, sorry. I should have caught that case. 

Maybe too much, but how about:

“Each face of the underlying icosahedron is a different projection plane. So if an edge of the hexagon crosses an icosahedron edge, an additional vertex must be introduced at that intersection point. Then each half of the cell edge can be projected to geographic coordinates using the appropriate icosahedron face projection. If a point of intersection occurs at a hexagon vertex, then each adjacent hexagon edge will lie completely on a single icosahedron face, and no additional vertex is requried.”

src/apps/testapps/testH3Api.c
```diff
@@ -47,4 +49,34 @@ TEST(geoToH3_coord) {
...
+    char* hexes[] = {"894cc5349b7ffff", "894cc534d97ffff", "894cc53682bffff",
+                     "894cc536b17ffff", "894cc53688bffff", "894cead92cbffff",
+                     "894cc536537ffff", "894cc5acbabffff", "894cc536597ffff"};
+    int numHexes = sizeof(hexes) / sizeof(hexes[0]);
+    H3Index h3;
+    GeoBoundary b;
+    for (int i = 0; i < numHexes; i++) {
+        h3 = H3_EXPORT(stringToH3)(hexes[i]);
+        H3_EXPORT(h3ToGeoBoundary)(h3, &b);
+        t_assert(b.numVerts == 7, "Got expected vertex count");
```
joegilley: consistency supernit: all other assertion messages start with lowercase :P

src/apps/testapps/testH3Api.c
```diff
@@ -47,4 +49,34 @@ TEST(geoToH3_coord) {
              "coordinates with infinity are rejected");
 }
 
+TEST(h3ToGeoBoundary_classIIEdgeVertex) {
```
joegilley: should this be `classIIIEdge` (not `classIIEdge`)?  unless i misunderstood issue...

src/h3lib/lib/faceijk.c
```diff
@@ -748,9 +748,14 @@ void _faceIjkToGeoBoundary(const FaceIJK* h, int res, int isPentagon,
...
+            if (!(_v2dEquals(&orig2d0, &inter) ||
```
joegilley: i might consider assigning this condition to a named variable to improve readability due to its complexity.

.gitignore
isaachier: It's new in 3.11. From the release notes:

 * The "FindDoxygen" module "doxygen_add_docs()" function now
   supports a new "DOXYGEN_VERBATIM_VARS" list variable.  Any
   "DOXYGEN_..." variable contained in that list will bypass the
   automatic quoting logic, leaving its contents untouched when
   transferring them to the output "Doxyfile".

src/h3lib/lib/faceijk.c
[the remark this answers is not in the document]
joegilley: def not too much, very clear description of whats happening 😄

isaachier:
Clever! But don't you mean `s/(   \-?\d+\.\d+ \d+\.\d+)\n\1/\1/g`?

src/h3lib/lib/faceijk.c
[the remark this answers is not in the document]
nrabinowitz: Took the recommendation, splitting into two comments, one for the outer block and one for the at-vertex check.

nrabinowitz:
> Clever! But don't you mean `s/( \-?\d+\.\d+ \d+\.\d+)\n\1/\1/g`?

Yeah, sorry, I actually did it through my IDE's search/replace but wanted to look all cool and `sed`-y. :)

nrabinowitz:
Addressed all comments, plus added a basic test suite for `Vec2d` functions.

isaacbrodsky:
I wrote a small program to count the lines that should be removed/should not be removed and the results looked OK to me.

src/apps/testapps/testVec2d.c
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2017 Uber Technologies, Inc.
```
dfellis: This should've been 2018.

nrabinowitz: True. Never known how to manage this, but presumably if we care we ought to run an annual script to update.

isaachier: OK so I thought the same but @yurishkuro advised me that the ideal solution is to update files you touch and instead of just replacing use this pattern `Copyright 2015, 2018` (if a range of years, I just do `2017-2018`). See this pull request in Jaeger to avoid blindly replacing year in all files https://github.com/jaegertracing/jaeger/commit/d7d1c74fa74f77d2c7faa50a0c568f979d9a30fd.

dfellis: In this case the entire file was being added, so just 2018 would be correct.

Personally I find copyright notices in source code anachronistic in the age of Git. You have the full history of what was edited by whom and the LICENSE file to tell you what rules apply.

isaachier: OK I see. And yes completely agree. IDK how legal feels about that but personally makes no sense. I think the fear is for those that copy a source file directly out of a project and use it in their own work. Seems like an unjustified fear.

PR #51, opened 2018-05-03
Replace existing bundle/styles with new bundle/styles

isaacbrodsky:
This fixes a small bug in the publish to gh-pages script, it wouldn't drop the old bundle and styles when making the commit.


nrabinowitz:
Looks good, merging.

PR #56, opened 2018-05-16
Change signature of h3NeighborRotations to not accept `CoordIJK*`

isaacbrodsky:
This always required looking up the CoordIJK structure in some array, and then prompty changing it back to an indexing digit. Using an indexing digit also allows for some code to be cleaned up (deduplicated) in H3 index rotation.

This could set up for using an enum type for the indexing digits, which might provide some compiler niceties.

Did not clean up some defines like UNIT_VECS, which could now be replaced with some bitwise manipulation in `_unitIjkToDigit` (changes undefined behavior though.)

### DIFF
--- a/src/h3lib/lib/coordijk.c
+++ b/src/h3lib/lib/coordijk.c
@@ -258,11 +258,12 @@ int _unitIjkToDigit(const CoordIJK* ijk) {
     _ijkNormalize(&c);
 
     int digit = INVALID_DIGIT;
-    for (int i = 0; i < 7; i++)
+    for (int i = 0; i < 7; i++) {
         if (_ijkMatches(&c, &UNIT_VECS[i])) {
             digit = i;
             break;
         }
+    }
 
     return digit;
 }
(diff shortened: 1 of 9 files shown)


src/h3lib/lib/coordijk.c
```diff
@@ -402,6 +403,54 @@ void _ijkRotate60cw(CoordIJK* ijk) {
     _ijkNormalize(ijk);
 }
 
+/**
+ * Rotates indexing digit 60 degrees counter-clockwise. Returns result.
+ *
+ * @param digit Indexing digit (between 1 and 6 inclusive)
+ */
+int _rotate60ccw(int digit) {
+    switch (digit) {
```
nrabinowitz: This is mostly for my info: I'm a little confused here. Given the ASCII diagram on L45, I was expecting this to look like `(digit + 1) % 6`, and was going to leave a note that this would be clearer, but looking through the cases here that's not the case. What exactly is being rotated here, and how does it relate to the diagram on L45?

src/h3lib/lib/h3UniEdge.c
```diff
@@ -108,7 +108,7 @@ H3Index H3_EXPORT(getH3UnidirectionalEdge)(H3Index origin,
     H3Index neighbor;
     for (int i = 1; i < 7; i++) {
         int rotations = 0;
-        neighbor = h3NeighborRotations(origin, &UNIT_VECS[i], &rotations);
+        neighbor = h3NeighborRotations(origin, i, &rotations);
```
nrabinowitz: Can we rename the iteration var here from `i` to `direction` for clarity?

src/h3lib/lib/coordijk.c
isaacbrodsky: I assume you mean algos.c line 45? That shows the order, but not the value. The value being rotated here is an [indexing digit](https://uber.github.io/h3/#/documentation/core-library/h3-index-representations), same as used in H3Index. Low bit is k, second bit is j, third bit is i.

nrabinowitz: Thanks, makes sense.

PR #60, opened 2018-05-22
Add examples for compact, kring, unidirectional edge

isaacbrodsky:
Adds some additional example programs.

It would be nice for CI to also check that the examples run (same for the filters), but that doesn't work directly because the examples assume `h3api.h` is at a different path. Perhaps this is something that can be added to the Travis CI config instead of to CMake? Or add a target for building the examples that can be run after installing in Travis CI?

### DIFF
--- a/examples/compact.c
+++ b/examples/compact.c
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2018 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * Example program that compacts a set of indexes, and then uncompacts
+ * to the original set.
+ */
+
+#include <assert.h>
+#include <h3/h3api.h>
+#include <stdio.h>
+
+int main(int argc, char* argv[]) {
+    H3Index input[] = {
+        // All with the same parent index
+        0x8a2a1072b587fffL, 0x8a2a1072b5b7fffL, 0x8a2a1072b597fffL,
+        0x8a2a1072b59ffffL, 0x8a2a1072b58ffffL, 0x8a2a1072b5affffL,
+        0x8a2a1072b5a7fffL,
+        // These don't have the same parent indexas above.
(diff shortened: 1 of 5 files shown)


examples/compact.c
```diff
@@ -0,0 +1,66 @@
...
+        0x8a2a1072b59ffffL, 0x8a2a1072b58ffffL, 0x8a2a1072b5affffL,
+        0x8a2a1072b5a7fffL,
+        // These don't have the same parent indexas above.
+        0x8a2a1070c96ffffL, 0x8a2a1072b4b7fffL, 0x8a2a1072b4a7fffL};
+    int inputSize = sizeof(input) / sizeof(H3Index);
+
+    H3Index* compacted = calloc(inputSize, sizeof(H3Index));
+    // neighboring is the input set to be compacted
+    int err = compact(input, compacted, inputSize);
+    assert(err == 0);
```
nrabinowitz: Include comments explaining why you might get errors here and below?

examples/compact.c
```diff
@@ -0,0 +1,66 @@
...
+
+    int count = 0;
+    printf("Compacted:\n");
+    for (int i = 0; i < inputSize; i++) {
+        if (compacted[i] != 0) {
+            printf("%llx\n", compacted[i]);
+            count++;
+        }
+    }
+    printf("Compacted to %d indexes.\n", count);
```
nrabinowitz: Nit: If users are meant to run the example, then I'd print out the starting and uncompacted counts as well.

examples/edge.c
```diff
@@ -0,0 +1,38 @@
...
+
+int main(int argc, char *argv[]) {
+    H3Index origin = 0x8a2a1072b59ffffL;
+    H3Index destination = 0x8a2a1072b597fffL;  // north of the origin
+
+    H3Index edge = getH3UnidirectionalEdge(origin, destination);
+    printf("The edge is %llx\n", edge);
+
+    GeoBoundary boundary;
+    getH3UnidirectionalEdgeBoundary(edge, &boundary);
```
nrabinowitz: Possibly just for my info: Is there an issue with uninitialized memory here? `getH3UnidirectionalEdgeBoundary` looks like it will only fill in the vertexes it needs, so it's possible that the rest of the `boundary` struct holds uninitialized memory - is this a problem?

isaacbrodsky: I don't believe so, but I'll make sure to run these in Valgrind. I don't think it's an issue because that's the same behavior as h3ToGeoBoundary, where `numVerts` is used to indicate how many vertices are valid.

examples/compact.c
```diff
@@ -0,0 +1,72 @@
...
+#include <h3/h3api.h>
+#include <stdio.h>
+
+int main(int argc, char* argv[]) {
+    H3Index input[] = {
+        // All with the same parent index
+        0x8a2a1072b587fffL, 0x8a2a1072b5b7fffL, 0x8a2a1072b597fffL,
+        0x8a2a1072b59ffffL, 0x8a2a1072b58ffffL, 0x8a2a1072b5affffL,
+        0x8a2a1072b5a7fffL,
+        // These don't have the same parent indexas above.
```
joegilley: indexas -> indexes

PR #62, opened 2018-05-24
Add CMakeLists for examples

isaacbrodsky:
Adds an example CMakeLists.txt using the CMake config installed by `make install`.

I'm experimenting with building and running the examples in Travis CI. It doesn't assert on the examples output. (Edit: also resolved some compiler warnings that showed up in CI)

Refactored the test wrapper (Valgrind) support code into its own file so it could be shared by both the main CMakeLists and the examples CMakeLists.

Depends on #60.


examples/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+
+# Testing the examples
+include(CTest)
+if(BUILD_TESTING)
+    include(../cmake/TestWrapValgrind.cmake)
+
+    enable_testing()
+
+    foreach(name ${EXAMPLES})
+        add_test(NAME ${name} COMMAND ${TEST_WRAPPER} "$<TARGET_FILE:${name}>")
```
isaachier: Have you tested with Valgrind. My guess is you don't need to replace semicolons above.

examples/compact.c
```diff
@@ -0,0 +1,73 @@
...
+        if (compacted[i] != 0) {
+            printf("%" PRIx64 "\n", compacted[i]);
+            compactedCount++;
+        }
+    }
+    printf("Compacted to %d indexes.\n", compactedCount);
+
+    int uncompactRes = 10;
+    int uncompactedSize = maxUncompactSize(compacted, inputSize, uncompactRes);
+    H3Index* uncompacted = calloc(uncompactedSize, sizeof(H3Index));
```
isaachier: Hate to nitpick but you could use stack allocation here.

examples/neighbors.c
```diff
@@ -0,0 +1,44 @@
...
+#include <inttypes.h>
+#include <stdio.h>
+
+int main(int argc, char* argv[]) {
+    H3Index indexed = 0x8a2a1072b59ffffL;
+    // Distance away from the origin to find:
+    int k = 2;
+
+    int maxNeighboring = maxKringSize(k);
+    H3Index* neighboring = calloc(maxNeighboring, sizeof(H3Index));
```
isaachier: Same.

examples/CMakeLists.txt
isaacbrodsky: The test wrapper code was moved from ./CMakeLists.txt, I think the semicolon replacement was for a different case there.

examples/compact.c
isaacbrodsky: I originally didn't use stack allocation because that would require including not only the compiler detection code (`../cmake/CheckAlloca.cmake` and `../cmake/CheckVLA.cmake`) but also including the internal header `stackAlloc.h`.

I could refactor the stack allocation code to be self contained in its own directory, but this seems like it produces a simple and easily portable set of examples.

isaachier: Oh right. I forgot that you weren't relying on the existing CMake includes in the main CMake file. No need to change.

examples/CMakeLists.txt
isaachier: OK then never mind.

PR #63, opened 2018-05-30
Optimize the face locating logic

wewei:
Hi, our was investigating the geo-indexing libraries recently. While comparing H3 and S2, we found the index calculation method (geoToH3) is a lot slower than that for S2. The icosahedron projection and the hexagon cells does require more calculation, but it shouldn't like that.

After reading the code, I found the triangle functions are heavily used in H3, while S2 uses 3D points to do the spherical calculation, which can largely reduce the amount of sin/cos calls.

Just logic deciding which face the point is projected to, it make 120 triangle function calls. (For each face, make 3 cos, 2 sin, 1 acos calls to get the great circle distance). Indeed, at most 5 triangle function calls is necessary.

1. The 3D coordinate of the face centers can be pre-calculated.
2. The center of the face the point projected to has the smallest great circle distance(a), thus it has the smallest square distance (S). Because S = (2 * sin(a/2))^2 = 4 * sin^2(a/2) = 2 * (1 - cos(a)), as cos(a) is monotonic decreasing on [0, PI], S is monotonic increasing according to a on [0, PI]. That means, instead of comparing the great circle distance, we can compare the square distance of the 3D points.

With the optimization proposal, we only need to calculate the cos/sin of the lat/lon of the given GeoCoord to translate it into (x, y, z), and calculate acos once to get the great cycle distance for later calculation of the IJK.

I tested the performance with a simple routine,
```c
#include <h3/h3api.h>

int main() {
    for (int lat = -90; lat <= 90; ++lat) {
        for (int lon = 0; lon < 360; ++ lon) {
            GeoCoord coord = {
                .lat = degsToRads(lat),
                .lon = degsToRads(lon),
            };
            for (int res = 0; res < 16; ++ res) {
                geoToH3(&coord, res);
            }
        }
    }
    return 0;
}
```

And here's the result
```
# Baseline
real    0m4.402s
user    0m4.400s
sys     0m0.000s
```

```
# Optimized
real    0m2.121s
user    0m2.119s
sys     0m0.000s
```

### DIFF
--- a/src/h3lib/lib/faceijk.c
+++ b/src/h3lib/lib/faceijk.c
@@ -56,6 +56,29 @@ static const GeoCoord faceCenterGeo[NUM_ICOSA_FACES] = {
     {-1.054751253523952054, 1.794075294689396615},   // face 19
 };
 
+static const Point faceCenterPoint[NUM_ICOSA_FACES] = {
+    {0.2199307791404606, 0.6583691780274996, 0.7198475378926182},     // face  0
+    {-0.2139234834501421, 0.1478171829550703, 0.9656017935214205},    // face  1
+    {0.1092625278784797, -0.4811951572873209, 0.8697775121287253},    // face  2
+    {0.7428567301586791, -0.3593941678278028, 0.5648005936517033},    // face  3
+    {0.8112534709140969, 0.3448953237639384, 0.4721387736413930},     // face  4
+    {-0.1055498149613921, 0.9794457296411413, 0.1718874610009365},    // face  5
+    {-0.8075407579970092, 0.1533552485898819, 0.5695261994882688},    // face  6
+    {-0.2846148069787907, -0.8644080972654206, 0.4144792552473539},   // face  7
+    {0.7405621473854481, -0.6673299564565524, -0.0789837646326737},   // face  8
+    {0.8512303986474293, 0.4722343788582681, -0.2289137388687808},    // face  9
+    {-0.7405621473854481, 0.6673299564565525, 0.0789837646326737},    // face 10
+    {-0.8512303986474292, -0.4722343788582682, 0.2289137388687808},   // face 11
+    {0.1055498149613920, -0.9794457296411413, -0.1718874610009365},   // face 12
+    {0.8075407579970092, -0.1533552485898819, -0.5695261994882688},   // face 13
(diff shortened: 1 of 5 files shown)


sahrk:
Since all of the coordinate systems describe a point, using`Point` in these names seems like too generic a term for the 3D coordinates. Maybe substitute with `xyz`, `pt3D`, or `ECEF` (for earth-centered, earth fixed)?

isaacbrodsky:
> Since all of the coordinate systems describe a point, usingPoint in these names seems like too generic a term for the 3D coordinates. Maybe substitute with xyz, pt3D, or ECEF (for earth-centered, earth fixed)?

Maybe Vec3d to parallel Vec2d?

sahrk:
I agree with `Vec3D`. Originally I discarded that idea because there is no `Vec3D` structure (as there is for `Vec2D`), but upon further reflection I don't think that is a real issue.

nrabinowitz:
Agree with @isaacbrodsky's comments, but this looks great to me otherwise. Thanks for submitting!

nrabinowitz:
FWIW the perf benefit here is reflected in the existing benchmark script as well:

master:
```
Built target benchmarkH3Api
	-- geoToH3: 0.002700 milliseconds per iteration (10000 iterations)
```

PR:
```
Built target benchmarkH3Api
	-- geoToH3: 0.001200 milliseconds per iteration (10000 iterations)
```

wewei:
Thanks for the quick response. I just updated the change accordingly.

For the new `Vec3d` structure, I put it into `geoCoord.h`. However, it sounds better to have a `vec3d.h` in parallel with `vec2d.h`, or merge them into something like `vec.h`? I'll leave the necessary refactoring to you owners, as I don't have confidence to fill the legal terms correctly for a new file.

src/h3lib/lib/faceijk.c
```diff
@@ -359,18 +383,27 @@ void _geoToFaceIjk(const GeoCoord* g, int res, FaceIJK* h) {
  * @param face The icosahedral face containing the spherical coordinates.
  * @param v The 2D hex coordinates of the cell containing the point.
  */
+
```
isaacbrodsky: nit: there is an unneeded extra line

wewei: Yeah, removed it.

PR #64, opened 2018-05-30
Do not require hexRange, hexRanges, and hexRangeDistances in bindings

nrabinowitz:
Does some comment juggling to remove the `hexRange`, `hexRanges`, and `hexRangeDistances` functions from the output of `make binding-functions`. This effectively removes the requirement that bindings implement these functions (end consumers should use `kRing` and `kRingDistances` instead).


isaacbrodsky:
Do we want to mark these functions, or some of them, as deprecated as part of this change?

nrabinowitz:
> Do we want to mark these functions, or some of them, as deprecated as part of this change?

I'd leave that for a later cleanup of the API. My main concern here is just making it clear we don't require them for the bindings, but bindings may choose to implement or make use of them as appropriate - I don't have an opinion on whether they should be actually deprecated.

PR #67, opened 2018-06-01
Move Vec3d to its own files

isaacbrodsky:
Follow up to @wewei's PR #63. Vec3d is moved to a new file, and added a few simple tests of those functions directly.

I added a `generateFaceCenterPoint` program since I like being able to regenerate tables in the library easily. :)


src/apps/testapps/testVec3d.c
```diff
@@ -0,0 +1,66 @@
...
+    t_assert(fabs(_pointSquareDist(&v1, &v3) - 2) < DBL_EPSILON,
+             "distance to <0,1,1> is 2");
+    t_assert(fabs(_pointSquareDist(&v1, &v4) - 3) < DBL_EPSILON,
+             "distance to <1,1,1> is 3");
+    t_assert(fabs(_pointSquareDist(&v1, &v5) - 6) < DBL_EPSILON,
+             "distance to <1,1,2> is 6");
+}
+
+TEST(_geoToVec3d) {
+    Vec3d origin = {0};
```
nrabinowitz: For my info - why not `{0, 0, 0}`? Doesn't this result in undefined memory for the uninitialized struct fields? Or is this functionally equivalent to `{0, 0, 0}`?

src/h3lib/lib/faceijk.c
```diff
@@ -28,12 +28,13 @@
...
 #include "geoCoord.h"
 #include "h3Index.h"
+#include "vec3d.h"
 
 /** square root of 7 */
 #define M_SQRT7 2.6457513110645905905016157536392604257102L
 
 /** @brief icosahedron face centers in lat/lon radians */
-static const GeoCoord faceCenterGeo[NUM_ICOSA_FACES] = {
+const GeoCoord faceCenterGeo[NUM_ICOSA_FACES] = {
```
nrabinowitz: For my info: Why drop `static`?

src/apps/miscapps/generateFaceCenterPoint.c
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Generates and prints the faceCenterPoint table.
+ */
+static void generate() {
+    printf("static const Vec3d faceCenterPoint[NUM_ICOSA_FACES] = {\n");
+    for (int i = 0; i < NUM_ICOSA_FACES; i++) {
+        GeoCoord centerCoords = faceCenterGeo[i];
+        Vec3d centerPoint;
+        _geoToVec3d(&centerCoords, &centerPoint);
+        printf("{%.16f, %.16f, %.16f}, // face %2d\n", centerPoint.x,
```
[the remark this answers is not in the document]
isaacbrodsky: I just left them for clang-format to insert. I see in the base cell neighbors generator it does have the spaces, I'll add them here for consistency.

src/apps/miscapps/generateFaceCenterPoint.c
```diff
@@ -0,0 +1,49 @@
...
+               centerPoint.y, centerPoint.z, i);
+    }
+    printf("};\n");
+}
+
+int main(int argc, char* argv[]) {
+    // check command line args
+    if (argc > 1) {
```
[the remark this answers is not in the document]
isaacbrodsky: It just checks that no (unused) args were provided.

src/apps/testapps/testVec3d.c
isaacbrodsky: I understand this should be the same as initializing all members to 0, https://stackoverflow.com/questions/11152160/initializing-a-struct-to-0

src/h3lib/lib/faceijk.c
isaacbrodsky: Static is incompatible with extern, because a static variable would only be visible within the same compilation unit (think .c file) and extern is needed for the generator.

nrabinowitz:
Thanks for the explanations, @isaacbrodsky!

src/apps/miscapps/generateFaceCenterPoint.c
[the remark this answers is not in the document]
isaachier: Same comment as other review (#69), you can always do `int main(void)` in this case. No need to be "strict" about it.

src/h3lib/lib/faceijk.c
isaachier: You don't need `extern` here if this is declared `extern` in a header file, which is the case here. I'd remove the `extern` here.

isaachier: My mistake. Looks good.

PR #69, opened 2018-06-04
Add the formula for number of indexes per resolution

isaacbrodsky:
This seems a little inefficient to put into its own file, but I'll put it there for the moment and later refactor the generators. Especially so the generators do not depend on the files whose contents they are partly generating (perhaps have a generated tables file, and a base tables/constants file it's based off of.)


src/apps/miscapps/generateNumHexagons.c
```diff
@@ -0,0 +1,55 @@
...
+    int64_t count = NUM_BASE_CELLS;
+    for (int i = 0; i <= MAX_H3_RES; i++) {
+        printf("    %" PRId64 "L,\n", count);
+        // Each hexagon has 7 children, and each pentagon has 6 children
+        count = ((count - NUM_PENTAGONS) * 7) + (NUM_PENTAGONS * 6);
+    }
+    printf("};\n");
+}
+
+int main(int argc, char* argv[]) {
```
isaachier: Alternatively, you could change this to `int main(void)` and ignore the arguments entirely.

nrabinowitz: +1

isaacbrodsky: If we make that change, I think it should be made to all the filters.

PR #70, opened 2018-06-04
Make warnings errors

isaacbrodsky:
I think we're already at or near 0 warnings, so we could make warnings errors to preserve that. It's also possible some of the compilers in CI have warnings which we don't notice when developing.

Thoughts?


cmake/toolchain.cmake
```diff
@@ -18,7 +18,7 @@ set(CMAKE_C_STANDARD 11)
 if(NOT WIN32)
     # Compiler options are set only on non-Windows, since these options
     # are not correct for MSVC.
-    set(CMAKE_C_FLAGS_INIT "-Wall")
+    set(CMAKE_C_FLAGS_INIT "-Wall -Werror")
     string(CONCAT CMAKE_C_FLAGS_DEBUG_INIT
            "-g -gdwarf-2 -g3 -O0 -fno-inline -fno-eliminate-unused-debug-types "
            "--coverage")
```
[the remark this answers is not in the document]
isaacbrodsky: At least a few of these were added by @dfellis in #8 (1bfa8631). Not sure about some of the others.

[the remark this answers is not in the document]
isaachier: OK thanks. At some point it might be worth revisiting the flags here. I am not sure if they are the best options and if there is a better way to use the toolchain file.

isaacbrodsky:
Made an update to this PR, warnings as errors are still defaulted on, but also has an option to disable.

CMakeLists.txt
```diff
@@ -47,6 +47,14 @@ set(H3_SOVERSION 1)
 
 project(h3 VERSION ${H3_VERSION})
 
+set(H3_FLAGS "")
+if(NOT WIN32)
+    option(WARNINGS_AS_ERRORS "Warnings are treated as errors" ON)
```
isaachier: 👍

CMakeLists.txt
```diff
@@ -47,6 +47,14 @@ set(H3_SOVERSION 1)
 
 project(h3 VERSION ${H3_VERSION})
 
+set(H3_FLAGS "")
```
isaachier: Where are the other flags? I'd remove them from `cmake/toolchain.cmake` and put them here instead.

isaacbrodsky: Seems reasonable. I've moved the other flags here.

CMakeLists.txt
```diff
@@ -47,6 +47,27 @@ set(H3_SOVERSION 1)
...
+if(NOT WIN32)
+    # Compiler options are set only on non-Windows, since these options
+    # are not correct for MSVC.
+    string(CONCAT H3_FLAGS ${H3_FLAGS} "-Wall ")
+
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
```
isaachier: So I once thought this was a good idea too. Turns out, checking for "Debug" build might be done at build time instead of configuration time, so you need to use a generator expression here. Something like this should do it:

```
string(CONCAT H3_FLAGS "$<$<CONFIG:Debug>:-g -gdwarf-2 -g3 -O0 -fno-inline -fno-eliminate-unused-debug-types> ")
```

More on this topic here: https://stackoverflow.com/a/24470998/1930331.

For the record, I'm not convinced it's a good idea to use `-O0` or `-g` in these options because CMake will add them as the default debug flags.

isaachier: To be clear, please remove the if statement entirely and you can append this generator expression to the earlier `-Wall` if you'd like.

isaachier:
I see some of the tests are failing. I've had this issue too, and it has to do with `target_compile_options` expecting a list instead of a string. This is actually a good thing because now you no longer have to worry about spaces and string concatenation. Just unquote the options above so everything works (for generator expression, just do `$<$<CONFIG:Debug>:...>` with no wrapping quotes).

CMakeLists.txt
```diff
@@ -47,6 +49,30 @@ set(H3_SOVERSION 1)
...
+set(H3_COMPILE_FLAGS "")
+set(H3_LINK_FLAGS "")
+if(NOT WIN32)
+    # Compiler options are set only on non-Windows, since these options
+    # are not correct for MSVC.
+    list(APPEND H3_COMPILE_FLAGS -Wall)
+
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        list(APPEND H3_COMPILE_FLAGS
+                -g -gdwarf-2 -g3 -O0 -fno-inline -fno-eliminate-unused-debug-types)
```
isaachier: What about the generator expression?

isaacbrodsky:
There was also a second issue that --coverage was no longer passed to the linker, which required adding an additional option there.

CMakeLists.txt
isaacbrodsky: Haven't been pushed yet, still running some tests locally

CMakeLists.txt
```diff
@@ -47,6 +49,30 @@ set(H3_SOVERSION 1)
...
+        if(ENABLE_COVERAGE)
+            list(APPEND H3_COMPILE_FLAGS --coverage)
+            # --coverage is not passed to the linker, so this option is needed
+            # to fully enable coverage.
+            list(APPEND H3_LINK_FLAGS -fprofile-arcs)
```
isaachier: I think `--coverage` would work here too (at least for GCC and clang). But you are right for adding the linker flags in general.

CMakeLists.txt
```diff
@@ -47,6 +49,28 @@ set(H3_SOVERSION 1)
...
 project(h3 VERSION ${H3_VERSION})
 
+set(H3_COMPILE_FLAGS "")
+set(H3_LINK_FLAGS "")
+if(NOT WIN32)
+    # Compiler options are set only on non-Windows, since these options
+    # are not correct for MSVC.
+    list(APPEND H3_COMPILE_FLAGS -Wall)
+
+    list(APPEND H3_COMPILE_FLAGS $<$<CONFIG:Debug>:-g -gdwarf-2 -g3 -O0 -fno-inline -fno-eliminate-unused-debug-types>)
```
isaachier: 👍

CMakeLists.txt
```diff
@@ -147,6 +171,9 @@ set(ALL_SOURCE_FILES
 # Build the H3 library
 add_library(h3 ${LIB_SOURCE_FILES})
 
+target_compile_options(h3 PRIVATE ${H3_COMPILE_FLAGS})
+target_link_libraries(h3 PUBLIC ${H3_LINK_FLAGS})
```
isaachier: Why is this `PUBLIC`?

CMakeLists.txt
isaachier: Please remove `-O0` and `-g`. They don't serve the same purpose here as in the toolchain file.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The public API of this library consists of the functions declared in file
 - Generator for the faceCenterPoint table (#67)
 ### Changed
 - Moved Vec3d structure to `vec3d.h` (#67)
+- Added CMake `WARNINGS_AS_ERRORS` option, defauled on, for Clang and GCC (#70)
```
isaachier: Typo, should be `defaulted on`.

CMakeLists.txt
isaacbrodsky: I was thinking of writing this in a different way but didn't update, will fix it.

CMakeLists.txt
isaacbrodsky: I see `-g` is in the default flags for Debug, but I don't see `-O0`. What is the reason for removing that option?

Issue #73, opened 2018-06-06
The unidirectional edge getOrigin / getDestination functions should check validity


nrabinowitz:
At present, the functions `getOriginH3IndexFromUnidirectionalEdge` and `getDestinationH3IndexFromUnidirectionalEdge` assume a valid edge. This leads to some weird behavior when non-edges are passed in:
```
getOriginH3IndexFromUnidirectionalEdge('891ea6d6533ffff'); // '891ea6d6533ffff'
getOriginH3IndexFromUnidirectionalEdge(0); // '800000000000000'
```
I don't think we need to call `h3UnidirectionalEdgeIsValid`, since it might(?) have more performance implications, but perhaps adding a bit check like 
```
if (H3_GET_MODE(edge) != H3_UNIEDGE_MODE) {
  return H3_INVALID_INDEX;
}
```
would catch some common bad cases?

isaachier:
What are the quotes?

isaachier:
Anyway, I think throughout the library, you should be using better error checking where possible. As of now, my biggest concern is still `calloc` returning `NULL` when OOM occurs. If you are really concerned about performance, you can always follow this classic, ugly C pattern:

```
int func(int argc, const char* argv[])
{
#ifndef NDEBUG  // If not not debugging... AKA in debug mode
    const int returnCode = checkTheAssumptions(args);
    if (returnCode != 0) {
        return returnCode;
    }
#endif  // NDEBUG
...
    return 0;
}
```

nrabinowitz:
> What are the quotes?

Sorry, I'm working on the JS bindings, so this was copied from there. It still holds in C.

nrabinowitz:
Fixed in #74

PR #74, opened 2018-06-06
Validate input when getting the origin or destination of an edge

nrabinowitz:
Per #73, adding validation for `getOriginH3IndexFromUnidirectionalEdge` and `getDestinationH3IndexFromUnidirectionalEdge` to avoid unexpected results for bad input.

I've said this before, but these are really verbosely named functions. Couldn't we have gone with `getUnidirectionalEdgeOrigin`?

### DIFF
--- a/src/h3lib/lib/h3UniEdge.c
+++ b/src/h3lib/lib/h3UniEdge.c
@@ -125,6 +125,9 @@ H3Index H3_EXPORT(getH3UnidirectionalEdge)(H3Index origin,
  * @return The origin H3 hexagon index
  */
 H3Index H3_EXPORT(getOriginH3IndexFromUnidirectionalEdge)(H3Index edge) {
+    if (H3_GET_MODE(edge) != H3_UNIEDGE_MODE) {
+        return 0;
+    }
     H3Index origin = edge;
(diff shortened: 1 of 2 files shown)


isaachier:
C tends to be verbose, but usually just due to the lack of scoping and lack of function overloading. This sort of sounds like the overloading case, but IDK. Either way, I have seen worse.

src/h3lib/lib/h3UniEdge.c
```diff
@@ -125,6 +125,9 @@ H3Index H3_EXPORT(getH3UnidirectionalEdge)(H3Index origin,
  * @return The origin H3 hexagon index
  */
 H3Index H3_EXPORT(getOriginH3IndexFromUnidirectionalEdge)(H3Index edge) {
+    if (H3_GET_MODE(edge) != H3_UNIEDGE_MODE) {
+        return 0;
```
isaacbrodsky: `return H3_INVALID_INDEX;`

nrabinowitz: Whoops, could have sworn I did that.

nrabinowitz: Updated.

PR #77, opened 2018-06-11
Switch `int` digit to an enum, `Direction`

isaacbrodsky:
Attempt to clarify when a direction digit is expected and when it is not. C permits arithmetic and assignment of ints/enums so this is not able to catch many issues at compile time.

The neighbor traversal tables in `algos.c` (like NEW_DIGIT_II) could also be updated to use the constants.

Also update copyright headers and replace 0 with H3_INVALID_INDEX when applicable.


src/h3lib/lib/algos.c
```diff
@@ -222,32 +223,32 @@ H3Index h3NeighborRotations(H3Index origin, int dir, int* rotations) {
         } else {
             // generated by hand
             // Current digit -> direction -> new digit
-            const int NEW_DIGIT_II[7][7] = {
+            const Direction NEW_DIGIT_II[7][7] = {
                 {0, 1, 2, 3, 4, 5, 6}, {1, 4, 3, 6, 5, 2, 0},
                 {2, 3, 1, 4, 6, 0, 5}, {3, 6, 4, 5, 0, 1, 2},
                 {4, 5, 6, 0, 2, 3, 1}, {5, 2, 0, 1, 3, 6, 4},
                 {6, 0, 5, 2, 1, 4, 3}};
```
isaachier: Is this generated code? It would have been nice to see this as `enum` members instead of raw integers, but I understand this is too much to ask for all of these values.

src/h3lib/include/coordijk.h
```diff
@@ -54,22 +54,27 @@ static const CoordIJK UNIT_VECS[] = {
...
+ */
+typedef enum {
+    /** H3 digit in center */
+    CENTER_DIGIT = 0,
+    /** H3 digit in k-axes direction */
+    K_AXES_DIGIT = 1,
+    /** H3 digit in j-axes direction */
+    J_AXES_DIGIT = 2,
+    /** H3 digit in j == k direction */
+    JK_AXES_DIGIT = K_AXES_DIGIT | J_AXES_DIGIT, /* 3 */
```
joegilley: tiniest nit ever: swap j & k OR to `J_AXES_DIGIT | K_AXES_DIGIT` s.t. `JK_... = J_... | K_...` :P

src/apps/miscapps/generateBaseCellNeighbors.c
```diff
@@ -175,15 +175,15 @@ static void generate() {
                         if (i == 4 || i == 117) {
                             // 'polar' type pentagon with all faces pointing
                             // towards i
-                            if (dir == 5) {
+                            if (dir == IK_AXES_DIGIT) {
                                 rotAdj = 2;
```
nrabinowitz: Can these `rotAdj` assignments get moved to constants as well?

src/h3lib/lib/algos.c
```diff
@@ -118,7 +119,7 @@ void H3_EXPORT(kRingDistances)(H3Index origin, int k, H3Index* out,
         // Fast algo failed, fall back to slower, correct algo
         // and also wipe out array because contents untrustworthy
         for (int i = 0; i < maxIdx; i++) {
-            out[i] = 0;
+            out[i] = H3_INVALID_INDEX;
             distances[i] = 0;
```
nrabinowitz: Not required, but it's worth noting that `0` is actually a valid value in `distances` - maybe worth defining `INVALID_DISTANCE = -1` or similar?

src/h3lib/lib/coordijk.c
```diff
@@ -354,8 +349,8 @@ void _downAp7r(CoordIJK* ijk) {
  * @param ijk The ijk coordinates.
  * @param digit The digit direction from the original ijk coordinates.
  */
-void _neighbor(CoordIJK* ijk, int digit) {
-    if (digit != 0) {
+void _neighbor(CoordIJK* ijk, Direction digit) {
```
nrabinowitz: Nit: Should this argument be `dir` now? It seems like we're a little inconsistent still about whether we call this a "direction" (the semantic name) or a "digit" (the concrete implementation).

src/h3lib/lib/coordijk.c
```diff
@@ -253,12 +248,12 @@ void _ijkNormalize(CoordIJK* c) {
...
  */
-int _unitIjkToDigit(const CoordIJK* ijk) {
+Direction _unitIjkToDigit(const CoordIJK* ijk) {
     CoordIJK c = *ijk;
     _ijkNormalize(&c);
 
-    int digit = INVALID_DIGIT;
-    for (int i = 0; i < 7; i++) {
+    Direction digit = INVALID_DIGIT;
+    for (Direction i = CENTER_DIGIT; i < INVALID_DIGIT; i++) {
```
nrabinowitz: Nit: This makes me think `INVALID_DIGIT` is the wrong name. I'd expect `INVALID_X` to be an opaque value, but here you have to know what the value is (or at least that it's an upper bound). Maybe `OUT_OF_RANGE_DIGIT`? Or define `MAX_DIGIT = IJ_AXES_DIGIT` and use `MAX_DIGIT + 1` here?

src/h3lib/lib/h3Index.c
```diff
@@ -446,12 +446,12 @@ int H3_EXPORT(h3IsPentagon)(H3Index h) {
  * @param h The H3Index.
  * @return The highest resolution non-zero digit in the H3Index.
  */
-int _h3LeadingNonZeroDigit(H3Index h) {
+Direction _h3LeadingNonZeroDigit(H3Index h) {
     for (int r = 1; r <= H3_GET_RESOLUTION(h); r++)
         if (H3_GET_INDEX_DIGIT(h, r)) return H3_GET_INDEX_DIGIT(h, r);
```
nrabinowitz: Nit: As above, this requires knowledge that the center digit is 0. Maybe slightly more semantic to rewrite this as `_h3LeadingNonCenterDigit` and compare `if (H3_GET_INDEX_DIGIT(h, r) !== CENTER_DIGIT)`? No strong feeling though, it might not be possible/desirable to fully abstract away the concrete implementation.

src/h3lib/lib/algos.c
isaacbrodsky: These tables were written by hand. I'll probably update the PR to make them constants as well, although I'm not sure that'll help readability.

isaacbrodsky:
Rebased on master and addressed comments

src/h3lib/lib/coordijk.c
isaacbrodsky: END_<Something> or LAST_<Something> seem like common patterns in C, I can update to use that instead of (or perhaps as an alias to) INVALID_DIGIT.

src/h3lib/lib/coordijk.c
isaacbrodsky: Even worse, the directions are themselves called "digits"

src/h3lib/lib/coordijk.c
nrabinowitz: Makes sense - I think if you use `LAST_DIGIT` you'd need to use `LAST_DIGIT + 1` to indicate the out-of-bounds value, since `LAST_DIGIT` sounds valid.

isaachier: I've actually seen most frequently `NUM_DIRECTIONS` or the equivalent.

PR #78, opened 2018-06-12
Fix for edge directions in getH3UnidirectionalEdgeBoundary

### LINKED ISSUE #75: Edge direction in getH3UnidirectionalEdgeBoundary is inconsistent
This may not be a big deal for most consumers, but the algorithm used for `getH3UnidirectionalEdgeBoundary` produces 5 edges that follow the direction of the origin's geo boundary output, and one edge that goes in the opposite direction (from the origin's vertex 5 to vertex 0).

Given that we might rewrite that algorithm to something more efficient in any case, it would be a useful requirement for the output that all edges follow consistent directionality.

nrabinowitz:
Fixes #75. This isn't the most elegant fix, but it does get the vertices into the correct order.

Longer term, I think the best option here is to do what we do in `h3SetToVertexGraph`, getting the edges and then walking them. But at the moment this was a bit too clunky to implement; it might be easier after a refactor of the `VertexGraph` structure.

### DIFF
--- a/src/apps/testapps/testH3UniEdge.c
+++ b/src/apps/testapps/testH3UniEdge.c
@@ -213,6 +213,8 @@ TEST(getH3UnidirectionalEdgeBoundary) {
     const int expectedVertices[][2] = {{3, 4}, {1, 2}, {2, 3},
                                  {5, 0}, {4, 5}, {0, 1}};
 
+    // TODO: The current implementation relies on lat/lon comparison and fails on
+    // resolutions finer than 12
     for (int res = 0; res < 13; res++) {
         sf = H3_EXPORT(geoToH3)(&sfGeo, res);
         H3_EXPORT(h3ToGeoBoundary)(sf, &boundary);
@@ -241,6 +243,8 @@ TEST(getH3UnidirectionalEdgeBoundaryPentagonClassIII) {
     const int expectedVertices[][3] = {{-1, -1, -1}, {2, 3, 4}, {4, 5, 6},
                                  {8, 9, 0},    {6, 7, 8}, {0, 1, 2}};
 
+    // TODO: The current implementation relies on lat/lon comparison and fails on
+    // resolutions finer than 12
     for (int res = 1; res < 13; res += 2) {
         setH3Index(&pentagon, res, 24, 0);
         H3_EXPORT(h3ToGeoBoundary)(pentagon, &boundary);
@@ -279,6 +283,8 @@ TEST(getH3UnidirectionalEdgeBoundaryPentagonClassII) {
     const int expectedVertices[][3] = {{-1, -1}, {1, 2}, {2, 3},
                                  {4, 0},   {3, 4}, {0, 1}};
 
+    // TODO: The current implementation relies on lat/lon comparison and fails on
+    // resolutions finer than 12
     for (int res = 0; res < 12; res += 2) {
         setH3Index(&pentagon, res, 24, 0);
         H3_EXPORT(h3ToGeoBoundary)(pentagon, &boundary);
--- a/src/apps/testapps/testH3UniEdge.c
+++ b/src/apps/testapps/testH3UniEdge.c
@@ -210,7 +210,7 @@ TEST(getH3UnidirectionalEdgeBoundary) {
     GeoBoundary edgeBoundary;
     STACK_ARRAY_CALLOC(H3Index, edges, 6);
 
-    int expectedVertices[][2] = {{3, 4}, {1, 2}, {2, 3},
+    const int expectedVertices[][2] = {{3, 4}, {1, 2}, {2, 3},
                                  {5, 0}, {4, 5}, {0, 1}};
 
     for (int res = 0; res < 13; res++) {
@@ -238,7 +238,7 @@ TEST(getH3UnidirectionalEdgeBoundaryPentagonClassIII) {
     GeoBoundary edgeBoundary;
     STACK_ARRAY_CALLOC(H3Index, edges, 6);
 
-    int expectedVertices[][3] = {{-1, -1, -1}, {2, 3, 4}, {4, 5, 6},
+    const int expectedVertices[][3] = {{-1, -1, -1}, {2, 3, 4}, {4, 5, 6},
                                  {8, 9, 0},    {6, 7, 8}, {0, 1, 2}};
 
     for (int res = 1; res < 13; res += 2) {
@@ -276,7 +276,7 @@ TEST(getH3UnidirectionalEdgeBoundaryPentagonClassII) {
     GeoBoundary edgeBoundary;
     STACK_ARRAY_CALLOC(H3Index, edges, 6);
 
-    int expectedVertices[][3] = {{-1, -1}, {1, 2}, {2, 3},
+    const int expectedVertices[][3] = {{-1, -1}, {1, 2}, {2, 3},
                                  {4, 0},   {3, 4}, {0, 1}};
 
     for (int res = 0; res < 12; res += 2) {
--- a/src/apps/testapps/testH3UniEdge.c
+++ b/src/apps/testapps/testH3UniEdge.c
@@ -211,10 +211,10 @@ TEST(getH3UnidirectionalEdgeBoundary) {
     STACK_ARRAY_CALLOC(H3Index, edges, 6);
 
     const int expectedVertices[][2] = {{3, 4}, {1, 2}, {2, 3},
-                                 {5, 0}, {4, 5}, {0, 1}};
+                                       {5, 0}, {4, 5}, {0, 1}};
 
-    // TODO: The current implementation relies on lat/lon comparison and fails on
-    // resolutions finer than 12
+    // TODO: The current implementation relies on lat/lon comparison and fails
+    // on resolutions finer than 12
     for (int res = 0; res < 13; res++) {
         sf = H3_EXPORT(geoToH3)(&sfGeo, res);
         H3_EXPORT(h3ToGeoBoundary)(sf, &boundary);
@@ -241,10 +241,10 @@ TEST(getH3UnidirectionalEdgeBoundaryPentagonClassIII) {
     STACK_ARRAY_CALLOC(H3Index, edges, 6);
 
     const int expectedVertices[][3] = {{-1, -1, -1}, {2, 3, 4}, {4, 5, 6},
-                                 {8, 9, 0},    {6, 7, 8}, {0, 1, 2}};
+                                       {8, 9, 0},    {6, 7, 8}, {0, 1, 2}};
 
-    // TODO: The current implementation relies on lat/lon comparison and fails on
-    // resolutions finer than 12
+    // TODO: The current implementation relies on lat/lon comparison and fails
+    // on resolutions finer than 12
     for (int res = 1; res < 13; res += 2) {
         setH3Index(&pentagon, res, 24, 0);
         H3_EXPORT(h3ToGeoBoundary)(pentagon, &boundary);
@@ -281,10 +281,10 @@ TEST(getH3UnidirectionalEdgeBoundaryPentagonClassII) {
     STACK_ARRAY_CALLOC(H3Index, edges, 6);
 
     const int expectedVertices[][3] = {{-1, -1}, {1, 2}, {2, 3},
-                                 {4, 0},   {3, 4}, {0, 1}};
+                                       {4, 0},   {3, 4}, {0, 1}};
 
-    // TODO: The current implementation relies on lat/lon comparison and fails on
-    // resolutions finer than 12
+    // TODO: The current implementation relies on lat/lon comparison and fails
+    // on resolutions finer than 12
     for (int res = 0; res < 12; res += 2) {
         setH3Index(&pentagon, res, 24, 0);
         H3_EXPORT(h3ToGeoBoundary)(pentagon, &boundary);
--- a/src/apps/testapps/testH3UniEdge.c
+++ b/src/apps/testapps/testH3UniEdge.c
@@ -23,6 +23,7 @@
 #include "constants.h"
 #include "geoCoord.h"
 #include "h3Index.h"
+#include "stackAlloc.h"
 #include "test.h"
 
 // Fixtures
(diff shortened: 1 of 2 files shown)


src/apps/testapps/testH3UniEdge.c
```diff
@@ -208,34 +209,113 @@ TEST(getH3UnidirectionalEdgesFromHexagon) {
...
-    H3_EXPORT(getH3UnidirectionalEdgesFromHexagon)(sf, edges);
 
-    GeoBoundary gb;
-    for (int i = 0; i < 6; i++) {
-        H3_EXPORT(getH3UnidirectionalEdgeBoundary)(edges[i], &gb);
-        t_assert(gb.numVerts == 2, "Got the expected number of vertices back");
+    int expectedVertices[][2] = {{3, 4}, {1, 2}, {2, 3},
+                                 {5, 0}, {4, 5}, {0, 1}};
+
+    for (int res = 0; res < 13; res++) {
```
nrabinowitz: Note that the current approach fails at res > 12 because of the geo comparison precision.

src/h3lib/lib/h3UniEdge.c
```diff
@@ -212,6 +213,15 @@ void H3_EXPORT(getH3UnidirectionalEdgesFromHexagon)(H3Index origin,
     }
 }
 
+bool _hasMatchingVertex(GeoCoord* vertex, GeoBoundary* boundary) {
```
isaacbrodsky: These arguments should be const.

src/h3lib/lib/h3UniEdge.c
```diff
@@ -229,14 +241,26 @@ void H3_EXPORT(getH3UnidirectionalEdgeBoundary)(H3Index edge, GeoBoundary* gb) {
...
-            if (geoAlmostEqualThreshold(&origin.verts[i], &destination.verts[j],
-                                        0.000001)) {
-                gb->verts[k].lat = origin.verts[i].lat;
-                gb->verts[k].lon = origin.verts[i].lon;
+        if (_hasMatchingVertex(&origin.verts[i], &destination)) {
+            // If we are on vertex 0, we need to handle the case where it's the
+            // end of the edge, not the beginning.
+            // TODO: This is fine for this limited case, but it might be
+            // better/cleaner to use a VertexGraph for this
+            printf("%d, %d\n", i,
```
isaacbrodsky: Please delete debugging statement

src/h3lib/lib/h3UniEdge.c
```diff
@@ -229,14 +241,26 @@ void H3_EXPORT(getH3UnidirectionalEdgeBoundary)(H3Index edge, GeoBoundary* gb) {
...
+                !_hasMatchingVertex(&origin.verts[i + 1], &destination)) {
+                postponedVertex = origin.verts[i];
+            } else {
+                gb->verts[k] = origin.verts[i];
                 k++;
             }
         }
     }
+    // If we postponed adding the last vertex, add it now
+    if (postponedVertex.lat && postponedVertex.lon) {
```
isaacbrodsky: This doesn't look safe, since (0,0) is a valid set of coordinates.

src/h3lib/lib/h3UniEdge.c
nrabinowitz: Will update.

src/h3lib/lib/h3UniEdge.c
nrabinowitz: Oh, sorry. Will delete.

src/h3lib/lib/h3UniEdge.c
nrabinowitz: True. Is there a better option here? The only thing I can think of is to add an additional var like `hasPostponedVertex`, which is pretty ugly.

src/apps/testapps/testH3UniEdge.c
isaachier: Might be worth adding as a comment.

src/apps/testapps/testH3UniEdge.c
```diff
@@ -208,34 +209,113 @@ TEST(getH3UnidirectionalEdgesFromHexagon) {
 }
 
 TEST(getH3UnidirectionalEdgeBoundary) {
-    H3Index sf = H3_EXPORT(geoToH3)(&sfGeo, 9);
+    H3Index sf;
+    GeoBoundary boundary;
+    GeoBoundary edgeBoundary;
     H3Index* edges = calloc(6, sizeof(H3Index));
```
isaachier: Somewhat outside of your changeset, but why is `calloc` here instead of `STACK_ARRAY_CALLOC` (https://github.com/uber/h3/blob/master/src/h3lib/include/stackAlloc.h#L29)?

src/apps/testapps/testH3UniEdge.c
```diff
@@ -208,34 +209,113 @@ TEST(getH3UnidirectionalEdgesFromHexagon) {
...
-    H3_EXPORT(getH3UnidirectionalEdgesFromHexagon)(sf, edges);
 
-    GeoBoundary gb;
-    for (int i = 0; i < 6; i++) {
-        H3_EXPORT(getH3UnidirectionalEdgeBoundary)(edges[i], &gb);
-        t_assert(gb.numVerts == 2, "Got the expected number of vertices back");
+    int expectedVertices[][2] = {{3, 4}, {1, 2}, {2, 3},
+                                 {5, 0}, {4, 5}, {0, 1}};
```
isaachier: `const`?

src/apps/testapps/testH3UniEdge.c
```diff
@@ -208,34 +209,113 @@ TEST(getH3UnidirectionalEdgesFromHexagon) {
...
-    int missingEdgeCount = 0;
-    for (int i = 0; i < 6; i++) {
-        if (edges[i] == 0) {
-            missingEdgeCount++;
-        } else {
-            H3_EXPORT(getH3UnidirectionalEdgeBoundary)(edges[i], &gb);
-            t_assert(gb.numVerts == 3,
-                     "Got the expected number of vertices back for a Class III "
-                     "pentagon");
+    int expectedVertices[][3] = {{-1, -1, -1}, {2, 3, 4}, {4, 5, 6},
```
isaachier: `const`?

src/apps/testapps/testH3UniEdge.c
```diff
@@ -208,34 +209,113 @@ TEST(getH3UnidirectionalEdgesFromHexagon) {
...
+}
 
-    H3Index pentagon = 0x811c0ffffffffff;
-    edges = calloc(6, sizeof(H3Index));
-    H3_EXPORT(getH3UnidirectionalEdgesFromHexagon)(pentagon, edges);
+TEST(getH3UnidirectionalEdgeBoundaryPentagonClassIII) {
+    H3Index pentagon;
+    GeoBoundary boundary;
+    GeoBoundary edgeBoundary;
+    H3Index* edges = calloc(6, sizeof(H3Index));
```
isaachier: Same stack allocation question.

src/apps/testapps/testH3UniEdge.c
```diff
@@ -208,34 +209,113 @@ TEST(getH3UnidirectionalEdgesFromHexagon) {
...
+        H3_EXPORT(getH3UnidirectionalEdgesFromHexagon)(sf, edges);
+
+        for (int i = 0; i < 6; i++) {
+            H3_EXPORT(getH3UnidirectionalEdgeBoundary)(edges[i], &edgeBoundary);
+            t_assert(edgeBoundary.numVerts == 2,
+                     "Got the expected number of vertices back");
+            for (int j = 0; j < edgeBoundary.numVerts; j++) {
+                t_assert(
+                    geoAlmostEqual(&edgeBoundary.verts[j],
+                                   &boundary.verts[expectedVertices[i][j]]),
```
isaachier: I'm too tired to figure this out at the moment. Looks fun :).

src/h3lib/lib/h3UniEdge.c
```diff
@@ -212,6 +213,15 @@ void H3_EXPORT(getH3UnidirectionalEdgesFromHexagon)(H3Index origin,
     }
 }
 
+bool _hasMatchingVertex(const GeoCoord* vertex, const GeoBoundary* boundary) {
```
isaachier: `static`?

src/h3lib/lib/h3UniEdge.c
```diff
@@ -229,14 +242,25 @@ void H3_EXPORT(getH3UnidirectionalEdgeBoundary)(H3Index edge, GeoBoundary* gb) {
...
-        for (int j = 0; j < destination.numVerts; j++) {
-            if (geoAlmostEqualThreshold(&origin.verts[i], &destination.verts[j],
-                                        0.000001)) {
-                gb->verts[k].lat = origin.verts[i].lat;
-                gb->verts[k].lon = origin.verts[i].lon;
+        if (_hasMatchingVertex(&origin.verts[i], &destination)) {
+            // If we are on vertex 0, we need to handle the case where it's the
+            // end of the edge, not the beginning.
+            // TODO: This is fine for this limited case, but it might be
+            // better/cleaner to use a VertexGraph for this
```
isaachier: You are better off avoiding allocations involved in creating a `VertexGraph` and hacking a solution with only the given inputs.

src/h3lib/lib/h3UniEdge.c
```diff
@@ -229,14 +242,25 @@ void H3_EXPORT(getH3UnidirectionalEdgeBoundary)(H3Index edge, GeoBoundary* gb) {
...
+            if (i == 0 &&
+                !_hasMatchingVertex(&origin.verts[i + 1], &destination)) {
+                postponedVertex = origin.verts[i];
+                hasPostponedVertex = true;
```
isaachier: Is zero a valid vertex? If not, you can just check that `postponedVertex != 0` instead of maintaining a separate flag.

src/h3lib/lib/h3UniEdge.c
```diff
@@ -229,14 +242,25 @@ void H3_EXPORT(getH3UnidirectionalEdgeBoundary)(H3Index edge, GeoBoundary* gb) {
...
         }
     }
+    // If we postponed adding the last vertex, add it now
+    if (hasPostponedVertex) {
+        gb->verts[k] = postponedVertex;
+        k++;
+    }
     gb->numVerts = k;
-}
+}
```
isaachier: Deleted newline from end of file. Put a little marker here to show me.

src/apps/testapps/testH3UniEdge.c
nrabinowitz: Will do.

src/apps/testapps/testH3UniEdge.c
nrabinowitz: Mostly likely because this was written before we introduced that. Will update.

src/apps/testapps/testH3UniEdge.c
nrabinowitz: Will add.

src/apps/testapps/testH3UniEdge.c
nrabinowitz: Will add.

src/h3lib/lib/h3UniEdge.c
nrabinowitz: Yeah, probably true. It just irks me slightly that this is basically the same thing we're doing in `h3SetToVertexGraph` but we're doing it totally differently :/. Also bothers me that this is `O(n^2)`, even though `n` is guaranteed to be <= 10 and is usually 6. I'll drop the comment though, I doubt there'd be any benefit.

src/h3lib/lib/h3UniEdge.c
nrabinowitz: `postponedVertex` is a `GeoCoord` struct, and unfortunately, as @isaacbrodsky points out, it's valid to have all the fields in the struct initialized to `0`. So I don't know of another way to check whether `postponedVertex` has been set or not, hence the flag. Open to suggestions.

src/h3lib/lib/h3UniEdge.c
nrabinowitz: Hm. I'm surprised `clang-format` doesn't have some sort of opinion on that. Will fix.

src/h3lib/lib/h3UniEdge.c
nrabinowitz: Will update.

src/apps/testapps/testH3UniEdge.c
```diff
@@ -32,7 +33,7 @@ BEGIN_TESTS(h3UniEdge);
 
 TEST(h3IndexesAreNeighbors) {
     H3Index sf = H3_EXPORT(geoToH3)(&sfGeo, 9);
-    H3Index* ring = calloc(H3_EXPORT(maxKringSize)(1), sizeof(H3Index));
+    STACK_ARRAY_CALLOC(H3Index, ring, H3_EXPORT(maxKringSize)(1));
```
isaachier: 👍

src/h3lib/lib/h3UniEdge.c
```diff
@@ -213,7 +213,14 @@ void H3_EXPORT(getH3UnidirectionalEdgesFromHexagon)(H3Index origin,
...
 }
 
-bool _hasMatchingVertex(const GeoCoord* vertex, const GeoBoundary* boundary) {
+/**
+ * Whether the given coordinate has a matching vertex in the given geo boundary.
+ * @param  vertex   Coordinate to check
+ * @param  boundary Geo boundary to look in
+ * @return          Whether a match was found
+ */
+static bool _hasMatchingVertex(const GeoCoord* vertex,
```
isaachier: Great!

src/apps/testapps/testH3UniEdge.c
isaachier: Waiting on this when you have a chance.

src/apps/testapps/testH3UniEdge.c
nrabinowitz: Added, that's what I get for trying to multitask.

src/apps/testapps/testH3UniEdge.c
```diff
@@ -196,47 +194,118 @@ TEST(getH3UnidirectionalEdgesFromHexagon) {
...
+TEST(getH3UnidirectionalEdgeBoundaryPentagonClassIII) {
+    H3Index pentagon;
+    GeoBoundary boundary;
+    GeoBoundary edgeBoundary;
+    STACK_ARRAY_CALLOC(H3Index, edges, 6);
+
+    const int expectedVertices[][3] = {{-1, -1, -1}, {2, 3, 4}, {4, 5, 6},
+                                 {8, 9, 0},    {6, 7, 8}, {0, 1, 2}};
+
+    for (int res = 1; res < 13; res += 2) {
```
isaacbrodsky: Please add comments for why we don't test all resolutions.

isaachier:
It seems Travis CI wants you to format the code.

nrabinowitz:
Yeah, having more trouble with this than expected due to laptop theft :(. Should update shortly.

PR #83, opened 2018-06-22
Add h3Distance, and internal h3ToIjk and ijkDistance

isaacbrodsky:
Experimental.

Adds an h3ToIjk function which produces users consumable IJK+ coordinates. An "origin" for the coordinate system must be specified, and coordinates are only comparable if they were obtained from the same origin. This is limited to obtaining coordinates for indexes which are on the same base cell or neighboring base cells. Pentagons have more restrictions about crossing the pentagonal distortion. The intention is in the future this could be improved to handle pentagon distortion better, which may make current coordinates invalid. Coordinates should not be persisted or passed between library versions.

Pentagon cases were mostly restricted by experimentally disabling a number of cases.

With IJK+ coordinates, it's possible to determine the grid distance of cells, with the same limitations of h3ToIjk. This is done under the assumption that the distances that are of interest are below the granularity of a base cell (at that resolution, haversine may be a better approach.)

Along with this change, CoordIJK is moved to the public API.

edit: this would be version 3.1.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ The public API of this library consists of the functions declared in file
 
 ## [Unreleased]
 ### Added
-- h3ToIjk function for getting IJK+ coordinates from an index
-- ijkDistance function for determining the grid distance between IJK+ coordinates
-- h3Distance function for determining the grid distance between H3 indexes
+- `h3Distance` function for determining the grid distance between H3 indexes (#83)
+- Internal `h3ToIjk` function for getting IJK+ coordinates from an index (#83)
+- Internal `ijkDistance` function for determining the grid distance between IJK+ coordinates (#83)
 
 ## [3.0.8] - 2018-07-18
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The public API of this library consists of the functions declared in file
 - `h3Distance` function for determining the grid distance between H3 indexes (#83)
 - Internal `h3ToIjk` function for getting IJK+ coordinates from an index (#83)
 - Internal `ijkDistance` function for determining the grid distance between IJK+ coordinates (#83)
+- `h3ToIjk` filter application for experimenting with `h3ToIjk` (#83)
 
 ## [3.0.8] - 2018-07-18
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The public API of this library consists of the functions declared in file
 [h3api.h](./src/h3lib/include/h3api.h).
 
 ## [Unreleased]
+### Added
+- h3ToIjk function for getting IJK+ coordinates from an index
+- ijkDistance function for determining the grid distance between IJK+ coordinates
+- h3Distance function for determining the grid distance between H3 indexes
 
 ## [3.0.8] - 2018-07-18
 ### Added
--- a/src/apps/applib/include/utility.h
+++ b/src/apps/applib/include/utility.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 Uber Technologies, Inc.
+ * Copyright 2016-2018 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/apps/applib/include/utility.h
+++ b/src/apps/applib/include/utility.h
@@ -42,4 +42,8 @@ void geoBoundaryPrint(const GeoBoundary* b);
 void geoBoundaryPrintln(const GeoBoundary* b);
 int readBoundary(FILE* f, GeoBoundary* b);
 
+void iterateAllIndexesAtRes(int res, void (*callback)(H3Index));
+void iterateAllIndexesAtResPartial(int res, void (*callback)(H3Index),
+                                   int maxBaseCell);
+
 #endif
--- a/src/apps/applib/lib/utility.c
+++ b/src/apps/applib/lib/utility.c
@@ -18,10 +18,12 @@
  */
 
 #include "utility.h"
+#include <assert.h>
 #include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include "geoCoord.h"
+#include "h3Index.h"
 #include "h3api.h"
 
 void error(const char* msg) {
@@ -162,4 +164,36 @@ int readBoundary(FILE* f, GeoBoundary* b) {
     }
 
     return 0;
+}
+
+/**
+ * Call the callback for every index at the given resolution.
+ */
+void iterateAllIndexesAtRes(int res, void (*callback)(H3Index)) {
+    iterateAllIndexesAtResPartial(res, callback, NUM_BASE_CELLS);
+}
+
+/**
+ * Call the callback for every index at the given resolution in base
+ * cell 0 up to the given base cell number.
+ */
+void iterateAllIndexesAtResPartial(int res, void (*callback)(H3Index),
+                                   int baseCells) {
+    assert(baseCells <= NUM_BASE_CELLS);
+    for (int i = 0; i < baseCells; i++) {
+        H3Index bc;
+        setH3Index(&bc, 0, i, 0);
+        int childrenSz = H3_EXPORT(maxUncompactSize)(&bc, 1, res);
+        H3Index* children = calloc(childrenSz, sizeof(H3Index));
+        H3_EXPORT(uncompact)(&bc, 1, children, childrenSz, res);
+
+        for (int j = 0; j < childrenSz; j++) {
+            if (children[j] == 0) {
+                continue;
+            }
+
+            (*callback)(children[j]);
+        }
+        free(children);
+    }
 }
--- a/src/apps/applib/lib/utility.c
+++ b/src/apps/applib/lib/utility.c
@@ -20,6 +20,7 @@
 #include "utility.h"
 #include <assert.h>
 #include <inttypes.h>
+#include <stackAlloc.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include "geoCoord.h"
@@ -184,7 +185,7 @@ void iterateAllIndexesAtResPartial(int res, void (*callback)(H3Index),
         H3Index bc;
         setH3Index(&bc, 0, i, 0);
         int childrenSz = H3_EXPORT(maxUncompactSize)(&bc, 1, res);
-        H3Index* children = calloc(childrenSz, sizeof(H3Index));
+        STACK_ARRAY_CALLOC(H3Index, children, childrenSz);
         H3_EXPORT(uncompact)(&bc, 1, children, childrenSz, res);
 
         for (int j = 0; j < childrenSz; j++) {
@@ -194,6 +195,5 @@ void iterateAllIndexesAtResPartial(int res, void (*callback)(H3Index),
 
             (*callback)(children[j]);
         }
-        free(children);
     }
 }
--- a/src/apps/applib/lib/utility.c
+++ b/src/apps/applib/lib/utility.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 Uber Technologies, Inc.
+ * Copyright 2016-2018 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/apps/testapps/testH3ToIjk.c
+++ b/src/apps/testapps/testH3ToIjk.c
@@ -37,7 +37,7 @@ void h3Distance_identity_assertions(H3Index h3) {
     t_assert(H3_EXPORT(h3Distance)(h3, h3) == 0, "distance to self is 0");
 
     CoordIJK ijk;
-    t_assert(H3_EXPORT(h3ToIjk(h3, h3, &ijk)) == 0, "failed to get ijk");
+    t_assert(h3ToIjk(h3, h3, &ijk) == 0, "failed to get ijk");
     if (r == 0) {
         t_assert(_ijkMatches(&ijk, &UNIT_VECS[0]) == 1, "not at 0,0,0 (res 0)");
     } else if (r == 1) {
@@ -225,9 +225,20 @@ TEST(h3DistanceBaseCells) {
              "distance to neighbor is invalid");
 
     CoordIJK ijk;
-    t_assert(H3_EXPORT(h3ToIjk(pent1, bc1, &ijk)) == 0,
-             "failed to get ijk (4, 15)");
+    t_assert(h3ToIjk(pent1, bc1, &ijk) == 0, "failed to get ijk (4, 15)");
     t_assert(_ijkMatches(&ijk, &UNIT_VECS[2]) == 1, "not at 0,1,0");
+}
+
+TEST(h3DistanceFailed) {
+    H3Index h3 = 0x832830fffffffffL;
+    H3Index edge = H3_EXPORT(getH3UnidirectionalEdge(h3, 0x832834fffffffffL));
+    H3Index h3res2 = 0x822837fffffffffL;
+
+    t_assert(H3_EXPORT(h3Distance)(edge, h3) == -1, "edges cannot be origins");
+    t_assert(H3_EXPORT(h3Distance)(h3, edge) == -1,
+             "edges cannot be destinations");
+    t_assert(H3_EXPORT(h3Distance)(h3, h3res2) == -1,
+             "cannot compare at different resolutions");
 }
 
 END_TESTS();
--- a/src/apps/testapps/testH3ToIjk.c
+++ b/src/apps/testapps/testH3ToIjk.c
@@ -55,7 +55,7 @@ void h3Distance_identity_assertions(H3Index h3) {
 
 void h3Distance_neighbors_assertions(H3Index h3) {
     CoordIJK origin = {0};
-    t_assert(H3_EXPORT(h3ToIjk)(h3, h3, &origin) == 0, "got ijk for origin");
+    t_assert(h3ToIjk(h3, h3, &origin) == 0, "got ijk for origin");
 
     for (int d = 1; d < 7; d++) {
         if (d == 1 && h3IsPentagon(h3)) {
@@ -66,8 +66,7 @@ void h3Distance_neighbors_assertions(H3Index h3) {
         H3Index offset = h3NeighborRotations(h3, d, &rotations);
 
         CoordIJK ijk = {0};
-        t_assert(H3_EXPORT(h3ToIjk)(h3, offset, &ijk) == 0,
-                 "got ijk for destination");
+        t_assert(h3ToIjk(h3, offset, &ijk) == 0, "got ijk for destination");
         CoordIJK invertedIjk = {0};
         _neighbor(&invertedIjk, d);
         for (int i = 0; i < 3; i++) {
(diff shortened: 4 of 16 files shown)


src/apps/applib/lib/utility.c
```diff
@@ -163,3 +165,35 @@ int readBoundary(FILE* f, GeoBoundary* b) {
...
+ * cell 0 up to the given base cell number.
+ */
+void iterateAllIndexesAtResPartial(int res, void (*callback)(H3Index),
+                                   int baseCells) {
+    assert(baseCells <= NUM_BASE_CELLS);
+    for (int i = 0; i < baseCells; i++) {
+        H3Index bc;
+        setH3Index(&bc, 0, i, 0);
+        int childrenSz = H3_EXPORT(maxUncompactSize)(&bc, 1, res);
+        H3Index* children = calloc(childrenSz, sizeof(H3Index));
```
isaacbrodsky: yes, this should be stack allocated

isaachier: :)

isaacbrodsky:
Looks like the build may have timed out due to iterating several resolutions of indexes.

dfellis:
What is the purpose of exposing this when its stated that the values produced should not be compared between different versions of H3? If this is for hexagon distance calculation, why not simply implement a function that does that but will return `-1` if any constraints are met, which can at first include the base cell constraint mentioned and potentially improved over time?

isaacbrodsky:
h3ToIjk has other uses, for example when using H3 bucketed data in systems which expect a matrix of data. I think some good examples would be using H3 bucketed data in a GPU or in a system like Tensorflow. Having non-hierarchical coordinates (which IJK+ fulfills) is needed for that case.

sahrk:
While I do understand the convenience of the IJK coordinates, I fear that most users won’t understand the issues involved and this might lead to confusion and frustration. Maybe better to keep those issues internal where they can be cleanly resolved; @dfellis solution gives a path to achieving that. Frankly, if I hadn’t been so over-cautious (cowardly?) there would be barely any `CoordIJK` in the H3 core. 

I wonder if many of the use cases that the IJK coordinates support might be served by a mapping of the H3 indexes of a single resolution, in their implicit integer order, to a compact 1-dimensional sequence. Function `h3ToSeqNum` would map an H3 index to its place in the sequence from 1 (or zero) to `numHexagons(res)`, and `seqNumToH3` would do the inverse. Then H3 bucketed data would be simple arrays, minimal sub-sequences of the total sequence. Note this sequence does a good job of preserving spatial locality, since cells which share a hierarchical ancestor would be closer together in the sequence (cells that share a parent would be contiguous). This would give you a non-hierarchical address format that can be safely externalized, since it needn’t change when the H3 internals evolve. I'd even volunteer to write the functions.

dfellis:
@sahrk aren't the compact 64-bit integer forms of H3 already a type of sequence number, though?

I do see @isaacbrodsky 's point about wanting a coordinate system and not just a 1-d sequence for matrix-type operations, but just exposing the CoordIJK values without being clear about their limitations seems like a not-great idea.

What about a function that takes an origin hexagon and a "k-ring" size and creates a simple 2D IJ coordinate system with `<0, 0>` centered on the specified hexagon (that would simply spit out the maximum I and J) and another function that takes the origin hexagon and I, J coordinates and spits out the specified hexagon id?

The first can be used to define boundaries (can return `<-1, -1>` if the k-ring size is too large for the algorithm) and the second can use `CoordIJK` internally to quickly get hexagons, but could in the future also fall back to the (I think I'm remembering correctly) `hexRing` algorithm to get all hexagons at a given radius (using the IJ coordinates to determine the radius and choosing the right hexagon based on the direction).

That one would take longer to figure out and probably wouldn't have to be in the first release, but this allows for an implementation that won't depend on internals that may or may not change in the future. I can implement that if you guys want.

sahrk:
The 64-bit integer H3 indexes form a sequence, just not a compact one. So they can't be used directly for  @isaacbrodsky 's data buffer case. For the physical layout of a data buffer it's not clear to me how a 3D buffer (with a redundant axis) is more tractable than a 1D buffer that retains the same structural information (e.g., you could define the neighbor function, metric distance, etc., using sequence numbers, even if it was just done by translating back-and-forth to H3 indexes).

Perhaps there are two use cases here that don't necessarily have the same solution: an H3-organized data buffer, and a planar hex grid coordinate system that allows users to perform arbitrary grid coordinate manipulations that aren't yet supported by H3 directly. For the latter @dfellis I like your idea of generating the planar coordinate system as needed for a particular problem. And if it's going to be used for the physical layout of data buffers then 2D makes more sense than 3D. I'm just afraid that the pentagons are going to make this approach very messy to scale to large geographical regions, whereas most of those complications go away for the 1D approach (the one complication I still see is that, since pentagons have a different sized descendent footprint than the hexagon base cells you can't use a constant stride length to step through the buffer by base cell).

isaacbrodsky:
I think keeping the function internal and only exposing a `h3Distance` as part of the public API could be a reasonable solution. In the future, `h3ToIjk` (or `h3ToIj`) could be exposed when we have more confidence in the coordinate space it creates for a given "origin". This API could naturally be complimented with an `H3Index ijToH3(H3Index origin, const CoordIJ *ij)`.

This approach is focused on smaller geographical regions, on the assumption that if you're working with data near the base cell level, performing the Haversine distance calculation is reasonable enough, and the distances could even be precomputed. It certainly gets very messy to deal with pentagons. :)

It was suggested to me that the API could provide the hexagons directly in a 2D planar coordinate format (perhaps named CoordIJ), but that would necessarily need to call a function like `h3ToIjk` because of H3's internal use of IJK+. We could wait on that being available to making the function part of the public API.

I'm not sure that @dfellis' solution will work. It seems like it would require the base cell grid to be an IJK coordinate system, and I'm not sure that works due to pentagon distortion. (I'd also hope for a solution which doesn't have time or memory complexity relative to the distance between the indexes, unless it can be precomputed i.e. if we could precompute something for all base cells).

I think the sequence number idea that @sahrk brought up is interesting, but as you mentioned above I think there are a few different use cases, so I think there may be cases where generating a problem-specific coordinate space would be advantageous.

src/h3lib/lib/h3Index.c
```diff
@@ -480,6 +481,31 @@ H3Index _h3RotatePent60ccw(H3Index h) {
     return h;
 }
 
+/**
+ * Rotate an H3Index 60 degrees clockwise about a pentagonal center.
+ * @param h The H3Index.
+ */
+H3Index _h3RotatePent60cw(H3Index h) {
```
nrabinowitz: Nit: Is `60cw` right here, or are we really rotating 72 degrees? Is there a more semantic way to express this, like `_h3RotatePentSideCW`?

src/h3lib/lib/h3Index.c
```diff
@@ -734,3 +760,232 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+ * origin index.
+ *
+ * @param origin An anchoring index for the ijk+ coordinate system.
+ * @param index Index to find the coordinates of
+ * @param out ijk+ coordinates of the index will be placed here on success
+ * @return 0 on success, or another value on failure.
+ */
+int h3ToIjk(H3Index origin, H3Index h3, CoordIJK* out) {
+    if (H3_GET_MODE(origin) != H3_HEXAGON_MODE ||
+        H3_GET_MODE(h3) != H3_HEXAGON_MODE) {
```
nrabinowitz: Nit: I'm not sure if this is worthwhile. We technically ought to be doing this check on every H3 index algo, and then there's a slippery slope about what we ought to be checking - e.g. would `h3IsValid` be better here? I have mixed feelings about this, since it's basically a service to the developer to guard against stuff we can't type-check, but because it has perf issues in some cases, we can only do it in some functions and not others. It would be nice if we had a consistent approach to this across all algo functions.

The ideal option would be to make this a compile-time check. I don't see any way to do that with externally-provided indexes though.

src/h3lib/lib/h3Index.c
```diff
@@ -734,3 +760,232 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+        // Only hexagon mode is relevant, since we can't
+        // encode directionality in CoordIJK.
+        return 1;
+    }
+
+    int res = H3_GET_RESOLUTION(origin);
+
+    if (res != H3_GET_RESOLUTION(h3)) {
```
nrabinowitz: Same here. Is there a reason to do this check here and not in other algos?

src/h3lib/lib/h3Index.c
```diff
@@ -734,3 +760,232 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+
+    int originBaseCell = H3_GET_BASE_CELL(origin);
+    int baseCell = H3_GET_BASE_CELL(h3);
+
+    // Direction from origin base cell to index base cell
+    int dir = 0;
+    int revDir = 0;
+    if (originBaseCell != baseCell) {
+        for (dir = 1; dir < 7; dir++) {
+            int testBaseCell = _getBaseCellNeighbor(originBaseCell, dir);
```
nrabinowitz: I'd suggest encapsulating this as `Direction getBaseCellNeighborDirection(int origin, int dest)`, in `baseCells.c`, returning `-1` on failure.

src/h3lib/lib/h3Index.c
```diff
@@ -734,3 +760,232 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+    };
+
+    if (dir != 0) {
+        assert(baseCell != originBaseCell);
+        assert(!(originOnPent && indexOnPent));
+
+        int pentagonRotations = 0;
+        int directionRotations = 0;
+
+        if (originOnPent) {
```
nrabinowitz: I don't have a concrete suggestion here, but is it possible to move the pentagon handling blocks to a separate function to hide the complexity they entail?

src/h3lib/lib/mathExtensions.c
```diff
@@ -37,3 +37,14 @@ int _ipow(int base, int exp) {
...
+/**
+ * _imax returns the maximum of two integers.
+ */
+int _imax(int a, int b) {
+    if (a > b) {
+        return a;
+    } else {
+        return b;
+    }
+}
```
isaachier: This is the classic case of where you should use a macro. Unless you can move this to a header (requires use of `static inline` so not strictly C89 compliant), a macro is the better choice here:

```
#define _IMAX(a, b) (((a) > (b)) ? (a) : (b))
```

src/apps/applib/include/utility.h
```diff
@@ -42,4 +42,8 @@ void geoBoundaryPrint(const GeoBoundary* b);
 void geoBoundaryPrintln(const GeoBoundary* b);
 int readBoundary(FILE* f, GeoBoundary* b);
 
+void iterateAllIndexesAtRes(int res, void (*callback)(H3Index));
+void iterateAllIndexesAtResPartial(int res, void (*callback)(H3Index),
```
isaachier: Again, C users would probably prefer macros here and you'll get better performance because C compiler can see what is going on in a macro, but function pointers are opaque. Side note: this is why using the C `qsort` in C++ is slower than `std::sort` from C++ STL, due to function pointers being worse off than templates.

Anyway, here's how I'd do it based on [Jansson](https://jansson.readthedocs.io/en/2.11/apiref.html#c.json_array_foreach):

```
#define H3_FOR_EACH_INDEX_RES(argName, res) H3_FOR_EACH_INDEX_PARTIAL((res), NUM_BASE_CELLS)
#define H3_FOR_EACH_INDEX_PARTIAL(argName, res, baseCells) \
    assert((baseCells) <= NUM_BASE_CELLS); \
     for (int i = 0; i < baseCells; i++) { \
         H3Index bc; \
         setH3Index(&bc, 0, i, 0); \
         int childrenSz = H3_EXPORT(maxUncompactSize)(&bc, 1, (res)); \
         STACK_ARRAY_CALLOC(H3Index, children, childrenSz); \
         H3_EXPORT(uncompact)(&bc, 1, children, childrenSz, (res)); \
         for (int j = 0; j < childrenSz; j++) { \
             if (children[j] == 0) { \
                 continue; \
             } \
             H3Index argName = children[j];
```

Then users can do this:

```
H3_FOR_EACH_INDEX_AT_RES(1, myIndex) {
    printf("%lld\n", myIndex);
}
```

Which expands to:

```
for (...) {
    ...
    H3Index myIndex = children[j];
    printf("%lld\n", myIndex);
}
```

src/apps/testapps/testH3ToIjk.c
```diff
@@ -0,0 +1,244 @@
...
+        _ijkNormalize(&ijk);
+
+        t_assert(_ijkMatches(&ijk, &origin), "back to origin");
+    }
+}
+
+void h3Distance_kRing_assertions(H3Index h3) {
+    int maxK = 5;
+    int r = H3_GET_RESOLUTION(h3);
+    if (r == 0) {
```
isaachier: Switch is more performant and cleaner than `else if` chain.

src/apps/testapps/testKRing.c
```diff
@@ -24,6 +24,52 @@
...
 #include "h3Index.h"
 #include "test.h"
+#include "utility.h"
+
+void kRing_equals_kRingInternal_assertions(H3Index h3) {
+    for (int k = 0; k < 3; k++) {
+        int kSz = H3_EXPORT(maxKringSize)(k);
+
+        H3Index *neighbors = calloc(kSz, sizeof(H3Index));
+        int *distances = calloc(kSz, sizeof(int));
```
isaachier: Looks like stack allocations make more sense, unless this is an example to avoid difficulties with build tools.

src/h3lib/lib/h3Index.c
isaachier: Using assertions makes sense if you have people using debug builds versus release builds. Other languages tend to default to release builds so that is still an issue.

src/h3lib/lib/h3Index.c
isaachier: Assertion is better as I stated above.

src/apps/testapps/testKRing.c
isaacbrodsky: This was refactored out of a previous test without updating the allocations.

src/h3lib/lib/h3Index.c
isaacbrodsky: This parallels the existing `_h3RotatePent60ccw`, which is a special case of `_h3Rotate60ccw`/`_h3Rotate60cw`. I think it could be argued we're rotating 60 degrees, but also skipping over part of the coordinate space.

src/h3lib/lib/h3Index.c
isaacbrodsky: I think some of the other algos don't do this because they do not have a way to return error cases to the caller. Other functions like geoToH3 do check for out of bounds inputs and return 0, and geoToH3 is certainly performance critical.

I would rather not use assertions, especially in release builds because instead of propagating an error it would abort the entire process, which would probably be bad.

src/apps/applib/include/utility.h
isaacbrodsky: I'll change this, but I am not a huge fan of macros. Since this is a test, I don't think performance is that big a concern.

isaachier: Ya everyone hates macros until they program in C for a while and realize it's the only way to avoid seriously ugly code. Also, underrated technique: [X macro](https://en.wikipedia.org/wiki/X_Macro).

Meanwhile, rethinking this API too so that the blocks have matching parentheses.

isaachier: OK seeing as this is a test I don't really think it's worth the effort. Thought this might lead to a convenience API in general. Never mind my suggestion.

isaacbrodsky:
Bumping for review from those who had comments before @dfellis @sahrk , if there are no comments I'd like to merge this and release 3.1.0.

dfellis:
I'm still not convinced exposing the `CoordIJK` struct is a good idea (rather than some coordinate system intentionally derived from an origin hexagon) but I won't block you on this.

isaacbrodsky:
@dfellis this pr no longer exposes CoordIJK as part of the public api. (It could be used by code willing to reach into internals)

I think you mean you prefer a 2d coordinate system derived from the origin, rather than the current ijk+ one used internally.

dfellis:
Wait, let me refresh this page. I leave Chrome tabs open for months, sometimes they break in weird ways, maybe I'm still seeing an old version of the code.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The public API of this library consists of the functions declared in file
 
 ## [Unreleased]
 ### Added
+- h3ToIjk function for getting IJK+ coordinates from an index (#83)
+- Internal `ijkDistance` function for determining the grid distance between IJK+ coordinates (#83)
+- Internal `h3Distance` function for determining the grid distance between H3 indexes (#83)
```
dfellis: Okay, you're right that it's not doing what I thought, but these CHANGELOG statements look wrong. `h3Distance` is marked as an export below, while `ijkDistance` and `h3ToIjk` are not.

Also the `h3ToIjk` filter application is not mentioned in these notes and I'm not entirely sure what it's purpose is, now?

isaacbrodsky: Oops, those are reversed. h3Distance should be the public one, the others are internal. I rebased and updated.

`h3ToIjk` is just for experimental use.

sahrk:
Until better solutions are available this sure is handy for lots of use cases. But I concur with @dfellis that a 2D `CoordIJ` result (with `ij.i = ijk.i - ijk.k; ij.j = ijk.j - ijk.k;`) would help dissociate these addresses from the internal H3 system and thus limit any potential confusion.

dfellis:
I agree with @sahrk and I don't even think it would be that hard? Especially if the proposed IJK -> IJ coord transform can be applied (though it looks like Dr. Sahr removed it?) I just referenced this PR from [another PR](https://github.com/uber/h3-java/issues/21) since it seems an efficient polyfill is needed, and a `CoordIJ` would make that much simpler to write and reason about, I think.

sahrk:
@dfellis I removed the straight-forward transformation into the `CoordIJ` system with i- and j-axes aligned with the original `CoordIJK` system because I realized that the best `CoordIJ` system would place the non-origin hex in the +i/+j quadrant, since there's no reason to try to preserve an arbitrary spherical concept of direction once we're on the plane. Since you have `ijk+` coordinates at least one of the components will be 0, and the other two axes (with components that are positive or zero) will map to the new i- and j-axes (@dfellis I think you suggested this in different words earlier).

sahrk:
Probably now moot on this particular PR, but @dfellis it occurred to me that for your polyfill use case (and many others) it probably makes sense to keep the orientation of cells constant around the same origin (within the limits of pentagon distortion), even if the resulting 2D components are negative. For that you could use `ij.i = ijk.i - ijk.k; ij.j = ijk.j - ij.k;`.

An alternative might be to use two H3 indexes to define a coordinate system, and then pass that coordinate system to `h3ToIJ`, as well as to `ijToH3`, which you'll probably also need to get your results back into H3.

isaacbrodsky:
@dfellis I'd like to point out regarding uber/h3-java#21 that this approach does support crossing icosahedron edges, it doesn't support crossing more than one base cell, and it doesn't support crossing pentagons.

@sahrk like you said, I think we need to preserve orientation.

I don't have any objection to adding CoordIJ and related public API functions, but I am wondering if adding them in a separate PR makes more sense. My main concern is that `h3ToIjk` has partial handling of pentagons, and it might be better to experiment with that more before adding it to the public API. If `h3ToIjk` is "good enough" then I can certainly add `int h3ToIj(H3Index origin, CoordIJ *ij);`

dfellis:
@isaacbrodsky if you preprocess the polygons into an array of polygons intersected by the icosahedron faces, that point is moot.

isaacbrodsky:
My plan is to merge this early next week. It might be best to open an issue to discuss a public "H3Index to planar coordinates" API, or else that discussion can happen on a separate PR proposing such an API.

src/h3lib/lib/h3Index.c
isaachier: Assertions are not compiled into release builds. The definition of `assert` looks something like this:

```
#ifdef NDEBUG  /* Release mode */
#define assert(cond) /* Empty definition */
#else  /* Debug mode */
#define assert(cond) ...  /* < Define assert macro */
#endif
```

PR #84, opened 2018-06-26
Feature: clear unused memory for pentagon children

zachasme:
This is a fantastic library!

I've encountered an inconsistency between `h3ToChildren` and `kRing`: The later sets memory to zero for pentagons. This left me confused when using `h3ToChildren` on pentagons which in my case returned valid indices (left over in memory from other calls) for the children that are skipped.

This pull request zeroes out memory for the unused slots.

Alternatively, the difference should at least be documented to let us know to use calloc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The public API of this library consists of the functions declared in file
 ## [Unreleased]
 ### Added
 - Added Direction enum, replacing int and defined constants (#77)
+- Ensured unused memory is cleared for pentagon children.
 
 ## [3.0.7] - 2018-06-08
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ The public API of this library consists of the functions declared in file
 ## [Unreleased]
 ### Added
 - Added Direction enum, replacing int and defined constants (#77)
-- Ensured unused memory is cleared for pentagon children.
+### Fixed
+- Ensured unused memory is cleared for pentagon children. (#84)
 
 ## [3.0.7] - 2018-06-08
 ### Added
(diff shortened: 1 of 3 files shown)


src/h3lib/lib/h3Index.c
```diff
@@ -194,6 +194,8 @@ void H3_EXPORT(h3ToChildren)(H3Index h, int childRes, H3Index* children) {
     for (int i = 0; i < 7; i++) {
         if (!isAPentagon || i != K_AXES_DIGIT) {
             H3_EXPORT(h3ToChildren)(makeDirectChild(h, i), childRes, children);
+        } else {
+            *children = H3_INVALID_INDEX;
```
isaacbrodsky: I'm not sure this will zero enough of the buffer, since the recursive call to h3ToChildren could have generated more than one index. I think it needs to zero out `bufferChildStep` indexes.

Here is the test case I used:
```
TEST(pentagonChildrenMultires) {
    H3Index pentagon;
    setH3Index(&pentagon, 1, 4, 0);

    const int expectedCount = (5 * 7) + (6);
    const int paddedCount = H3_EXPORT(maxH3ToChildrenSize)(pentagon, 3);

    H3Index* children = calloc(paddedCount, sizeof(H3Index));
    H3_EXPORT(h3ToChildren)(sfHex8, 10, children);
    H3_EXPORT(h3ToChildren)(pentagon, 3, children);

    // just for testing:
    for (int i = 0; i < paddedCount; i++) {
        printf("test: %llx\n", children[i]);
    }

    verifyCountAndUniqueness(children, paddedCount, expectedCount);
    free(children);
}
```

zachasme:
Thanks! I've updated the PR accordingly.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The public API of this library consists of the functions declared in file
 ## [Unreleased]
 ### Added
 - Added Direction enum, replacing int and defined constants (#77)
+- Ensured unused memory is cleared for pentagon children.
```
joegilley: tiny nit: i'd put this under `Fixed` instead of `Added`

PR #86, opened 2018-07-01
Add erlang binding

madninja:
Adds an erlang-h3 binding. 

The remainder bindings, docs and travis are in the works but this is a functioning binding. 

Thanks for h3, it's really cool!


isaacbrodsky:
Thanks for your PR! I'll republish the docs site today.

Issue #38, opened 2018-03-26
Apparent geoToH3 or H3ToGeo bug on master


travisbrady:
I used `geoToH3` to encode the coordinates for NYC and then ran the H3 hex string through `H3toGeo` and got a confusing result.
This is with a fresh clone of master built per readme instructions and running on a Mac.

```
$ ./bin/geoToH3 12
40.7128 -74.0060
8c2a107289061ff
```
Then
```
$ ./bin/h3ToGeo
8c2a107289061ff
40.7127470097 285.9940373051
```

Unless I misunderstand the purpose of these tools, 285.9940373051 is obviously an illegal value and far from the expected -74.0060.

```
$ uname -a
Darwin HA002727 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64
```

sahrk:
285.9940373051 is the same longitude as -74.0060. `geoToH3` returns longitude values in the range 0-360 degrees.

travisbrady:
Ahh, I see. Still somewhat confusing to me as I'd typically expect `ThingToInt(IntToThing(blah)) == blah`. 

Does H3 provide any docs or functions for these sorts of conversions? I think many people often use latitudes in the range -180 to 180, so this kind of thing would tend (I'd think) to confuse other users too. But I'll close for now.

dfellis:
I will note that the bindings (that haven't been released, yet, hi again @isaacbrodsky :) ) do convert back to -180 -> 180 (and -90 -> 90 for latitude) because that's what people tend to expect. I don't recall why we haven't pushed this down to the C code, but I do know it will cause some churn with the test suite.

isaacbrodsky:
I'm OK with changing the C library/applications/examples to return negative angles, since that's what the bindings do anyways. (hi @dfellis :) ) Or, we should add this to the [bindings documentation](https://github.com/uber/h3/blob/master/dev-docs/creating_bindings.md).

@travisbrady Please feel free to open a PR to add your bindings to the [bindings list](https://github.com/uber/h3/blob/master/docs/community/bindings.md) when you feel they're ready.

nrabinowitz:
I'd support updating the C lib here. I think it subverts user expectations, and agree with @travisbrady's point that it breaks assumptions about round-trip stability.

sahrk:
[...]

Issue #53, opened 2018-05-07
Move GeoJSON normalization logic into core C lib


nrabinowitz:
At present, the `h3SetToLinkedGeo` function produces a single polygon with multiple outer loops, without a guaranteed order. The correct GeoJSON output should only have one outer loop per polygon, and the order of loops within each polygon should be `outer, hole, hole, ...`.

The logic for this is included in the (as yet unpublished) JS and Python bindings, but should be moved into the core C lib. The API of `h3SetToLinkedGeo` would not change, so this might not require a new API function, but its correctness guarantees would improve.

alexkokorin2:
@isaacbrodsky any update on this?

alexkokorin2:
Could you give me a hint/source how to create (at least) temporary fix for this?

alexkokorin2:
@nrabinowitz @isaacbrodsky ping about temporary fix.

nrabinowitz:
@alexkokorin2 The basic algorithm looks like:

- Loop through all polygons
  - If the polygon winds counter-clockwise, it's an outer loop, put it into an array of outer loops
  - If the polygon winds clockwise, it's an inner loop (hole). Put it in a temp array of inner loops
- Loop through all inner loops
  - Take the first point. 
  - Loop through all outer loops. If the point is inside the outer loop, allocate the hole to that polygon.
- Return the array of outer loops.

This doesn't deal with nested donuts, but otherwise should work. I'm going to be releasing a JavaScript GeoJSON lib shortly, and will probably include a JS version of this algorithm in that lib until this issue is resolved in the core library.

PR #93, opened 2018-07-17
WIP: Constrain longitude between -180 and 180

zachasme:
The longitude range [0,360] tripped me up. Are you still interested in pushing the [-180,180] range down to C code, cf. [this comment](https://github.com/uber/h3/issues/38#issuecomment-376670998)?

If so I will go over the test suite as well.


nrabinowitz:
I don't object, but technically this might need to be a major version change, and as such might be worth putting off. Though in practice it wouldn't actually break the Java or JS bindings, it would just render their normalization functions unnecessary. Would like thoughts from @isaacbrodsky and @dfellis.

nrabinowitz:
(I should have added, thanks for the PR! I didn't actually realize this change could be implemented in such a minimal fashion.)

dfellis:
I'm really excited for this change, and also surprised that nothing needed to be done to the handling of latitudes, which I thought were similarly in the `[0, Pi)` range instead of `[-Pi/2, Pi/2)` range. (But then looking at the rest of the function it assumes that the north pole is `Pi/2` and the south pole at `-Pi/2`.)

Since it still *accepts* inputs greater than Pi for latitude and then converts them to the expected range, I don't think this is a backwards incompatible change (since the greater than Pi radians are equivalent to the less than 0 radians on the globe), especially for those using the bindings where it will run it through the `constrain` functions and convert to degrees from radians.

My only concerns are whether or not other parts of the code have hardwired expectations of only positive radians. The test suite shows that `geoToH3` is perfectly fine, which is great and was my biggest worry, but the failures in `testH3SetToLinkedGeo` and `testH3SetToVertexGraph` imply that there may be positive-only expectations hardwired in those features and the failure in `testH3Api` implies that the `constrainLng` call might be breaking `NAN` and/or `INFINITY` detection on the inputs.

nrabinowitz:
The test failures for `testH3SetToLinkedGeo` and `testH3SetToVertexGraph` are segfaults, which suggests they might be fixed by #94 - it would be worth rebasing to master and seeing if those issues go away.

isaacbrodsky:
@zachasme thanks for the PR, I think it definitely makes sense to move this into the core rather than the bindings.

You may want to rebase now that #94 is merged, since that improves support for negative coordinates. `testH3SetToVertexGraph` gets fixed, but it looks like `testH3SetToLinkedGeo` still has some failure have rebasing. I didn't look into that yet. `testH3Api` looks good to me, it's failing on a test for exact vertices, not the Infinity/NaN handling.

I don't think this requires a major version bump. Input should not be impacted and supported both negative and positive coordinates before. Output is different but is equivalent, and bindings were doing this transform before anyways.

nrabinowitz:
Hi @zachasme - I think we're all on board for this change. Are you planning to do the test updates, or should one of the maintainers take this on?

zachasme:
I'm planning to finish this when I get back from vacation next week.

nrabinowitz:
> The failure in testH3SetToLinkedGeo is due to the loops being reversed, possibly related to #53?

Hm. I will look into this today.
- Changes in the order of the list of loops are fine (we will normalize in #53, but even after that the order of the holes is undefined), and can be addressed in tests
- Changes in the starting coordinate of a given loop sequence are fine (all loops are closed, so the start coord is undefined), and can be addressed in tests
- Changes in the order or direction (clockwise vs counter-clockwise) of a given loop sequence is **not** fine, and will need to be addressed in code

nrabinowitz:
Re test failures in `testH3SetToLinkedGeo` - this is okay; the order of the loops within the list of loops is currently undefined/arbitrary, so reversing that order is acceptable.

Patch for this:

```
--- a/src/apps/testapps/testH3SetToLinkedGeo.c
+++ b/src/apps/testapps/testH3SetToLinkedGeo.c
@@ -119,10 +119,12 @@ TEST(hole) {
     H3_EXPORT(h3SetToLinkedGeo)(set, numHexes, &polygon);
 
     t_assert(countLinkedLoops(&polygon) == 2, "2 loops added to polygon");
-    t_assert(countLinkedCoords(polygon.first) == 6 * 3,
-             "All outer coords added to first loop");
-    t_assert(countLinkedCoords(polygon.first->next) == 6,
-             "All inner coords added to second loop");
+    // Note: This isn't strictly correct, and should be reversed when
+    // https://github.com/uber/h3/issues/53 is resolved
+    t_assert(countLinkedCoords(polygon.first) == 6,
+             "All inner coords added to first loop");
+    t_assert(countLinkedCoords(polygon.first->next) == 6 * 3,
+             "All outer coords added to second loop");
 
     H3_EXPORT(destroyLinkedPolygon)(&polygon);
```

zachasme:
I have applied the patch from @nrabinowitz, and fixed the remaining tests as well.

src/h3lib/lib/geoCoord.c
```diff
@@ -127,6 +127,9 @@ double constrainLng(double lng) {
     while (lng > M_PI) {
```
isaacbrodsky: Might be nice to replace this with something constant time in the future

PR #101, opened 2018-07-26
Add docs for h3Distance

isaacbrodsky:
Begins docs for #83, and adds `h3Distance` to the example.

### DIFF
--- a/docs/api/distance.md
+++ b/docs/api/distance.md
@@ -1,0 +1,13 @@
+# Distance functions
+
+## h3Distance
+
+```
+int h3Distance(H3Index origin, H3Index h3);
+```
+
+Returns the distance in grid cells between the two indexes.
+
+Returns a negative number if finding the distance failed. Finding the distance can fail because the two
+indexes are not comparable (different resolutions), too far apart, or are separated by pentagonal
+distortion.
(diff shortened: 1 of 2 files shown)


docs/api/distance.md
```diff
@@ -0,0 +1,13 @@
...
+
+## h3Distance
+
+```
+int h3Distance(H3Index origin, H3Index h3);
+```
+
+Returns the distance in grid cells between the two indexes.
+
+Returns a negative number if finding the distance failed. Finding the distance can fail because the two
```
nrabinowitz: Is it worth outlining the meaning of the different error codes here?

nrabinowitz:
Out of curiousity, does our doc setup mean this will be published right away, before we actually cut a release with this functionality? Can we add something to the docs that indicates what version of the library they document?

docs/api/distance.md
isaacbrodsky: `h3Distance` itself only returns -1 on any kind of error right now, it doesn't preserve the more specific error code from `h3ToIjk`

isaacbrodsky:
@nrabinowitz Our doc setup requires manually building the docs and pushing them to the gh-pages branch, so this will not be published until both merged and the docs are rebuilt.

It would be really nice to have a proper control on each of the docs page that says e.g. "Since: 3.1.0" or "Since: 3.0.0" or other annotations.

Issue #99, opened 2018-07-25
Add h3ToIjk to Java API and publish in a branch 


yujialuo:
Convert h3 hexagon id to users consumable IJK+ coordinates
Need to be able to import this functionality in a production level java codebase.

If you don't want to include this feature in public API, can you create an experimental branch and publish a separate .jar for us to import?

Existing PR:
[https://github.com/uber/h3/pull/83](url)

PR #102, opened 2018-07-26
Add h3ToIj

isaacbrodsky:
h3ToIj wraps the internal function h3ToIjk and produces a two-dimensional IJ coordinate system. It has the same limitations of h3ToIjk regarding cells too far away from the origin or across pentagonal distortion.

Do we want to make any changes to support evolving the CoordIJ system in the future?

Builds on #83 and relates to #99.


src/apps/testapps/testH3ToIj.c
```diff
@@ -111,6 +120,48 @@ void h3Distance_kRing_assertions(H3Index h3) {
 
 BEGIN_TESTS(h3ToIjk);
```
nrabinowitz: Nit: `h3ToIj` now? Not actually sure this is shown anywhere.

src/apps/testapps/testH3ToIj.c
```diff
@@ -111,6 +120,48 @@ void h3Distance_kRing_assertions(H3Index h3) {
...
 
+H3Index bc1 = H3_INIT;
+setH3Index(&bc1, 0, 15, 0);
```
nrabinowitz: Nit: Can you use clearer var names or add comments explaining what these are?

src/h3lib/lib/h3Index.c
```diff
@@ -960,6 +960,32 @@ int h3ToIjk(H3Index origin, H3Index h3, CoordIJK* out) {
...
+ * The coordinate space used by this function may have deleted
+ * regions or warping due to pentagonal distortion.
+ *
+ * Coordinates are only comparable if they come from the same
+ * origin index.
+ *
+ * @param origin An anchoring index for the ij coordinate system.
+ * @param index Index to find the coordinates of
+ * @param out ij coordinates of the index will be placed here on success
+ * @return 0 on success, or another value on failure.
```
nrabinowitz: Can you explain in the doc comment why you might get failure values?

nrabinowitz:
Looks good, just a few comments on documentation. I'm happy making this public if the behavior is well-defined. @dfellis or @sahrk might want to weigh in.

src/apps/testapps/testH3ToIj.c
```diff
@@ -38,8 +39,11 @@ void h3Distance_identity_assertions(H3Index h3) {
 
     t_assert(H3_EXPORT(h3Distance)(h3, h3) == 0, "distance to self is 0");
 
+    // Test that coordinates are as expected
+    CoordIJ ij;
+    t_assert(H3_EXPORT(h3ToIj)(h3, h3, &ij) == 0, "failed to get ij");
```
dfellis: The comment on this test doesn't make sense?

src/apps/testapps/testH3ToIj.c
```diff
@@ -57,8 +61,10 @@ void h3Distance_identity_assertions(H3Index h3) {
 }
 
 void h3Distance_neighbors_assertions(H3Index h3) {
-    CoordIJK origin = {0};
-    t_assert(h3ToIjk(h3, h3, &origin) == 0, "got ijk for origin");
+    CoordIJ origin = {0};
+    t_assert(H3_EXPORT(h3ToIj)(h3, h3, &origin) == 0, "got ijk for origin");
```
dfellis: Nit: Should say `got ij for origin`

src/apps/testapps/testH3ToIj.c
```diff
@@ -68,8 +74,11 @@ void h3Distance_neighbors_assertions(H3Index h3) {
         int rotations = 0;
         H3Index offset = h3NeighborRotations(h3, d, &rotations);
 
-        CoordIJK ijk = {0};
-        t_assert(h3ToIjk(h3, offset, &ijk) == 0, "got ijk for destination");
+        CoordIJ ij = {0};
+        t_assert(H3_EXPORT(h3ToIj)(h3, offset, &ij) == 0,
+                 "got ijk for destination");
```
dfellis: Same nit here on `ijk -> ij`

src/h3lib/include/h3api.h
```diff
@@ -129,6 +129,16 @@ struct LinkedGeoPolygon {
     LinkedGeoPolygon *next;
 };
 
+/** @struct CoordIJ
+ * @brief IJ hexagon coordinates
+ *
+ * Each axis is spaced 120 degrees apart.
+ */
+typedef struct {
+    int i;  ///< i component
```
dfellis: Nit: Do we really need these inline comments?

sahrk: @dfellis they parallel the inline comments on components in `CoordIJK`, so consistency would argue for keeping them here (or changing them there). Putting them in `CoordIJK` was a coin-flip in favor of completeness, but I agree they're overkill.

sahrk:
@isaacbrodsky asked awhile ago "Do we want to make any changes to support evolving the CoordIJ system in the future?"

You might want to specifically mention in the documentation that the result of `h3ToIj` for a specific pair of indexes is not guaranteed to be the same in future versions, unless you're really sure you want that to be the case.

Otherwise LGTM.

src/apps/testapps/testH3ToIj.c
isaacbrodsky: Done

src/apps/testapps/testH3ToIj.c
isaacbrodsky: Clarified that comment.

src/apps/testapps/testH3ToIj.c
isaacbrodsky: Done

isaacbrodsky:
Sorry I haven't had time to return to this PR.

Re: @sahrk 
> You might want to specifically mention in the documentation that the result of h3ToIj for a specific pair of indexes is not guaranteed to be the same in future versions, unless you're really sure you want that to be the case.

While I like the flexibility that gives me as a library developer, I'm concerned it might be confusing for library users. This also came up in https://github.com/uber/h3/pull/83#issuecomment-399594455, but most of the discussion there was about whether to expose `h3ToIjk` or `h3ToIj`. I'm interested in what others think of adding that caveat to the API.

dfellis:
> Sorry I haven't had time to return to this PR.

I've been on vacation until today so I didn't notice. ;)

> Re: @sahrk
>
>> You might want to specifically mention in the documentation that the result of h3ToIj for a specific pair of indexes is not guaranteed to be the same in future versions, unless you're really sure you want that to be the case.
>
> While I like the flexibility that gives me as a library developer, I'm concerned it might be confusing for library users. This also came up in #83 (comment), but most of the discussion there was about whether to expose h3ToIjk or h3ToIj. I'm interested in what others think of adding that caveat to the API.

I'm with @sahrk about noting that these values could change between versions of the code. As long as they're internally consistent and operations performed on them will work the same each time (eg `h3ToIj(h3Index1, h3Index2,&ijVar); ijVar.i += 2; H3Index h3Index3 = ijToH3(&ijVar, h3Index1)` will produce the same `h3Index3`) then the actual IJ values don't matter.

dfellis:
That does bring up another point: I don't see the paired `ijToH3` function in this PR, yet.

sahrk:
`ijToH3` gets to the crux of my discomfort, an `ij` isn’t a global location in and of itself; the user has to hang on to metadata (the origin) to get back to H3. It almost makes me want to create a struct to wrapper each `ij` with its origin, but that sounds inconvenient/inefficient. It might be better if users were encouraged up front to think of their `ij`’s as primarily local planar coordinates, with limited extent. Then when they start moving around in `ij` space and wander into a singularity they will be fore-warned.

dfellis:
Yes, but without something like that, this library would need to implement every kind of planar operation users desire (such as line tracing that's desired for more efficient polyfill operations). Having an "out" into a coordinate system will reduce the bloating of the API.

sahrk:
Agreed @dfellis, I think the utility case is very convincing. I was just encouraging more of a warning label.

nrabinowitz:
To throw in my two cents here:
- If we are fairly certain that the coordinate system is unstable, then I think we should explicitly mark it `experimentalH3ToIj` or something. As a developer, I would expect that the output of `h3ToIj(h3Index1, h3Index2, &ijVar)` would be stable across minor version changes, and if we're not providing that guarantee then we should be very clear about it.
- Supporting currently unreachable coordinate spaces would be a minor version bump as long as working pairs of indexes continued to work in the same way. That would be fine with a comment.

> As long as they're internally consistent and operations performed on them will work the same each time (eg h3ToIj(h3Index1, h3Index2,&ijVar); ijVar.i += 2; H3Index h3Index3 = ijToH3(&ijVar, h3Index1) will produce the same h3Index3) then the actual IJ values don't matter.

Is that true, or will people use the IJ values directly in their own code to implement their own algorithms in planar space? If so, I'd want a guarantee that output is stable or a very clear delineation of the unstable interfaces. I don't think a comment is sufficient unless we're willing to stick with current (successful) output until the next major version change.

isaacbrodsky:
@sahrk 
> It might be better if users were encouraged up front to think of their ij’s as primarily local planar coordinates, with limited extent.

I definitely want to encourage that. Would changing the name of the function help, perhaps `h3ToLocalIj`?

I like @nrabinowitz' suggestion of adding an "experimental" prefix, which could then become a deprecated alias when the function is promoted to not have the prefix. (`experimentalH3ToLocalIj`?)

sahrk:
@isaacbrodsky encoding a warning in the name sounds like a great idea to me. I think that adding `local` is sufficient to make the point, but would support either.

isaacbrodsky:
Now that we have a working build, I think this is ready for a new review.

CHANGELOG.md
```diff
@@ -8,19 +8,22 @@ The public API of this library consists of the functions declared in file
 ## [Unreleased]
 ### Added
 - `experimentalH3ToLocalIj` function for getting local coordinates for an index. (#102)
-### Fixed
-- Added #include <stdio.h> to benchmark.h
-### Added
+- `experimentalLocalIjToH3` function for getting an index from local coordinates. (#102)
 - Benchmarks for the kRing method for k's of size 10, 20, 30, and 40.
+### Changed
+- Internal `h3ToIjk` function renamed to `h3ToLocalIjk`. (#102)
```
dfellis: Is this supposed to be `h3ToLocalIj` like elsewhere or no?

src/h3lib/lib/localij.c
```diff
@@ -30,6 +30,83 @@
...
+    {0, -1, 0, 5, 0, 0, 0},        // 4
+    {0, -1, 5, 5, 0, 0, 0},        // 5
+    {0, -1, 0, 0, 0, 0, 0},        // 6
+};
+/**
+ * Reverse base cell direction -> leading index digit -> roations 60 ccw.
+ * For reversing the rotation introduced in PENTAGON_ROTATIONS when
+ * the origin is on a pentagon.
+ */
+const int PENTAGON_ROTATIONS_REVERSE[7][7] = {
```
dfellis: I get `PENTAGON_ROTATIONS`, `PENTAGON_ROTATIONS_REVERSE_NONPOLAR` and `PENTAGON_ROTATIONS_REVERSE_POLAR`, but I don't get this one. Is this one applied in both cases, but then why have the other two? Or why not simply push the compound rotation results into both of the ones below if this is applied and then one of the following two are applied?

src/h3lib/lib/localij.c
```diff
@@ -226,6 +271,193 @@ int h3ToLocalIjk(H3Index origin, H3Index h3, CoordIJK *out) {
...
+ *
+ * Failure may occur if the coordinates are too far away from the origin
+ * or if the index is on the other side of a pentagon.
+ *
+ * @param origin An anchoring index for the ijk+ coordinate system.
+ * @param ijk IJK+ Coordinates to find the index of
+ * @param out The index will be placed here on success
+ * @return 0 on success, or another value on failure.
+ */
+int localIjkToH3(H3Index origin, const CoordIJK* ijk, H3Index* out) {
```
dfellis: I thought IJK was going away from this diff? Or are we going H3 -> IJK -> IJ?

src/h3lib/lib/localij.c
```diff
@@ -263,6 +495,35 @@ int H3_EXPORT(experimentalH3ToLocalIj)(H3Index origin, H3Index h3,
...
+int H3_EXPORT(experimentalLocalIjToH3)(H3Index origin, const CoordIJ* ij,
+                                       H3Index* out) {
+    // This function is currently experimental. Once ready to be part of the
+    // non-experimental API, this function (with the experimental prefix) will
+    // be marked as deprecated and to be removed in the next major version. It
+    // will be replaced with a non-prefixed function name.
+    CoordIJK ijk;
+    ijToIjk(ij, &ijk);
+
+    return localIjkToH3(origin, &ijk, out);
```
dfellis: I see, you are doing exactly that. Alright, then.

CHANGELOG.md
isaacbrodsky: No, as below the IJK functions need to be kept for internal implementation.

src/h3lib/lib/localij.c
isaacbrodsky: `PENTAGON_ROTATIONS_REVERSE` is not used in the same cases. The comments try to differentiate that this table is used when the origin index is on a pentagon base cell, the two "reverse" tables are used when the origin index is not on a pentagon base cell.

src/apps/testapps/testH3ToLocalIj.c
```diff
@@ -0,0 +1,354 @@
...
+
+        H3Index testH3;
+
+        int failed = H3_EXPORT(experimentalLocalIjToH3)(h3, &ij, &testH3);
+        if (!failed) {
+            t_assert(H3_EXPORT(h3IsValid)(testH3),
+                     "test coordinates result in valid index");
+            if (_isBaseCellPentagon(H3_EXPORT(h3GetBaseCell)(testH3))) {
+                // h3IsValid doesn't test for indexes representing part of the
+                // deleted subsequence.
```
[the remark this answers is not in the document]
isaacbrodsky: I'd say such an index is invalid, similar to if the index had digits set to `7` (invalid child). I think opening an issue to make those indexes invalid would be good.

src/apps/testapps/testH3ToLocalIj.c
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    TEST(localIjToH3_kRing) {
+        iterateAllIndexesAtRes(0, localIjToH3_kRing_assertions);
+        iterateAllIndexesAtRes(1, localIjToH3_kRing_assertions);
+        iterateAllIndexesAtRes(2, localIjToH3_kRing_assertions);
+        // Don't iterate all of res 3, to save time
+        iterateAllIndexesAtResPartial(3, localIjToH3_kRing_assertions, 27);
+        // These would take too long, even at partial execution
+        // iterateAllIndexesAtResPartial(4, localIjToH3_kRing_assertions, 20);
```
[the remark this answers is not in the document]
isaacbrodsky: Replaced in all cases with the comment that further resolutions aren't tested for time reasons.

We could also consider adding some tests to sample indexes at higher resolutions in the future.

src/h3lib/lib/localij.c
```diff
@@ -0,0 +1,551 @@
...
+ * or if the index is on the other side of a pentagon.
+ *
+ * @param origin An anchoring index for the ijk+ coordinate system.
+ * @param index Index to find the coordinates of
+ * @param out ijk+ coordinates of the index will be placed here on success
+ * @return 0 on success, or another value on failure.
+ */
+int h3ToLocalIjk(H3Index origin, H3Index h3, CoordIJK* out) {
+    if (H3_GET_MODE(origin) != H3_HEXAGON_MODE ||
+        H3_GET_MODE(h3) != H3_HEXAGON_MODE) {
```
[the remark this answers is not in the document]
isaacbrodsky: If we have the ability to validate some user inputs, is there a reason we would not want to? I think some functions aren't able to cleanly report failures right now, which certainly limits where we can add checks (until 4.0.0 and #104)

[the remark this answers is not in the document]
dfellis: Yes, actually. If the function could be conceivably used in a looping hotspot I could see users want to have a faster iteration time if they're confident in the data already being valid.

This feels like one of those functions. H3 indexes could already be validated and stored in DB, then they want fast conversion to IJK for some planar operations and then back again.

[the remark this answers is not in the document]
isaacbrodsky: I should have replied directly to the performance concern rather than asking a question. I think in this specific case I am OK with removing the mode check because it doesn't appear this will cause issues, for example there are no array accesses based on the mode of the indexes.

Aside: This code was not introduced in this PR, it was introduced in #83 where we had a similar discussion but didn't come to a conclusion as @nrabinowitz suggested.

I believe there are at least some inputs we should validate, a good example being where there are array accesses based on the inputs, since that could potentially result in segfault, etc.

isaacbrodsky:
@dfellis and others, please review again.

src/apps/testapps/testH3ToLocalIj.c
```diff
@@ -66,22 +66,22 @@ void h3ToLocalIj_coordinates_assertions(H3Index h3) {
...
     } else if (r == 2) {
         CoordIJK expected = UNIT_VECS[H3_GET_INDEX_DIGIT(h3, 1)];
         _downAp7r(&expected);
         _neighbor(&expected, H3_GET_INDEX_DIGIT(h3, 2));
         t_assert(_ijkMatches(&ijk, &expected) == 1,
-                 "not at expected coordinates (res 2)");
+                 "res 2 cell at expected coordinates");
     } else {
-        t_assert(0, "wrong resolution");
+        t_assert(0, "resolution supported by test function (coordinates)");
```
nrabinowitz: Nit: This (essentially `assert.fail`) is the one place where negative assertion messages actually make more sense IMO.

src/h3lib/lib/localij.c
```diff
@@ -0,0 +1,544 @@
...
+ *
+ * @param origin Index to find the distance from.
+ * @param index Index to find the distance to.
+ * @return The distance, or a negative number if the library could not
+ * compute the distance.
+ */
+int H3_EXPORT(h3Distance)(H3Index origin, H3Index h3) {
+    CoordIJK originIjk, h3Ijk;
+    if (h3ToLocalIjk(origin, origin, &originIjk)) {
+        // Currently there are no tests that would cause getting the coordinates
```
isaachier: What does this mean practically? Is there no way to create a new test case that would trigger this check to fail?

isaacbrodsky: As far as I can see there is no input to `h3ToLocalIjk` where the origin is the same as the index that would fail. The only cases that return failure involve either resolution mismatch, being on different base cells, or deleted subsequence on the pentagon (which should be unreachable when origin=index.)

isaachier: I see what you mean.

PR #108, opened 2018-07-30
Add Libraries Using H3 page, link geojson2h3

nrabinowitz:
Adds [uber/geojson2H3](https://github.com/uber/geojson2H3) to the new Libraries page. This assumes that I don't need to do anything additional to add the new page to the nav bar.


nrabinowitz:
> You might need to add it to website/src/mdRoutes.js

Done, and built the website locally to confirm it's working.

PR #113, opened 2018-08-02
Fix compiling with a prefix set

isaacbrodsky:
We might want to add a test for this to CI, possibly by overriding one of the existing GCC or Clang tests.


isaacbrodsky:
#114 confirms this catches such errors in the future.

PR #116, opened 2018-08-05
Remove unused toolchain file

isaachier:


nrabinowitz:
Test failures suggest this was used after all?

isaachier:
Not really... The issue of using the `-Wall -Werror` with default C means that the expectation you are using C89, which you aren't. The for loop variables are causing errors.

cmake/toolchain.cmake
```diff
@@ -12,6 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
isaachier: Just pointing out I cannot remember why I originally put this here. I think it might have been to make linking in shared objects easier, but really cannot remember.

isaachier:
About the C standard in use: you aren't supposed to be using C11 anymore, and C99 shouldn't be required. This library doesn't really use C89 either, so I settled for C99 but did not require it.

Position independent code is not about platforms as much as libraries. When you make a static archive (`libh3.a`), your code is generally viewed as one big blob. Each jump instruction is encoded to an absolute position in the file, same as object files (`.o`). Dynamic libraries (`.so`/`.dylib`) have difficulties due to the way they are loaded into memory, and they cannot rely on absolute jumps. So they encode jump instructions as relative jumps from the current position.

Now back to the problem at hand. When you try to make a dynamic library from an archive, it won't succeed because the jump instructions are in the wrong format. Archives can be made to use relative jumps (`-fPIC` flag), but it is potentially slower for the CPU. In general, only dynamic libraries use relative jumps and everything else uses absolute jumps.

Here, I had made the code always use relative jumps. There seems to be no good reason to do this. Thus, I reverted it.

isaachier:
Got it. Thanks!

PR #118, opened 2018-08-10
Update examples and docs for longitude range change (#93)

isaacbrodsky:


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The public API of this library consists of the functions declared in file
 [h3api.h](./src/h3lib/include/h3api.h).
 
 ## [Unreleased]
+### Changed
+- Longitude outputs are now ine range [-Pi, Pi]. (#93)
```
nrabinowitz: Typo: `in range`. I'd add `guaranteed to be` as well.

examples/distance.c
```diff
@@ -65,8 +65,8 @@ int main(int argc, char *argv[]) {
         radsToDegs(geoHQ2.lon), h3Distance(h3HQ1, h3HQ2),
         haversineDistance(geoHQ1.lat, geoHQ1.lon, geoHQ2.lat, geoHQ2.lon));
     // Output:
-    // origin: (37.775236, 237.580245)
-    // destination: (37.789991, 237.597879)
+    // origin: (37.775236, -122.419755)
+    // destination: (37.789991, -122.402121)
     // grid distance: 2340
-    // distance in km: 2.256850km
+    // distance in km: 2.256853km
```
nrabinowitz: Curious why this changed - FPE? Why are all of the coords here so different in their fractional digits from the previous versions?

isaacbrodsky: The only difference I see here is 0.000003km difference on the distance. Not sure if that's important. The coords look like they are correctly updated to `unconstrained_longitude - 360`.

nrabinowitz: ...because apparently I can't do simple math, and wasn't expecting a different fraction with the sign change. Stet.

PR #122, opened 2018-08-17
Move polygon utilities to a separate file, add benchmarks

nrabinowitz:
This is prep for #53. Moves polygon utilities (point-in-poly, point-in-loop) to a separate `polygon.c` file, moves their corresponding tests to a new file, and adds benchmarks for `loopContainsPoint` and `bboxFromGeofence`. The main reasons for this are:

* I'll be adding several more utils here to deal with multipolygon normalization
* I'll need to refactor the current point-in-poly code to accept `LinkedGeoLoop` structs as well as `Geofence` structs

Current benchmarks are super-fast, which implies that a refactor should probably worry more about code complexity than small perf degradations:

```
[100%] Built target benchmarkPolygon
	-- loopContainsPointSmall: 0.000080 milliseconds per iteration (100000 iterations)
	-- loopContainsPointLarge: 0.000610 milliseconds per iteration (100000 iterations)
	-- bboxFromGeofenceSmall: 0.000050 milliseconds per iteration (100000 iterations)
	-- bboxFromGeofenceLarge: 0.000740 milliseconds per iteration (100000 iterations)
```


src/h3lib/lib/polygon.c
```diff
@@ -0,0 +1,138 @@
...
+    bool isTransmeridian = bboxIsTransmeridian(bbox);
+    bool contains = false;
+
+    double lat = coord->lat;
+    double lng = _normalizeLng(coord->lon, isTransmeridian);
+
+    for (int i = 0; i < geofence->numVerts; i++) {
+        GeoCoord a = geofence->verts[i];
+        GeoCoord b;
+        if (i + 1 == geofence->numVerts) {
```
isaachier: First, isn't `geofence->verts[(i + 1) % geofence->numVerts]` easier?

Fun optimization. If you always allocate powers of two (which is probably what happens when you call `malloc` anyway), the above simplifies to: `geofence->verts[(i + 1) & (geofence->numVerts - 1)]`. Supposedly, this is much more efficient. The compiler will generally do this optimization if it knows the capacity is always a power of two.

src/h3lib/lib/polygon.c
```diff
@@ -0,0 +1,138 @@
...
+        if (aLng == lng || bLng == lng) {
+            lng -= DBL_EPSILON;
+        }
+
+        // For the latitude of the point, compute the longitude of the
+        // point that lies on the line segment defined by a and b
+        // This is done by computing the percent above a the lat is,
+        // and traversing the same percent in the longitudinal direction
+        // of a to b
+        double ratio = (lat - a.lat) / (b.lat - a.lat);
```
isaachier: Might as well use `const` here and by other temporaries that won't change.

src/h3lib/lib/polygon.c
```diff
@@ -0,0 +1,138 @@
...
+        double testLng =
+            _normalizeLng(aLng + (bLng - aLng) * ratio, isTransmeridian);
+
+        // Intersection of the ray
+        if (testLng > lng) {
+            contains = !contains;
+        };
```
isaachier: Unnecessary/weird semicolon.

src/h3lib/lib/polygon.c
```diff
@@ -0,0 +1,138 @@
...
+    if (!bboxContains(bbox, coord)) {
+        return false;
+    }
+    bool isTransmeridian = bboxIsTransmeridian(bbox);
+    bool contains = false;
+
+    double lat = coord->lat;
+    double lng = _normalizeLng(coord->lon, isTransmeridian);
+
+    for (int i = 0; i < geofence->numVerts; i++) {
```
isaachier: Could you refactor the body of the loop into a helper like `static GeoCoord pairContainsPoint(const GeoCoord* a, const GeoCoord* b);`? That would allow iteration logic to lie in caller but ray casting algo to be generic.

src/h3lib/lib/polygon.c
```diff
@@ -0,0 +1,138 @@
...
+        for (int i = 0; i < geoPolygon->numHoles; i++) {
+            if (loopContainsPoint(&(geoPolygon->holes[i]), &bboxes[i + 1],
+                                  coord)) {
+                return false;
+            }
+        }
+    }
+
+    return contains;
+}
```
isaachier: No EOF.

src/h3lib/lib/polygon.c
```diff
@@ -0,0 +1,138 @@
...
+bool polygonContainsPoint(const GeoPolygon* geoPolygon, const BBox* bboxes,
+                          const GeoCoord* coord) {
+    // Start with contains state of primary geofence
+    bool contains =
+        loopContainsPoint(&(geoPolygon->geofence), &bboxes[0], coord);
+
+    // If the point is contained in the primary geofence, but there are holes in
+    // the geofence iterate through all holes and return false if the point is
+    // contained in any hole
+    if (contains && geoPolygon->numHoles > 0) {
```
isaachier: Wouldn't Golang teach you that you should always return early if you can 😉.

nrabinowitz:
@isaachier These are all good notes, but at present I'm just moving the existing code. Can I take them up in my next diff, which is refactoring point-in-poly anyway?

nrabinowitz:
Additionally moved `bboxFromVertices` to `polygon.c`, as it will require similar work to accept both `Geofence` and `LinkedGeoLoop` structs. Again, no changes to the code here, just moving it.

PR #123, opened 2018-08-20
Only enable coverage in debug builds

isaachier:


isaachier:
This breaks when build type changes after CMake configuration (ninja and other tools do this).

PR #126, opened 2018-08-21
Use clock_gettime instead of clock for benchmarks

isaachier:


src/apps/applib/include/benchmark.h
```diff
@@ -20,30 +20,51 @@
...
+#define BENCHMARK(NAME, ITERATIONS, BODY)                               \
+    do {                                                                \
+        const int iterations = ITERATIONS;                              \
+        const char* name = #NAME;                                       \
+        START_TIMER;                                                    \
+        for (int i = 0; i < iterations; i++) {                          \
+            BODY;                                                       \
+        }                                                               \
+        END_TIMER(duration);                                            \
+        printf("\t-- %s: %Le ns per iteration (%d iterations)\n", name, \
```
nrabinowitz: Can we change the formatting for nanoseconds here? In the example you posted on #125 I find it really hard to easily see order-of-magnitude differences. I'd prefer to still express this in ms, or maybe translate this to ops/second?

isaachier: I get that. I was considering working out a way to change the unit dynamically.

src/apps/applib/include/benchmark.h
```diff
@@ -20,30 +20,59 @@
...
+    struct timespec end;                           \
+    clock_gettime(CLOCK_MONOTONIC, &end);          \
+    struct timespec elapsed;                       \
+    elapsed.tv_nsec = end.tv_nsec - start.tv_nsec; \
+    elapsed.tv_sec = end.tv_sec - start.tv_sec;    \
+    if (elapsed.tv_nsec < 0) {                     \
+        elapsed.tv_sec--;                          \
+        elapsed.tv_nsec = 1E9 + elapsed.tv_nsec;   \
+    }                                              \
+    const long double var = elapsed.tv_sec * 1E9 + elapsed.tv_nsec
```
nrabinowitz: This still looks like nanoseconds. Am I missing something?

isaachier: Wow good catch. I must have forgotten about the final conversion back to microseconds.

src/apps/applib/include/benchmark.h
```diff
@@ -20,30 +20,59 @@
...
+    struct timespec end;                           \
+    clock_gettime(CLOCK_MONOTONIC, &end);          \
+    struct timespec elapsed;                       \
+    elapsed.tv_nsec = end.tv_nsec - start.tv_nsec; \
+    elapsed.tv_sec = end.tv_sec - start.tv_sec;    \
+    if (elapsed.tv_nsec < 0) {                     \
+        elapsed.tv_sec--;                          \
+        elapsed.tv_nsec = 1E9 + elapsed.tv_nsec;   \
+    }                                              \
+    const long double var = (elapsed.tv_sec * 1E9 + elapsed.tv_nsec) / 1E3
```
isaacbrodsky: Maybe add named constants for these scaling factors?

isaachier: I had thought that might be more straightforward, just got a bit verbose. Let me post the diff and you can tell me your opinion.

isaacbrodsky: Looks good to me, makes it clearer what the units are

PR #129, opened 2018-08-23
Add isClockwise algorithm

nrabinowitz:
More prep for #53. 

Adds `isClockwiseLinkedGeoLoop` and `isClockwiseGeofence` (we don't actually need `isClockwiseGeofence` but we get it for free and it was easier to include it than try to remove it). The algorithm is cribbed from [Turf.js](https://github.com/Turfjs/turf/blob/master/packages/turf-boolean-clockwise/index.ts) and adapted to use the existing H3 approach for handling shapes that cross the antimeridian.

### DIFF
--- a/src/apps/testapps/testPolygon.c
+++ b/src/apps/testapps/testPolygon.c
@@ -253,6 +253,22 @@ TEST(isClockwiseLinkedGeoLoop) {
     destroyLinkedGeoLoop(&loop);
 }
 
+TEST(isClockwiseLinkedGeoLoopFloats) {
+    const GeoCoord verts[] = {{0.1, 0.1}, {0.2, 0.2}, {0.1, 0.2}};
+
+    LinkedGeoLoop loop;
+    initLinkedLoop(&loop);
+
+    for (int i = 0; i < 3; i++) {
+        addLinkedCoord(&loop, &verts[i]);
+    }
+
+    t_assert(isClockwiseLinkedGeoLoop(&loop) == true,
+             "Got true for clockwise loop");
+
+    destroyLinkedGeoLoop(&loop);
+}
+
 TEST(isNotClockwiseLinkedGeoLoop) {
     const GeoCoord verts[] = {{0, 0}, {0, 1}, {1, 1}};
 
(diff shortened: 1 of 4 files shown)


src/h3lib/lib/polygon.c
```diff
@@ -57,13 +50,34 @@
  * @return              Whether the point is contained
  */
 
+/**
+ * Create a bounding box from a Geofence
+ * @name bboxFromGeofence
+ * @param geofence Input Geofence
+ * @param bbox     Output bbox
+ */
```
isaachier: Is this a case where Doxygen cannot easily determine the function name? Are there any problems with the resulting output?

src/apps/testapps/testPolygon.c
```diff
@@ -237,4 +237,85 @@ TEST(bboxFromLinkedGeoLoopNoVertices) {
...
 }
 
+TEST(isClockwiseGeofence) {
+    GeoCoord verts[] = {{0, 0}, {0.1, 0.1}, {0, 0.1}};
+
+    Geofence geofence;
+    geofence.verts = verts;
+    geofence.numVerts = 3;
+
+    t_assert(isClockwiseGeofence(&geofence) == true,
```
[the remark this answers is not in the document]
nrabinowitz: I was following the style in some of the other tests, but will update.

src/apps/testapps/testPolygon.c
```diff
@@ -237,4 +237,85 @@ TEST(bboxFromLinkedGeoLoopNoVertices) {
...
+
+    LinkedGeoLoop loop;
+    initLinkedLoop(&loop);
+
+    for (int i = 0; i < 3; i++) {
+        addLinkedCoord(&loop, &verts[i]);
+    }
+
+    t_assert(isClockwiseLinkedGeoLoop(&loop) == true,
+             "Got true for clockwise loop");
```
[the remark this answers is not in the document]
nrabinowitz: There are various points of view on assertion messages:
- Print what happened in the success case: `Got true for clockwise loop`
- Print what happened in the failure case: `Got false for clockwise loop`
- Print the requirement: `Clockwise loop should return true`
- Print a description of what's under test: `Clockwise loop`

These messages are valuable when tests fail, but I tend to think that the more common role of tests is as spec documentation, and these messages are going to be most frequently read by other developers looking into the function under test or adding additional tests. The "failure case" option is the worst for this use case, because it states what should not happen, not what should happen, making tests-as-spec practically useless.

I usually think the "requirement" option is the best for both use cases, especially if you have a nicer test framework that pretty prints successes and failures, so you see a list of passing and failing requirements. But the general style in this repo is the "success case" version, so I've used that here.

src/h3lib/lib/polygon.c
nrabinowitz: These are the functions created via macro in the include file, so there's nowhere for Doxygen to find the name (the symbol won't exist until compiled). I actually put these in to be human-readable; Doxygen doesn't do a great job of picking them up. Probably I should move them into `polygon.h`.

nrabinowitz: Moved to `polygon.h`, which fixes the Doxygen output.

nrabinowitz:
Thanks for the review @isaachier!

isaachier:
Anytime

src/apps/testapps/testPolygon.c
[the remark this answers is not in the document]
isaachier: OK I see. Sort of like BDD style ("it should do x").

src/h3lib/lib/polygon.c
isaachier: Nice.

PR #130, opened 2018-08-23
Fix bounding box bug for polygons crossing the antimeridian

nrabinowitz:
This fixes a previously unknown bug calculating the bounding box for most polygons crossing the antimeridian. Bounding boxes were using the max longitude as west and the min longitude as east, which only considers the vertices closest to the antimeridian, instead of using the min positive longitude as west and the max negative longitude as east (vertices furthest from the antimeridian).

My tests failed to catch this because my test polygons were all orthogonal rectangles 😞.

### DIFF
--- a/src/apps/testapps/testBBox.c
+++ b/src/apps/testapps/testBBox.c
@@ -88,9 +88,16 @@ TEST(transmeridian) {
                         {-0.4, M_PI - 0.1}};
     const Geofence geofence = {.numVerts = 4, .verts = verts};
     const BBox expected = {0.4, -0.4, -M_PI + 0.1, M_PI - 0.1};
-    const GeoCoord inside = {-0.1, M_PI};
+    const GeoCoord insideOnMeridian = {-0.1, M_PI};
     const GeoCoord outside = {1.0, M_PI - 0.5};
-    assertBBox(&geofence, &expected, &inside, &outside);
+    assertBBox(&geofence, &expected, &insideOnMeridian, &outside);
+
+    const GeoCoord westInside = {0.1, M_PI - 0.05};
+    t_assert(bboxContains(&expected, &westInside),
+             "Contains expected west inside point");
+    const GeoCoord eastInside = {0.1, -M_PI + 0.05};
+    t_assert(bboxContains(&expected, &eastInside),
+             "Contains expected east outside point");
 
     const GeoCoord westOutside = {0.1, M_PI - 0.5};
     t_assert(!bboxContains(&expected, &westOutside),
--- a/src/apps/testapps/testPolyfill.c
+++ b/src/apps/testapps/testPolyfill.c
@@ -26,38 +26,20 @@ GeoCoord sfVerts[] = {
     {0.659966917655, -2.1364398519396},  {0.6595011102219, -2.1359434279405},
     {0.6583348114025, -2.1354884206045}, {0.6581220034068, -2.1382437718946},
     {0.6594479998527, -2.1384597563896}, {0.6599990002976, -2.1376771158464}};
-Geofence sfGeofence;
+Geofence sfGeofence = {.numVerts = 6, .verts = sfVerts};
 GeoPolygon sfGeoPolygon;
+
 GeoCoord holeVerts[] = {{0.6595072188743, -2.1371053983433},
                         {0.6591482046471, -2.1373141048153},
                         {0.6592295020837, -2.1365222838402}};
-Geofence holeGeofence;
+Geofence holeGeofence = {.numVerts = 3, .verts = holeVerts};
 GeoPolygon holeGeoPolygon;
+
 GeoCoord emptyVerts[] = {{0.659966917655, -2.1364398519394},
                          {0.659966917655, -2.1364398519395},
                          {0.659966917655, -2.1364398519396}};
-Geofence emptyGeofence;
+Geofence emptyGeofence = {.numVerts = 3, .verts = emptyVerts};
 GeoPolygon emptyGeoPolygon;
-
-GeoCoord primeMeridianVerts[] = {
-    {0.01, 0.01}, {0.01, -0.01}, {-0.01, -0.01}, {-0.01, 0.01}};
-Geofence primeMeridianGeofence;
-GeoPolygon primeMeridianGeoPolygon;
-
-GeoCoord transMeridianVerts[] = {{0.01, -M_PI + 0.01},
-                                 {0.01, M_PI - 0.01},
-                                 {-0.01, M_PI - 0.01},
-                                 {-0.01, -M_PI + 0.01}};
-Geofence transMeridianGeofence;
-GeoPolygon transMeridianGeoPolygon;
-
-GeoCoord transMeridianHoleVerts[] = {{0.005, -M_PI + 0.005},
-                                     {0.005, M_PI - 0.005},
-                                     {-0.005, M_PI - 0.005},
-                                     {-0.005, -M_PI + 0.005}};
-Geofence transMeridianHoleGeofence;
-GeoPolygon transMeridianHoleGeoPolygon;
-GeoPolygon transMeridianFilledHoleGeoPolygon;
 
 static int countActualHexagons(H3Index* hexagons, int numHexagons) {
     int actualNumHexagons = 0;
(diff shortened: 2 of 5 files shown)


src/apps/testapps/testBBox.c
```diff
@@ -88,9 +88,16 @@ TEST(transmeridian) {
                         {-0.4, M_PI - 0.1}};
     const Geofence geofence = {.numVerts = 4, .verts = verts};
     const BBox expected = {0.4, -0.4, -M_PI + 0.1, M_PI - 0.1};
-    const GeoCoord inside = {-0.1, M_PI};
+    const GeoCoord insideOnMeridian = {-0.1, M_PI};
```
isaachier: IDK how you all feel about this idea, but I find it would be easier to read this code if it used [designated initializers](https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html):

```
const GeoCoord insideOnMeridian = { .lat = -0.1, .lon = M_PI }
```

src/apps/testapps/testPolyfill.c
```diff
@@ -59,6 +59,16 @@ Geofence transMeridianHoleGeofence;
 GeoPolygon transMeridianHoleGeoPolygon;
 GeoPolygon transMeridianFilledHoleGeoPolygon;
 
+static int countActualHexagons(H3Index* hexagons, int numHexagons) {
+    int actualNumHexagons = 0;
+    for (int i = 0; i < numHexagons; i++) {
+        if (hexagons[i] != 0) {
```
isaachier: I have been wondering where in the code do we rely on zero-initialization. There are reasons to avoid zero-initialization if possible. Is this a case when the caller relies on the fact that `hexagons` contains only valid hexagons or zeros?

src/apps/testapps/testPolyfill.c
```diff
@@ -262,6 +231,35 @@ TEST(polyfillTransmeridian) {
...
 
+TEST(polyfillTransmeridianComplex) {
+    // This polygon is "complex" in that it has > 4 vertices - this
+    // tests for a bug that was taking the max and min longitude as
+    // the bounds for transmeridian polygons
+    GeoCoord verts[] = {{0.1, -M_PI + 0.00001},  {0.1, M_PI - 0.00001},
+                        {0.05, M_PI - 0.2},      {-0.1, M_PI - 0.00001},
+                        {-0.1, -M_PI + 0.00001}, {-0.05, -M_PI + 0.2}};
+
+    Geofence geofence;
```
isaachier: Why not `Geofence geofence = { verts, 6 };`?

src/apps/testapps/testPolyfill.c
```diff
@@ -262,6 +231,35 @@ TEST(polyfillTransmeridian) {
...
+    geofence.verts = verts;
+    geofence.numVerts = 6;
+
+    GeoPolygon polygon;
+    polygon.geofence = geofence;
+    polygon.numHoles = 0;
```
isaachier: Same question.

src/apps/testapps/testPolyfill.c
```diff
@@ -262,6 +231,35 @@ TEST(polyfillTransmeridian) {
...
+
+    int numHexagons = H3_EXPORT(maxPolyfillSize)(&polygon, 4);
+
+    H3Index* hexagons = calloc(numHexagons, sizeof(H3Index));
+    H3_EXPORT(polyfill)(&polygon, 4, hexagons);
+
+    int actualNumHexagons = countActualHexagons(hexagons, numHexagons);
+
+    t_assert(actualNumHexagons == 1204,
```
isaachier: If we keep `t_assert` as immediately exiting program, this is fine. Changing that to another behavior might result in a memory leak if we don't hit line 260.

src/apps/testapps/testPolygon.c
```diff
@@ -141,6 +141,22 @@ TEST(bboxFromGeofence) {
     t_assert(bboxEquals(&result, &expected), "Got expected bbox");
 }
 
+TEST(bboxFromGeofenceTransmeridian) {
+    GeoCoord verts[] = {{0.1, -M_PI + 0.1},  {0.1, M_PI - 0.1},
+                        {0.05, M_PI - 0.2},  {-0.1, M_PI - 0.1},
+                        {-0.1, -M_PI + 0.1}, {-0.05, -M_PI + 0.2}};
+
+    Geofence geofence;
```
isaachier: Same question.

src/h3lib/include/polygonAlgos.h
```diff
@@ -138,15 +138,17 @@ void GENERIC_LOOP_ALGO(bboxFrom)(const TYPE* loop, BBox* bbox) {
...
     }
-    double lat;
-    double lon;
 
     bbox->south = DBL_MAX;
     bbox->west = DBL_MAX;
     bbox->north = -1.0 * DBL_MAX;
     bbox->east = -1.0 * DBL_MAX;
+    double minPosLon = DBL_MAX;
+    double maxNegLon = -1.0 * DBL_MAX;
```
isaachier: For any case of `-1.0 * DBL_MAX` I would highly suggest changing to `-DBL_MIN`. I'm sure this is what happens in most compilers anyway, but unary minus avoids theoretical multiplication.

src/h3lib/include/polygonAlgos.h
```diff
@@ -161,15 +163,18 @@ void GENERIC_LOOP_ALGO(bboxFrom)(const TYPE* loop, BBox* bbox) {
         if (lon < bbox->west) bbox->west = lon;
         if (lat > bbox->north) bbox->north = lat;
         if (lon > bbox->east) bbox->east = lon;
+        // Save the min positive and max negative longitude for
+        // use in the transmeridian case
+        if (lon > 0 && lon < minPosLon) minPosLon = lon;
+        if (lon < 0 && lon > maxNegLon) maxNegLon = lon;
```
isaachier: Nice.

src/apps/testapps/testBBox.c
nrabinowitz: Yeah, we've been really inconsistent on this in tests, but I think that's probably the best/most legible option for most structs. `GeoCoord` might be the exception, because a) we instantiate a lot of them, and b) it's really equivalent to a `lat, lon` tuple, so instantiating it with the same syntax as a `double[]` array makes sense.

isaachier: I get that. Just a warning about designated initializers: it's not 100% portable, but should be okay on most platforms.

src/apps/testapps/testPolyfill.c
nrabinowitz: Yes. In algorithms where we return a list of hexagons and cannot say ahead of time how many we need, we require the caller to pass in a zero-filled array of the max possible size and then only fill H3 addresses for the real hexagons. We could theoretically fill in index order and then add a NULL as a stop signal afterward, but internally this isn't as efficient because we often need to use the array as a set, which we do via modulo arithmetic on the index, resulting in arbitrary index positions.

src/apps/testapps/testPolyfill.c
nrabinowitz: Will update using designated initializers.

src/apps/testapps/testPolyfill.c
nrabinowitz: ditto.

src/apps/testapps/testPolyfill.c
nrabinowitz: This would be true in a lot of places. I'm assuming the current behavior is fine in this test, otherwise we'd have the same problem all over. If `t_assert` doesn't exit the program, I'd assume it doesn't exit the function either, so we'd still hit L260.

src/h3lib/include/polygonAlgos.h
nrabinowitz: Do you mean `-DBL_MAX`? Will update.

isaachier: Ya my mistake. Thanks.

src/apps/testapps/testPolyfill.c
isaachier: Ya I have a feeling that's true. OK for now, assertions will remain fatal.

PR #131, opened 2018-08-23
Normalize MultiPolygon output in h3SetToLinkedGeo

### LINKED ISSUE #53: Move GeoJSON normalization logic into core C lib

nrabinowitz:
Fixes #53 

Adds an internal function `normalizeMultiPolygon` to align output of `h3SetToMultiPolygon` with the GeoJSON spec, ensuring that each polygon has only one outer loop, followed by holes. Note that this is more correct than the implementation currently in `geojson2h3`, as it should properly handle all topologies, including nested donuts.

I will update with a benchmark, but in most simple cases (no holes) this should be close to master performance due to early exits/skipped loops.


src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+ * Find the polygon to which a given hole should be allocated. Note that this
+ * function will return null if no parent is found.
+ * @param  loop         Inner loop describing a hole
+ * @param  polygon      Head of a linked list of polygons to check
+ * @param  polygonCount Number of polygons to check
+ * @return              Pointer to parent polygon, or null if not found
+ */
+static LinkedGeoPolygon* findPolygonForHole(LinkedGeoLoop* loop,
+                                            LinkedGeoPolygon* polygon,
+                                            BBox* bboxes, int polygonCount) {
```
isaachier: `bboxes` are not part of above `@param` documentation, making the logic here harder to follow.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+    // Early exit with no polygons
+    if (polygonCount == 0) {
+        return NULL;
+    }
+    // Initialize arrays for candidate loops and their bounding boxes
+    LinkedGeoPolygon** candidates =
```
isaachier: `calloc` might be overkill if you know that you will overwrite the whole array with new values.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+            candidates[containerCount] = polygon;
+            candidateBBoxes[containerCount] = &bboxes[index];
+            containerCount++;
+        }
+        polygon = polygon->next;
+        index++;
+    }
+
+    // Set the initial return value to the first candidate
+    LinkedGeoPolygon* parent = candidates[0];
```
isaachier: Probably better as a `const LinkedGeoPolygon*` seeing as you do not intend to change it.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+            }
+        }
+    }
+
+    // Free allocated memory
+    free(candidates);
+    free(candidateBBoxes);
+
+    return parent;
+}
```
isaachier: This function could use a few helpers to clarify what is going on. I'm losing the flow of it due to the length.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+
+    // Set up iteration variables
+    LinkedGeoPolygon* polygon = NULL;
+    LinkedGeoLoop* next;
+    int innerCount = 0;
+    int outerCount = 0;
+
+    // Iterate over all loops, moving inner loops into an array and
+    // assigning outer loops to new polygons
+    while (loop) {
```
isaachier: Is there any way to refactor these loops? I find them to be the hardest to grasp conceptually. A good practice I picked up from C++ was to avoid loops when an [std::algorithm](https://en.cppreference.com/w/cpp/algorithm) was available to replace it. That made code flow more idiomatic. This piece of code may be doing something like `std::partition`. Perhaps other parts are more like `std::transform`. If you find a matching pattern, consider refactoring with a clear name, such as `selectInnerLoops`.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+        if (polygon) {
+            addLinkedLoop(polygon, innerLoops[i]);
+        } else {
+            // If we can't find a polygon (possible with invalid input), then
+            // we need to release the memory for the hole, because the loop has
+            // been unlinked from the root and the caller will no longer have
+            // a way to destroy it with destroyLinkedPolygon.
+            destroyLinkedGeoLoop(innerLoops[i]);
+            free(innerLoops[i]);
+            resultCode = NORMALIZATION_ERR_UNASSIGNED_HOLES;
```
isaachier: Should you `break` or are the other loops still potentially valid?

src/h3lib/lib/algos.c
```diff
@@ -814,5 +814,6 @@ void H3_EXPORT(h3SetToLinkedGeo)(const H3Index* h3Set, const int numHexes,
     VertexGraph graph;
     h3SetToVertexGraph(h3Set, numHexes, &graph);
     _vertexGraphToLinkedGeo(&graph, out);
+    normalizeMultiPolygon(out);
```
isaacbrodsky: No check of the return value?

src/apps/testapps/testH3SetToLinkedGeo.c
```diff
@@ -69,35 +55,35 @@ TEST(contiguous2) {
...
     H3_EXPORT(destroyLinkedPolygon)(&polygon);
-    free(set);
 }
 
 TEST(contiguous3) {
     LinkedGeoPolygon polygon;
-    char* hexes[] = {"8928308288bffff", "892830828d7ffff", "8928308289bffff"};
-    int numHexes = sizeof(hexes) / sizeof(hexes[0]);
-    H3Index* set = makeSet(hexes, numHexes);
+    H3Index set[] = {0x8928308288bffffl, 0x892830828d7ffffl,
```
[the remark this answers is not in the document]
nrabinowitz: Will check, was following the pattern in other tests.

src/apps/testapps/testH3SetToLinkedGeo.c
```diff
@@ -189,14 +166,83 @@ TEST(2RingUnordered) {
...
+    H3_EXPORT(destroyLinkedPolygon)(&polygon);
 }
 
 TEST(contiguous2distorted) {
     LinkedGeoPolygon polygon;
-    char* hexes[] = {"894cc5365afffff", "894cc536537ffff"};
-    int numHexes = sizeof(hexes) / sizeof(hexes[0]);
-    H3Index* set = makeSet(hexes, numHexes);
+    H3Index set[] = {0x894cc5365afffffl, 0x894cc536537ffffl};
+    int numHexes = sizeof(set) / sizeof(set[0]);
```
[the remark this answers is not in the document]
nrabinowitz: Reasonable, will update.

src/apps/testapps/testPolygon.c
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <assert.h>
```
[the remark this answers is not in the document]
nrabinowitz: I use it for assertions on memory allocation.

src/apps/testapps/testPolygon.c
```diff
@@ -239,72 +214,343 @@ TEST(isClockwiseGeofence) {
...
     LinkedGeoLoop loop;
-    initLinkedLoop(&loop);
-
-    for (int i = 0; i < 4; i++) {
-        addLinkedCoord(&loop, &verts[i]);
-    }
+    createLinkedLoop(&loop, &verts[0], 4);
 
-    t_assert(!isClockwiseLinkedGeoLoop(&loop),
+    t_assert(isClockwiseLinkedGeoLoop(&loop) == false,
```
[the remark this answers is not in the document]
nrabinowitz: Bad rebase, will fix.

src/apps/testapps/testPolygon.c
```diff
@@ -239,72 +214,343 @@ TEST(isClockwiseGeofence) {
...
              "Got false for counter-clockwise transmeridian loop");
 
     destroyLinkedGeoLoop(&loop);
 }
 
+TEST(normalizeMultiPolygonSingle) {
+    GeoCoord verts[] = {{0, 0}, {0, 1}, {1, 1}};
+
+    LinkedGeoLoop* outer = calloc(1, sizeof(*outer));
+    assert(outer != NULL);
```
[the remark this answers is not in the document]
nrabinowitz: Again, following the pattern in other tests. Probably best to use `t_assert` for "spec" style assertions - this is a little lower level.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+
+/**
+ * Count the number of polygons containing a given loop.
+ * TODO: It would be really nice to memoize this.
+ * @param  loop         Loop to count containers for
+ * @param  polygons     Polygons to test
+ * @param  polygonCount Number of polygons in the test array
+ * @return              Number of polygons containing the loop
+ */
+static int countContainers(LinkedGeoLoop* loop, LinkedGeoPolygon** polygons,
```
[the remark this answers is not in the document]
nrabinowitz: I use the doc comments mostly for human readers - hopefully the `static` piece is sufficient to tell Doxygen that this is private. Not sure it's worth an extra notation - we don't use it in any other internal function.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
 #undef IS_EMPTY
 #undef INIT_ITERATION
 #undef ITERATE
+
+/**
+ * Count the number of polygons containing a given loop.
+ * TODO: It would be really nice to memoize this.
```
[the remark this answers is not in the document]
nrabinowitz: No, probably not - will remove the comment. There are extreme cases - many-level-deep nesting - where memoization would likely make a big difference, but I doubt they have any real-world use.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,172 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+
+/**
+ * Find the polygon to which a given hole should be allocated. Note that this
+ * function will return null if no parent is found.
+ * @param  loop         Inner loop describing a hole
+ * @param  polygon      Head of a linked list of polygons to check
+ * @param  polygonCount Number of polygons to check
+ * @return              Pointer to parent polygon, or null if not found
+ */
+static LinkedGeoPolygon* findPolygonForHole(LinkedGeoLoop* loop,
```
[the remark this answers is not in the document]
nrabinowitz: Will update.

src/h3lib/lib/polygon.c
nrabinowitz: Will add to the docs. A `bbox` is necessary for each polygon in order to run the `pointInside` algorithm, which relies on the `bbox` for fast failure and identification of transmeridian polygons.

src/h3lib/lib/polygon.c
nrabinowitz: If I don't use `calloc` I can't refer to `candidates[0]` without guards... I'll see if this can be changed to `malloc`.

src/h3lib/lib/polygon.c
nrabinowitz: I *do* intend to change it - this is the current candidate, but gets overridden on L168 if there's another one with more parents.

src/h3lib/lib/polygon.c
nrabinowitz: I could probably break out the second loop into a function. I'll take a look.

src/h3lib/lib/polygon.c
nrabinowitz: It's not really either - it's like `partition` but it handles the two groups differently. This loop populates the `innerLoops` array, but it attaches the outer loops to a linked structure instead, and additionally calculates the `bbox` for each outer loop (we don't need it for the inner). I could move this into function, but I'd need to pass in the three different data structures to populate, which make that seem weird.

I'll think about this, but I already broke out several pieces, and I think passing the necessary arguments around is going to make this harder to follow with more functions, not easier.

src/h3lib/lib/polygon.c
nrabinowitz: Other loops may still be valid.

src/apps/benchmarks/benchmarkH3SetToLinkedGeo.c
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2017 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
nrabinowitz: Will update.

src/h3lib/lib/algos.c
nrabinowitz: There's no return value for `h3SetToLinkedGeo`, so I really wasn't sure what to do with the return value if it's > 0. There's no known case where this would be possible in any case - the rest of the algo should guarantee valid input.

src/h3lib/lib/polygon.c
nrabinowitz: Switch memory assignments to `malloc`.

src/h3lib/lib/algos.c
isaachier: That is a problem... worst case scenario, you can set `errno`. If you find a scenario where this fails, either add a return value (which might be a breaking API change depending how you look at it), add a variant of this function that has a return value (`h3SetToLinkedGeoWithReturnCode`), or document that using such a polygon would result in undefined behavior.

src/h3lib/lib/polygon.c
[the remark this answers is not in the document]
isaachier: Interesting. If you do encounter that, maybe memoize by adding a field to the struct or something.

src/h3lib/lib/polygon.c
isaachier: Nice. Just trying to reduce the `calloc` throughout. I have heard using `calloc` is actually a bad idea when using a tool like Valgrind or AddressSanitizer. The zeroing suppresses errors that are likely due to uninitialized data. Like this:

```
int* array = calloc(10, sizeof(int));
// TODO: Load values of array from file.
for (int i = 0; i < 10; i++) {
    array[i] *= 5;
}
```

In the above case, the initialization is marked as `TODO`, but Valgrind won't warn about it because it sees you set the values to zero. On the other hand, using `malloc` would have caused Valgrind to warn about an invalid access to uninitialized data and reminded the programmer to implement the `TODO` behavior.

src/h3lib/lib/polygon.c
isaachier: Got it.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,192 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+ * Find the polygon to which a given hole should be allocated. Note that this
+ * function will return null if no parent is found.
+ * @param  loop         Inner loop describing a hole
+ * @param  polygon      Head of a linked list of polygons to check
+ * @param  polygonCount Number of polygons to check
+ * @return              Pointer to parent polygon, or null if not found
+ */
+static LinkedGeoPolygon* findPolygonForHole(const LinkedGeoLoop* loop,
+                                            LinkedGeoPolygon* polygon,
+                                            const BBox* bboxes,
```
isaachier: Still missing docs about `bboxes`?

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,192 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+            candidateBBoxes[candidateCount] = &bboxes[index];
+            candidateCount++;
+        }
+        polygon = polygon->next;
+        index++;
+    }
+
+    // The most deeply nested container is the immediate parent
+    LinkedGeoPolygon* parent =
+        findDeepestContainer(candidates, candidateBBoxes, candidateCount);
```
isaachier: Love this!

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,192 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+}
+
+/**
+ * Given a list of nested containers, find the one most deeply nested.
+ * @param  polygons     Polygon containers to check
+ * @param  bboxes       Bounding boxes for polygons, used in point-in-poly check
+ * @param  polygonCount Number of polygons in the list
+ * @return              Deepest container, or null if list is empty
+ */
+static LinkedGeoPolygon* findDeepestContainer(LinkedGeoPolygon** polygons,
```
isaachier: OK I understand the motivation to use non-`const` for both return type and polygons, but it makes it seem like you modify them in this function, which as far as I can tell, isn't the case. Maybe return `const LinkedGeoPolygon*` and just cast away the `const` in the calling code. This is pretty standard C style when you have an unnecessary `const`. C++ is a little better about it because you can have `const` and non-`const` overloads of the same function, but since C does not, it is accepted to cast away the `const`.

My obsession with `const` is from what I have heard that using `const` wherever possible is better practice and allows for better compiler optimizations.

src/h3lib/lib/polygon.c
isaachier: Thanks.

src/h3lib/lib/algos.c
nrabinowitz: The issue here is that the user isn't passing in a polygon, they're passing in an array of hexagons. An array of hexagons with no dupes should *never* hit the error cases here, and if it does that's indicative of a problem in H3, not user error. I don't think it's reasonable to make all callers check for errors so that they can file bugs against our codebase if they get one.

I think it's possible for a user to hit the error cases if they pass in dupe hexagons, though I'm not sure. But this is possibly the worst way to check for that particular case.

src/h3lib/lib/polygon.c
nrabinowitz: What's the syntax for casting it away? The issue is that I pass in polygons, get one of them back, and then modify it. If I say it's `const` in the param list, then I have to return it as `const`, and modification is a compile error. Happy to have it treated as const in the inner functions if there's a way to do so.

nrabinowitz: Nvm, got it and added the `const` qualifiers.

src/h3lib/lib/algos.c
isaachier: Ya. That's what I mean by "worst case scenario." I think your assumption that failure isn't possible makes sense for now. In the unlikely event that there is a situation that we haven't seen yet, then you still have a few options for emitting an error without breaking the API.

src/h3lib/lib/polygon.c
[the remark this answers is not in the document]
isaachier: `TODO` is still here.

src/h3lib/lib/polygon.c
```diff
@@ -264,7 +265,8 @@ int normalizeMultiPolygon(LinkedGeoPolygon* root) {
 
     // Find polygon for each inner loop and assign the hole to it
     for (int i = 0; i < innerCount; i++) {
-        polygon = findPolygonForHole(innerLoops[i], root, bboxes, outerCount);
+        polygon = (LinkedGeoPolygon*)findPolygonForHole(innerLoops[i], root,
```
isaachier: Great!

src/apps/benchmarks/benchmarkH3SetToLinkedGeo.c
```diff
@@ -0,0 +1,64 @@
...
+});
+
+BENCHMARK(h3SetToLinkedGeoDonut, 10000, {
+    initLinkedPolygon(&polygon);
+    H3_EXPORT(h3SetToLinkedGeo)(donut, donutCount, &polygon);
+    H3_EXPORT(destroyLinkedPolygon)(&polygon);
+});
+
+BENCHMARK(h3SetToLinkedGeoNestedDonuts, 10000, {
+    initLinkedPolygon(&polygon);
```
isaacbrodsky: This init step looks unnecessary - if it's needed, it should have been part of the public API.

src/h3lib/lib/algos.c
isaacbrodsky: I suggest in 4.0.0 we make essentially all functions able to indicate error. For the time being maybe add a comment that the error is being discarded?

src/h3lib/lib/linkedGeo.c
```diff
@@ -100,6 +109,19 @@ LinkedGeoCoord* addLinkedCoord(LinkedGeoLoop* loop, const GeoCoord* vertex) {
     return coord;
 }
 
+/**
+ * Free all allocated memory for a linked geo loop. The caller is
+ * responsible for freeing memory allocated to input loop struct.
+ * @param loop Loop to free
+ */
+void destroyLinkedGeoLoop(LinkedGeoLoop* loop) {
+    for (LinkedGeoCoord *currentCoord = loop->first, *nextCoord;
```
isaacbrodsky: nit: Move `LinkedGeoCoord *nextCoord;` onto it's own line? I'm not sure if we have a policy of permitting/forbidding multiple declarations per line, but I would suggest against it.

src/h3lib/lib/polygon.c
```diff
@@ -93,3 +95,193 @@ bool pointInsidePolygon(const GeoPolygon* geoPolygon, const BBox* bboxes,
...
+    int resultCode = NORMALIZATION_SUCCESS;
+    LinkedGeoPolygon* polygon = NULL;
+    LinkedGeoLoop* next;
+    int innerCount = 0;
+    int outerCount = 0;
+
+    // Create an array to hold all of the inner loops. Note that
+    // this array will never be full, as there will always be fewer
+    // inner loops than outer loops.
+    LinkedGeoLoop** innerLoops = malloc(loopCount * sizeof(LinkedGeoLoop*));
```
isaacbrodsky: Why not calloc for these count * size allocations?

nrabinowitz: Because @isaachier told me not to :). I have no particular opinion, but zeroing out the memory isn't required here.

src/h3lib/lib/linkedGeo.c
nrabinowitz: I can move it out. This is just a straight code move, though, the code hasn't changed from what it was.

src/apps/benchmarks/benchmarkH3SetToLinkedGeo.c
nrabinowitz: The `initLinkedPolygon` step just zeros out the memory in the struct. Maybe this is unnecessary - I was a little worried that after it was destroyed in one test run it would have bad pointers in it when it started the next run. I'll take it out and see if there are any issues.

src/h3lib/lib/algos.c
nrabinowitz: Will add. Were you thinking a user-facing doc comment, or just an inline comment?

src/apps/benchmarks/benchmarkH3SetToLinkedGeo.c
isaacbrodsky: My understanding is that h3SetToLinkedGeo will initialize the entire structure (`h3SetToLinkedGeo` calls `_vertexGraphToLinkedGeo`) before reading it. If this were the case, every use of `h3SetToLinkedGeo` would need to be paired with the appropriate init call.

src/h3lib/lib/algos.c
isaacbrodsky: I think inline comment. I don't see a way a user could use this comment.

src/apps/benchmarks/benchmarkH3SetToLinkedGeo.c
nrabinowitz: Actually this is called in the internals of `h3SetToLinkedGeo` and there is no need for it here. Dropping.

nrabinowitz:
@isaacbrodsky - all comments addressed, thanks for the review!

src/apps/testapps/testCompact.c
```diff
@@ -19,10 +19,10 @@
 #include "h3Index.h"
 #include "test.h"
 
-H3Index sunnyvale = 0x89283470c27ffffl;
+H3Index sunnyvale = 0x89283470c27ffff;
```
dfellis: Removing the long indicator on this constant will get you a different number (at least the last time I tested this it did). Try putting a `printf` statement with the H3Index in it and see if the number is the same.

isaachier: Really? I didn't have any issue here:

Source:
```
#include <inttypes.h>
#include <stdint.h>
#include <stdio.h>

int main(void) {
  uint64_t x = 0x89283470c27ffff;
  uint64_t y = 0x89283470c27ffffl;
  printf("0x%" PRIx64 ", 0x%" PRIx64 "\n", x, y);
  return 0;
}
```
Output:
`0x89283470c27ffff, 0x89283470c27ffff`

src/h3lib/lib/linkedGeo.c
isaachier: There's probably a clang-format setting for that. I'll try finding it.

src/h3lib/lib/algos.c
isaachier: There is a way to check that return values are not ignored in many compilers: ` __attribute__ ((warn_unused_result))`. If you do that on the declaration of `normalizeMultiPolygon`, you will get compiler warnings every time you compile this function. Not necessarily what you want, but something.

src/h3lib/lib/polygon.c
isaachier: Understood. Thanks for the effort.

src/apps/testapps/testCompact.c
nrabinowitz: I get the same results as @isaachier here, and I'm pretty sure the tests wouldn't pass if these weren't being interpreted correctly. Leaving with no suffix for now.

PR #132, opened 2018-08-24
Move coverage.sh.in to the scripts directory

isaacbrodsky:
I'm placing the generated script file into `scripts` because by default CMake does not have a build type defined. I moved the template into `scripts` to be consistent with `dev-docs/Doxyfile.in` generating `dev-docs/Doxyfile`.

The generated `coverage.sh` file, and `Debug` and `Release` directories are now gitignored.


isaachier:
I don't mind either way. The only potential problem is that it should not look like a real script. It is more like a template.

.gitignore
```diff
@@ -10,6 +10,9 @@
 /lib/
 # Travis CI build directory
 /build/
+# Local build directories
+/Debug/
+/Release/
```
isaachier: In theory, there's a few more than this: RelWithDebInfo, MinSizeRel.

PR #134, opened 2018-08-26
Use malloc when value is immediately overwritten

isaachier:
Tested this with Valgrind without any warnings/errors.


src/h3lib/include/linkedGeo.h
```diff
@@ -24,8 +24,24 @@
 #include "geoCoord.h"
 #include "h3api.h"
 
-void initLinkedPolygon(LinkedGeoPolygon* polygon);
-void initLinkedLoop(LinkedGeoLoop* loop);
+/**
+ * Initialize a linked polygon
+ * @param  polygon Polygon to init
+ */
+#define initLinkedPolygon(polygon)          \
```
isaacbrodsky: Why not something like the existing `H3_INIT`, so in code `*polygon = LINKED_GEO_POLYGON_INIT;`
Please make macros ALL_CAPS.

I assume this is for performance reasons. I'd suggest that performance changes should have associated benchmarking.

isaachier: Ya sure. Actually not really for performance. I just think this is easy enough to be a one-liner, not a separate function. Might try removing entirely.

isaachier:
Good idea. Will add there too.

src/h3lib/lib/h3Index.c
```diff
@@ -355,7 +355,7 @@ int H3_EXPORT(compact)(const H3Index* h3Set, H3Index* compactedSet,
         memset(hashSetArray, 0, numHexes * sizeof(H3Index));
         compactedSetOffset += uncompactableCount;
         memcpy(remainingHexes, compactableHexes,
-               maxCompactableCount * sizeof(H3Index));
+               compactableCount * sizeof(H3Index));
```
isaachier: This was causing a clang-tidy warning. It seems `maxCompactableCount` is less precise than `compactableCount` for specifying the size to copy. Relates to accessing array out of bounds.

src/h3lib/include/polygonAlgos.h
```diff
@@ -53,7 +53,7 @@
 
 /** Macro: Normalize longitude, dealing with transmeridian arcs */
 #define NORMALIZE_LON(lon, isTransmeridian) \
-    (isTransmeridian && lon < 0 ? lon + (double) M_2PI : lon)
+    (isTransmeridian && lon < 0 ? lon + (double)M_2PI : lon)
```
isaacbrodsky: This change shouldn't have been possible - the error is this file is not in `LIB_SOURCE_FILES` in CMakeLists.txt

isaachier: Oh OK I'll wait on that.

nrabinowitz:
Changes look good, thanks for dealing with all the merge conflicts!

isaachier:
Anytime :)

PR #138, opened 2018-08-29
Updated maxKringSize to closed form formula

temporalparts:
Source of closed form formula: https://oeis.org/A003215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ The public API of this library consists of the functions declared in file
 - Normalize output of h3SetToMultiPolygon to align with the GeoJSON spec, ensuring that each polygon has only one outer loop, followed by holes (#131)
 ### Changed
 - Longitude outputs are now guaranteed to be in the range [-Pi, Pi]. (#93)
-### Updated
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ The public API of this library consists of the functions declared in file
 - Normalize output of h3SetToMultiPolygon to align with the GeoJSON spec, ensuring that each polygon has only one outer loop, followed by holes (#131)
 ### Changed
 - Longitude outputs are now guaranteed to be in the range [-Pi, Pi]. (#93)
+### Updated
```
nrabinowitz: Nit: This should just go under `Changed` I think.

temporalparts: Well I put it under Updated because this mathematically does not change the output for k >= 0, but will put it under Changed.

isaacbrodsky:
> @isaacbrodsky - would it be worthwhile from a perf perspective to actually hard-code the integer sequence for k less than some threshold?

Maybe, but I would hesitate to add a microoptimization like that. Without benchmarking, I'd guess the possible performance gain is pretty small.

PR #142, opened 2018-08-31
Added benchmarks for kRing

temporalparts:


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The public API of this library consists of the functions declared in file
 [h3api.h](./src/h3lib/include/h3api.h).
 
 ## [Unreleased]
+### Changed
```
isaacbrodsky: I think this should be "Added"

src/apps/benchmarks/benchmarkH3Api.c
```diff
@@ -34,4 +34,11 @@ BENCHMARK(h3ToGeoBoundary, 10000, {
     H3_EXPORT(h3ToGeoBoundary)(hex, &outBoundary);
 });
 
+H3Index* out = malloc(H3_EXPORT(maxKringSize)(40) * sizeof(H3Index));
```
isaacbrodsky: Could you add a new file for this (`benchmarkKRing.c`)? If you're not comfortable editing `CMakeLists.txt` etc I could break this out to a new benchmark in a separate PR.

src/apps/benchmarks/benchmarkH3Api.c
```diff
@@ -34,4 +34,11 @@ BENCHMARK(h3ToGeoBoundary, 10000, {
...
+
+BENCHMARK(kRing10, 10000, { H3_EXPORT(kRing)(hex, 10, out); });
+BENCHMARK(kRing20, 10000, { H3_EXPORT(kRing)(hex, 20, out); });
+BENCHMARK(kRing30, 10000, { H3_EXPORT(kRing)(hex, 30, out); });
+BENCHMARK(kRing40, 10000, { H3_EXPORT(kRing)(hex, 40, out); });
+
```
isaacbrodsky: We should free `out`. It will be torn down by the OS immediately afterwards, but it helps let us run Valgrind and find other memory issues.

src/apps/benchmarks/benchmarkH3Api.c
nrabinowitz: Looks like we still need to remove these benchmarks from this file, right?

nrabinowitz: This might not have been clear in context, but the added benchmarks were appearing twice, once in `benchmarkH3Api.c` and once in `benchmarkKRing.c` - that's what I was blocking on.

PR #144, opened 2018-09-04
Add a benchmark for kRing on a pentagon

isaacbrodsky:
This case falls back to the DFS approach which is much slower
than non-pentagon cases.

### DIFF
--- a/src/apps/benchmarks/benchmarkKRing.c
+++ b/src/apps/benchmarks/benchmarkKRing.c
@@ -14,19 +14,28 @@
  * limitations under the License.
  */
 #include "benchmark.h"
+#include "h3Index.h"
 #include "h3api.h"
 
 // Fixtures
 H3Index hex = 0x89283080ddbffff;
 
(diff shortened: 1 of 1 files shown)


src/apps/benchmarks/benchmarkKRing.c
```diff
@@ -14,20 +14,29 @@
...
 #include "h3api.h"
 
 // Fixtures
 H3Index hex = 0x89283080ddbffff;
 
+H3Index pentagon;
+
 BEGIN_BENCHMARKS();
 
+setH3Index(&pentagon, 9, 4, CENTER_DIGIT);
```
nrabinowitz: This is fine, but I don't see any issue just using the indexes of "well-known pentagons" directly in tests. If you tell me a test address is a pentagon, I'll believe you :).

isaacbrodsky: I had to first calculate the index of the pentagon :)

PR #149, opened 2018-10-04
Version 3.2.0, add Local IJ to docs

isaacbrodsky:
The Distance docs are replaced with docs for the Local IJ group of functions.

The coordinate systems documentation page is converted to use the expected single asterisk for italics, rather than underscores which aren't understood by Ocular.


docs/api/localij.md
```diff
@@ -0,0 +1,44 @@
+# Local IJ coordinate functions
+
+These functions operate on a local IJ coordinate space. The
+coordinates are anchored by an **origin index**. The I and J
+axes are 120&deg; apart. The origin index might not be at
+(0, 0) in the local coordinate system.
+
+LOcal IJ coordinates are only valid within a certain radius of the origin.
```
dfellis: Typo: `LOcal` -> `Local`

docs/api/localij.md
```diff
@@ -0,0 +1,44 @@
...
+
+## experimentalH3ToLocalIj
+
+```
+int experimentalH3ToLocalIj(H3Index origin, H3Index h3, CoordIJ *out);
+```
+
+Produces local IJ coordinates for an index anchored by an origin.
+
+This function is experimental, and its output is not gauranteed
```
dfellis: `gauranteed` -> `guaranteed`

docs/api/localij.md
```diff
@@ -0,0 +1,44 @@
...
+
+## experimentalLocalIjToH3
+
+```
+int experimentalLocalIjToH3(H3Index origin, const CoordIJ *ij, H3Index *out);
+```
+
+Produces an index for local IJ coordinates anchored by an origin.
+
+This function is experimental, and its output is not gauranteed
```
dfellis: `gauranteed` -> `guaranteed`

docs/api/localij.md
```diff
@@ -0,0 +1,44 @@
...
+to be compatible across different versions of H3.
+
+## experimentalLocalIjToH3
+
+```
+int experimentalLocalIjToH3(H3Index origin, const CoordIJ *ij, H3Index *out);
+```
+
+Produces an index for local IJ coordinates anchored by an origin.
```
dfellis: Nit: Might read better as `Produces an H3 index from local IJ coordinates anchored by an origin.`

docs/api/localij.md
```diff
@@ -0,0 +1,44 @@
...
+```
+int experimentalLocalIjToH3(H3Index origin, const CoordIJ *ij, H3Index *out);
+```
+
+Produces an H3 index from local IJ coordinates anchored by an origin.
+
+This function is experimental, and its output is not guaranteed
+to be compatible across different versions of H3.
+
+## h3Distance
```
nrabinowitz: This is a little tricky. `h3Distance` is a generally applicable function that many users may want without understanding a concept of local coordinates. I'm a little worried that we're burying it here because its implementation is based on IJ coords, rather than putting it somewhere that would make more sense to the reader.

Would it make more sense, for example, to rename `Neighbors` to `Traversal` and put this there?

isaacbrodsky: I agree, organizing this is not easy. While they are not connected from a use perspective, these functions have a similar set of limitations, which I wanted to make the user aware of.

I feel like if `Neighbors` is renamed to `Traversal` that could also encompass the `Local IJ` functions. Maybe all of them should be on the same page?

nrabinowitz: Sure, I'd support that.

PR #158, opened 2018-11-19
Fix additional formatting on the docs site

isaacbrodsky:
* Incorrectly rendered italics are fixed. It seems as if not rendering `_r_` as _r_ is a bug in the Markdown parser dealing with short italics text since `_ijk_` does render correctly.
* Fixed `C-sharp` to `C#` via HTML entities.

There are still some misrendered math symbols on "Overview of the H3 Geospatial Indexing System".


isaacbrodsky:
Good idea @dfellis, applied that change to this PR.

Also formatted the term "H3" consistently.

PR #159, opened 2018-11-21
Formatting change to use MathJax in rendered Ocular docs

isaacbrodsky:
It looks to me like the need to write `\\(` is from some part of the Markdown parsing chain interpreting the backslashes along the way.


isaacbrodsky:
@nrabinowitz True, but my intention is we can expand the documentation to offer more comprehensive details about H3, which might benefit from MathJax.

PR #160, opened 2018-12-08
Add Ruby bindings to community docs

seanhandley:


nrabinowitz:
Thanks so much for submitting! Adding Ruby support would be great.

However, it looks like this binding isn't really implemented yet, unless I'm missing something. While it's great to let others know you're working on this, to avoid duplicate work, I'd prefer to leave this off the bindings page until it's production-ready. Thoughts?

seanhandley:
Hi @nrabinowitz

I think that's fair.

We've wrapped around 2/3 of the API calls and expect to have the rest done soon.

I'll update this PR when we release the first tagged version.

PR #165, opened 2018-12-13
Add hexLine function to return the line of indexes between a given start and end

### LINKED ISSUE #162: Add h3Line algorithm
Add a function `h3Line(start, end)` that returns the list of hexagons between the start and end indexes, inclusive.

I'm currently working on this.

### LINKED ISSUE #163: h3ToLocalIjk incorrectly succeeds across a pentagon
I'm working on the `h3Line` algorithm (#162), and my tests are running into issues when crossing a pentagon onto an adjacent base cell. The following image shows the results of `h3ToLocalIjk` from the res 3 cell `1, 0, 0` (next to the `0, 0, 0` pentagon):

screen shot 2018-12-11 at 5 49 07 pm

This takes a k-ring of 12 and shows only cells with valid coordinates from `h3ToLocalIjk`. The cells on the failing sides of the pentagon on the same base cell are excluded, but cells on the adjacent base cell across the pentagon are included (incorrectly, I believe).

Looks like the reason for this are the [Class III checks here](https://github.com/uber/h3/blob/36a2cebd1cd292cfcfe171626df666d1885c0f18/src/h3lib/lib/localij.c#L188). In the same-cell branch, both `FAILED_DIRECTIONS_II` and `FAILED_DIRECTIONS_III` fail, but in the cross-cell branch they depend on Class III status. Is there a reason for this? This pentagon is base cell 4, which seems to have a few special properties (no CW offset rotation faces in `baseCellData`), but I'm not sure if that's relevant here.

nrabinowitz:
* Adds the `hexLine` function and exposes it in the API. 
* Adds property-based tests for all hexagons in res 0-2 and some base cells in res 3.
* Adds `ijkToCube` and `cubeToIjk` helpers to translate from IJK to cube coordinate space

screen shot 2018-12-13 at 11 25 35 am

This uses the algorithm described on the [Red Blob Games site](https://www.redblobgames.com/grids/hexagons/#line-drawing), converting IJK to cube coordinates in order to use the `cubeRound` algo.

TODO:
- [x] Add to CHANGELOG
- [x] Add to docs

Fixes #163 
Closes #162

### DIFF
--- a/src/apps/testapps/testH3Line.c
+++ b/src/apps/testapps/testH3Line.c
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2018 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/** @file
+ * @brief tests H3 distance function.
+ *
+ *  usage: `testH3Distance`
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "h3Index.h"
+#include "h3api.h"
+#include "localij.h"
+#include "stackAlloc.h"
+#include "test.h"
+#include "utility.h"
+
+static const int MAX_DISTANCES[] = {1, 2, 5, 12, 19, 26};
+
+/**
+ * Property-based testing of h3Line output
+ */
+static void h3Line_assertions(H3Index start, H3Index end) {
+    int sz = H3_EXPORT(h3LineSize)(start, end);
+    t_assert(sz > 0, "got valid size");
+    STACK_ARRAY_CALLOC(H3Index, line, sz);
+
+    int err = H3_EXPORT(h3Line)(start, end, line);
+
+    t_assert(err == 0, "no error on line");
+    t_assert(line[0] == start, "line starts with start index");
+    t_assert(line[sz - 1] == end, "line ends with end index");
+
+    for (int i = 1; i < sz; i++) {
+        t_assert(H3_EXPORT(h3IsValid)(line[i]), "index is valid");
+        t_assert(H3_EXPORT(h3IndexesAreNeighbors)(line[i], line[i - 1]),
+                 "index is a neighbor of the previous index");
+        if (i > 1) {
+            t_assert(
+                !H3_EXPORT(h3IndexesAreNeighbors)(line[i], line[i - 3]),
+                "index is not a neighbor of the index before the previous");
+        }
+    }
+}
+
+/**
+ * Tests for invalid h3Line input
+ */
+static void h3Line_invalid_assertions(H3Index start, H3Index end) {
+    int sz = H3_EXPORT(h3LineSize)(start, end);
+    t_assert(sz < 0, "line size marked as invalid");
+
+    H3Index* line = {0};
+    int err = H3_EXPORT(h3Line)(start, end, line);
+    t_assert(err != 0, "line marked as invalid");
+}
+
+/**
+ * Test for lines from an index to all neighbors within a kRing
+ */
+static void h3Line_kRing_assertions(H3Index h3) {
+    int r = H3_GET_RESOLUTION(h3);
+    t_assert(r <= 5, "resolution supported by test function (kRing)");
+    int maxK = MAX_DISTANCES[r];
+
+    int sz = H3_EXPORT(maxKringSize)(maxK);
+    STACK_ARRAY_CALLOC(H3Index, neighbors, sz);
+    H3_EXPORT(kRing)(h3, maxK, neighbors);
+
+    if (H3_EXPORT(h3IsPentagon)(h3)) {
+        return;
+    }
+
+    for (int i = 0; i < sz; i++) {
+        if (neighbors[i] == 0) {
+            continue;
+        }
+        int distance = H3_EXPORT(h3Distance)(h3, neighbors[i]);
+        if (distance >= 0) {
+            h3Line_assertions(h3, neighbors[i]);
+        } else {
+            h3Line_invalid_assertions(h3, neighbors[i]);
+        }
+    }
+}
+
+SUITE(h3Distance) {
+    TEST(h3Distance_kRing) {
+        iterateAllIndexesAtRes(0, h3Line_kRing_assertions);
+        iterateAllIndexesAtRes(1, h3Line_kRing_assertions);
+        iterateAllIndexesAtRes(2, h3Line_kRing_assertions);
+        // Don't iterate all of res 3, to save time
+        iterateAllIndexesAtResPartial(3, h3Line_kRing_assertions, 6);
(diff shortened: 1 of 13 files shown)


src/h3lib/lib/localij.c
```diff
@@ -183,10 +183,8 @@ int h3ToLocalIjk(H3Index origin, H3Index h3, CoordIJK* out) {
         if (originOnPent) {
             int originLeadingDigit = _h3LeadingNonZeroDigit(origin);
 
-            if ((isResClassIII(res) &&
-                 FAILED_DIRECTIONS_III[originLeadingDigit][dir]) ||
-                (!isResClassIII(res) &&
-                 FAILED_DIRECTIONS_II[originLeadingDigit][dir])) {
```
nrabinowitz: This part fixes #163. The change may exclude valid cases, but I think it's okay to be more conservative here. Would like feedback from @isaacbrodsky if available on how the Class III check was expected to work, but I don't think that has to block this change.

src/apps/testapps/testH3Line.c
```diff
@@ -0,0 +1,110 @@
...
+        int distance = H3_EXPORT(h3Distance)(h3, neighbors[i]);
+        if (distance >= 0) {
+            h3Line_assertions(h3, neighbors[i]);
+        } else {
+            h3Line_invalid_assertions(h3, neighbors[i]);
+        }
+    }
+}
+
+SUITE(h3Distance) {
```
dfellis: Nit: this should be `h3Line`.

src/h3lib/lib/localij.c
```diff
@@ -542,3 +540,109 @@ int H3_EXPORT(h3Distance)(H3Index origin, H3Index h3) {
...
+ *  - Lines are drawn in grid space, and may not correspond exactly to either
+ *    Cartesian lines or great arcs.
+ *
+ * @param start Start index of the line
+ * @param end End index of the line
+ * @param out Output array, which must be of size h3LineSize(start, end)
+ * @return 0 on success, or another value on failure.
+ */
+int H3_EXPORT(h3Line)(H3Index start, H3Index end, H3Index* out) {
+    int distance = H3_EXPORT(h3Distance)(start, end);
```
dfellis: Unfortunate that this effectively has to be called twice (once for the mem allocation and once for the safeguard here).

Perhaps there can be two functions, here, one that takes the chunk of memory and it's size and one with this signature that takes just the chunk of memory, assumes that it's big enough, and calls `h3Distance` to determine the usable size (or fail out early).

This would allow h3Distance to conceivably be called only once by users wanting to squeeze as much performance out of these calls as they can (such as the potential `polyfill` algorithm to reduce memory consumption and maybe run quicker, too).

src/h3lib/lib/localij.c
```diff
@@ -542,3 +540,109 @@ int H3_EXPORT(h3Distance)(H3Index origin, H3Index h3) {
...
+        return distance;
+    }
+
+    // Get IJK coords for the start and end. We've already confirmed
+    // that these can be calculated with the distance check above.
+    CoordIJK startIjk = {0};
+    CoordIJK endIjk = {0};
+
+    // Convert H3 addresses to IJK coords
+    h3ToLocalIjk(start, start, &startIjk);
```
dfellis: Just a comment on our `experimentalH3ToLocalIj` transforms: It's unfortunate that the coordinates are not offset by the origin coordinates in the calculation, then you would only have to call `h3ToLocalIjk(start, end, &endIjk)` because `h3ToLocalIjk(start, start, &startIjk)` would've been defined as `(0, 0, 0)`. A bit less computation and the relative nature of the coordinate system would be more obvious to users.

@isaacbrodsky thoughts?

src/h3lib/lib/localij.c
dfellis: Definitely want to defer to @isaacbrodsky on this piece, but my gut feeling of applying direction blocks for both hexagon orientations is not the right solution. Either these two should be merged (because there is no difference in reality), or the algorithm behind the `FAILED_DIRECTIONS_*` lookup tables needs to be debugged more closely to see why it missed a path.

src/h3lib/lib/localij.c
nrabinowitz: Well, `h3Distance` is constant time and pretty fast, so I'm not sure this is really an issue. `polyfill`, `kRing`, etc all have the same problem. But if we care, then the simplest solution is to require the caller to pass in the size of the memory allocation:
```
H3_EXPORT(h3Line)(H3Index start, H3Index end, H3Index* out, int numIndexes);
```
All callers ought to call `h3LineSize` in any case, and can just pass the result in. The only cost is a slightly uglier API that's inconsistent with `kRing` and others.

src/h3lib/lib/localij.c
nrabinowitz: FWIW, I don't think it's any less computation, since you'd still be calculating `h3ToLocalIjk(start, start, &startIjk)` under the hood in order to get the coordinates for offset. In fact, it's probably worse, because calling `h3ToLocalIjk` repeatedly will perform the calculation of the offset every time - the only time you save work is when the `origin` is the same as the index you want to get coords for.

src/h3lib/lib/localij.c
```diff
@@ -542,3 +540,109 @@ int H3_EXPORT(h3Distance)(H3Index origin, H3Index h3) {
...
+    float iStep =
+        distance ? (float)(endIjk.i - startIjk.i) / (float)distance : 0;
+    float jStep =
+        distance ? (float)(endIjk.j - startIjk.j) / (float)distance : 0;
+    float kStep =
+        distance ? (float)(endIjk.k - startIjk.k) / (float)distance : 0;
+
+    CoordIJK currentIjk = {startIjk.i, startIjk.j, startIjk.k};
+    for (int n = 0; n <= distance; n++) {
+        cubeRound((float)startIjk.i + iStep * n, (float)startIjk.j + jStep * n,
```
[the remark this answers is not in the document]
nrabinowitz: No real reason other than thinking we didn't need the precision. Will update.

src/h3lib/lib/localij.c
dfellis: Well, if `h3Distance` doesn't add a significant amount of time, then we don't have to worry about it.

src/h3lib/lib/localij.c
dfellis: `h3ToLocalIjk` includes several branches and loops. I'd assume that three subtractions would be minimal in comparison to the rest of the function, and it would let you call that function once instead of twice *and* produce coordinates that are simpler to reason about.

nrabinowitz: But you'd need to get the coordinates to subtract, which is effectively an `h3ToLocalIjk` call on your origin, right? Right now, the origin is just used to get the base cell, and then the IJK is calculated for the target; if you offset by the origin, you'd need to calculate the IJK for the origin, the IJK for the target, and then translate by the origin. It's the additional IJK calculation that adds the perf impact, not the translation.

I could be wrong here, but pretty sure that in a case like `h3Line`, where you are finding the coordinates for indexes `1, 2, 3, ...N`, you'd have `N - 1` calls instead of the `N` we have now, but you'd actually be doing `(N - 1) * 2` IJK calculations, instead of the `N` we do now.

src/h3lib/lib/localij.c
dfellis: Since @isaacbrodsky can't look into it at the moment, could we add a comment to revisit this problem in the future? (The only downside is that it might prevent usage of this new function in situations it would be valid in, but since there is no general-purpose fallback here even for cases where this would have failed, that doesn't really change the perceptible end-user behavior.)

nrabinowitz: Makes sense, I will add a TODO comment here explaining what was removed.

PR #173, opened 2019-01-02
ENH: Include defined constants for current library version

### LINKED ISSUE #148: Include defined constants for current library version in h3Api header
Per discussion in #104:

> it would be great to have defines for the version number in the h3api.h header file. For example something like
> 
> ```
> #define H3_VERSION_MAJOR 4
> #define H3_VERSION_MINOR 0
> #define H3_VERSION_PATCH 0
> ```
> would really help projects depending on this library to deal with the API-change.

Ideally this would generated from the `VERSION` file.

zachasme:
Resolves #148.

Splits version string in CMakeLists.txt and configures a new header file, `version.h`, with number constants.


zachasme:
Hm, I've added CMAKE_BINARY_DIR as include directory in a force push to resolve the build errors. Does force-pushing mess up CI?

zachasme:
Sorry, I didn't mean to close the pull request. Back from holidays and all that.

nrabinowitz:
Thanks for the contribution! It looks like all the build jobs are failing with some variation of 
```
Scanning dependencies of target format
Formatting sources
No such file or directory
make[3]: *** [CMakeFiles/format] Error 1
make[2]: *** [CMakeFiles/format.dir/all] Error 2
make[1]: *** [CMakeFiles/format.dir/rule] Error 2
make: *** [format] Error 2
```
Maybe `make format` is trying to run on the `version.h` file that hasn't been built yet?

zachasme:
You were right, I hadn't installed `clang-format`. I've removed `version.h` and configure `h3api.h` directly instead. Alternatively you'll have to also install `version.h`. I'm not sure what is preferable.

zachasme:
This will be very helpful for ensuring the correct version of `h3.so` in bindings, by the way.

.gitignore
```diff
@@ -96,3 +96,6 @@ binding-functions
 # These KML files are generated by `make kml`
 KML/res*cells.kml
 KML/res*centers.kml
+
+# Generated header files
+src/h3lib/include/version.h
```
isaacbrodsky: I think this line is no longer needed?

CMakeLists.txt
```diff
@@ -99,7 +103,6 @@ set(LIB_SOURCE_FILES
     src/h3lib/include/constants.h
     src/h3lib/include/coordijk.h
     src/h3lib/include/algos.h
-    src/h3lib/include/h3api.h
```
isaacbrodsky: I haven't tested this, but will this still cause clang format to auto format h3api.h?

zachasme: Good catch, it wont be auto formatted.

nrabinowitz: Should this line still be removed?

zachasme: It's now autoformatting h3api.h.in

zachasme: There might be a smarter way, the line is removed such that cmake doesn't try to format a non-existent file. The generated h3api.h is still added to the library (using the var `CONFIGURED_API_HEADER`).

nrabinowitz: Ok, thanks!

zachasme:
I've added a test, which failed, and fixed the formatting error causing the problem.

PR #176, opened 2019-01-04
Add function to get all base cells

### LINKED ISSUE #174: Add function to get all base cells
You can get all base cells now by taking a 10-ring of some arbitrary base cell, but that's a bit hidden - it would be nice to offer all 122 base cells in order, through a function like `getBaseCells`.

zachasme:
Resolves #174.

You could instead export a defined constant with all 122 base cells. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ The public API of this library consists of the functions declared in file
 
 ## [Unreleased]
 ### Added
-- Include defined constants for current library version
+- `getBaseCells` function for getting all base cells (#174)
+- Include defined constants for current library version (#173)
 
 ## [3.3.0] - 2018-12-25
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The public API of this library consists of the functions declared in file
 
 ## [Unreleased]
 ### Added
-- `getBaseCells` function for getting all base cells (#174)
+- `getRes0Indexes` function for getting all base cells, and helper function `res0IndexCount` (#174)
 - Include defined constants for current library version (#173)
 
 ## [3.3.0] - 2018-12-25
--- a/src/apps/testapps/testBaseCells.c
+++ b/src/apps/testapps/testBaseCells.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 Uber Technologies, Inc.
+ * Copyright 2017-2019 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,9 +21,10 @@
 
 SUITE(baseCells) {
     TEST(getBaseCells) {
-        H3Index* baseCells = calloc(NUM_BASE_CELLS, sizeof(H3Index));
-        H3_EXPORT(getBaseCells)(baseCells);
-        t_assert(baseCells[0] == 0x8001fffffffffff, "correct first basecell");
-        t_assert(baseCells[121] == 0x80f3fffffffffff, "correct last basecell");
+        H3Index* indexes = malloc(NUM_BASE_CELLS * sizeof(H3Index));
+        H3_EXPORT(getRes0Indexes)(indexes);
+        t_assert(indexes[0] == 0x8001fffffffffff, "correct first basecell");
+        t_assert(indexes[121] == 0x80f3fffffffffff, "correct last basecell");
+        free(indexes);
     }
 }
--- a/src/apps/testapps/testBaseCells.c
+++ b/src/apps/testapps/testBaseCells.c
@@ -15,7 +15,6 @@
  */
 
 #include <stdlib.h>
-#include "constants.h"
 #include "h3api.h"
 #include "test.h"
 
--- a/src/apps/testapps/testBaseCells.c
+++ b/src/apps/testapps/testBaseCells.c
@@ -21,7 +21,7 @@
 
 SUITE(baseCells) {
     TEST(getBaseCells) {
-        int count = res0IndexCount();
+        int count = H3_EXPORT(res0IndexCount)();
         H3Index* indexes = malloc(count * sizeof(H3Index));
         H3_EXPORT(getRes0Indexes)(indexes);
         t_assert(indexes[0] == 0x8001fffffffffff, "correct first basecell");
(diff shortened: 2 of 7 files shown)


src/apps/testapps/testBaseCells.c
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2017-2018 Uber Technologies, Inc.
```
isaacbrodsky: 2019

src/apps/testapps/testBaseCells.c
```diff
@@ -0,0 +1,29 @@
...
+#include "h3api.h"
+#include "test.h"
+
+SUITE(baseCells) {
+    TEST(getBaseCells) {
+        H3Index* baseCells = calloc(NUM_BASE_CELLS, sizeof(H3Index));
+        H3_EXPORT(getBaseCells)(baseCells);
+        t_assert(baseCells[0] == 0x8001fffffffffff, "correct first basecell");
+        t_assert(baseCells[121] == 0x80f3fffffffffff, "correct last basecell");
+    }
```
isaacbrodsky: We should free baseCells, since it can help locate actual leaks using Valgrind a little easier.

Nit: we could use malloc since we don't need it zeroed, but that's definitely not a big deal.

src/h3lib/include/h3api.h
```diff
@@ -275,6 +275,14 @@ double H3_EXPORT(edgeLengthM)(int res);
 int64_t H3_EXPORT(numHexagons)(int res);
 /** @} */
 
+/** @defgroup getBaseCells getBaseCells
+ * Functions for getBaseCells
+ * @{
+ */
+/** @brief provides all base cells */
+void H3_EXPORT(getBaseCells)(H3Index *out);
```
isaacbrodsky: I am wondering about the name of this function, since in the library a "base cell" usually refers to an integer identifier 0-122. Maybe getRes0Indexes or something?

src/h3lib/lib/baseCells.c
```diff
@@ -886,3 +887,18 @@ Direction _getBaseCellDirection(int originBaseCell, int neighboringBaseCell) {
     }
     return INVALID_DIGIT;
 }
+
+/**
+ * getBaseCells generates all base cells storing them into the provided
+ * memory pointer.
+ *
+ * @param out H3Index* the memory to store the resulting base cells in
```
isaacbrodsky: We should document how large this buffer needs to be - if NUM_BASE_CELLS is needed, that perhaps should be in h3api.h.

zachasme:
Thank you for the comments, I've updated the pull request accordingly.

What are your thoughts on exporting a defined constant with the base cells instead?

src/h3lib/lib/baseCells.c
nrabinowitz: I'm not sure the JS build can currently access constants, even if they're in `h3api.h`. I think the version that would be most consistent with other pass-in-allocated-memory functions would be a new function `res0IndexCount` or similar, which would return `NUM_BASE_CELLS`.

nrabinowitz:
Thanks for the submission! I personally prefer the function, particularly because the JS binding can't access constants directly (though as that's a binding issue we could resolve in the binding if the constant was a better option). Honestly though the generator function is likely so fast that there would be very little benefit to a constant.

src/apps/testapps/testBaseCells.c
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#include <stdlib.h>
+#include "constants.h"
+#include "h3api.h"
+#include "test.h"
+
+SUITE(baseCells) {
+    TEST(getBaseCells) {
+        int count = res0IndexCount();
```
nrabinowitz: This line needs `H3_EXPORT(res0IndexCount)` now.

src/apps/testapps/testBaseCells.c
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include "constants.h"
```
nrabinowitz: include still needed?

src/apps/testapps/testBaseCells.c
zachasme: Good catch!

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The public API of this library consists of the functions declared in file
 [h3api.h](./src/h3lib/include/h3api.h).
 
 ## [Unreleased]
+### Added
+- `getBaseCells` function for getting all base cells (#174)
```
nrabinowitz: Whoops, this should now be `getRes0Indexes` (and you might want to add a note about `res0IndexCount` as well)

src/apps/testapps/testBaseCells.c
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include "h3api.h"
+#include "test.h"
+
+SUITE(baseCells) {
+    TEST(getBaseCells) {
```
nrabinowitz: Nit: `getRes0Indexes`

src/h3lib/include/h3api.h
```diff
@@ -275,6 +275,17 @@ double H3_EXPORT(edgeLengthM)(int res);
 int64_t H3_EXPORT(numHexagons)(int res);
 /** @} */
 
+/** @defgroup getBaseCells getBaseCells
+ * Functions for getBaseCells
```
nrabinowitz: Sorry, this too - `getRes0Indexes`

zachasme:
I've rebased and addressed the comments.

nrabinowitz:
Looks great. Thanks so much for adding this!

seanhandley:
@nrabinowitz I notice there's no doc (yet) for this. Would this be classed as a miscellaneous function?

nrabinowitz:
Yes, realized this after I pushed the docs - I think miscellaneous is probably correct. Am I correctly inferring that you'll take this on?

nrabinowitz:
@seanhandley Logged as #182, please note there if you're likely to add, otherwise I'll put up a PR.

seanhandley:
Sure I'll do it 👍

PR #177, opened 2019-01-08
Add publish-website script

nrabinowitz:
* Moves steps in `website/README` to a script
* Adds a `make` target for the script
* Updates the release steps to include publishing the website


scripts/publish_website.sh
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script builds the current version of the H3 documentation website
+# and pushes it to production. This assumes that you have set the remote
+# for gh-pages to the upstream (i.e. uber/h3) remote.
+
+git checkout gh-pages
+git pull
```
dfellis: Nit: `git pull origin gh-pages` for those that have the newer default config to not guess which remote and branch to pull from.

scripts/publish_website.sh
```diff
@@ -0,0 +1,29 @@
...
+git checkout master
+pushd website
+./scripts/build-to-gh-pages.sh
+git push
```
dfellis: Nit: Same as above.

scripts/publish_website.sh
nrabinowitz: Well, part of the issue here is that my remote in this context is actually `upstream`, not `origin` - `origin` is my fork on GH, `upstream` is the `uber/h3` repo. Just using `git pull` and requiring the caller to set the upstream seemed like a reasonable way not to hard-code this (and git will prompt you to set the upstream if not set).

PR #186, opened 2019-02-12
Add link to isaacbrodsky/h3measurements

isaacbrodsky:


docs/community/tutorials.md
```diff
@@ -2,6 +2,10 @@
 
 This page lists further learning materials and code walkthroughs for the H3 library and bindings. Contributions to this list are welcome, please feel free to open a [pull request](https://github.com/uber/h3/tree/master/docs/community/tutorials.md).
 
+## Java
+
+- [H3 Measurements](https://github.com/isaacbrodsky/h3measurements)
```
nrabinowitz: Worth adding a one-line description on what this repo covers?

isaacbrodsky: Added a short description - I'm not sure if we should put this under "Java", because I assume most readers would be interested in the measurements themselves.

nrabinowitz: I'm fine with this being under Java for now - potentially we can reorganize this by something other than language when we have more entries.

PR #188, opened 2019-02-13
Fix binding-functions when building out of source

isaacbrodsky:
The binding functions scripts were looking in the original source directory, rather than for the generated `h3api.h` which is now under the binary directory. Also updated `binding_functions.sh` to fail the build when the file it expects is not found.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,10 +39,12 @@ environment:
 build_script:
   - cmake "-G%GENERATOR%" -H. -Bbuild
   - cmake --build build --config "%CONFIG%"
-  # Intentionally failing the build to check the test
-  #- cmake --build build --config "%CONFIG%" binding-functions
+  - cmake --build build --config "%CONFIG%" --target binding-functions
 
 test_script:
   - ps: cd build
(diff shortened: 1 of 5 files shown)


nrabinowitz:
This looks good, but can we add a Travis test that verifies `binding_functions` can be built correctly?

appveyor.yml
```diff
@@ -39,7 +39,10 @@ environment:
 build_script:
   - cmake "-G%GENERATOR%" -H. -Bbuild
   - cmake --build build --config "%CONFIG%"
+  # Intentionally failing the build to check the test
+  #- cmake --build build --config "%CONFIG%" binding-functions
```
nrabinowitz: Should this still be here? Does leaving it commented out imply that the test should fail but doesn't?

isaacbrodsky: The test should fail, but it just emits an error message and marks it as succeeded anyways. I'd like to take a look at that tomorrow and see if I can get it working before merging.

isaacbrodsky: I updated the PR so Appveyor also tests the binding-functions target.

PR #193, opened 2019-02-21
binding-functions encoded in ASCII on Windows

isaacbrodsky:
The default is for Powershell to generate UTF-16 with BOM, and setting to UTF-8
also generates a BOM. (I understand newer versions of Powershell have a UTF8
with no BOM option, but that wasn't available for me.) The BOM is a problem
because Java for example does not discard the BOM, requiring workarounds like
Commons IO.

This limits function names in the C API to the ASCII character set, which
currently appears to be a reasonable limitation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@ The public API of this library consists of the functions declared in file
 [h3api.h.in](./src/h3lib/include/h3api.h.in).
 
 ## [Unreleased]
+### Changed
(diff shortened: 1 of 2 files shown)


isaacbrodsky:
I confirmed in H3-Java / Appveyor that this does work: https://github.com/isaacbrodsky/h3-java/tree/bump-3.4.0-test-with-newer-h3

PR #196, opened 2019-02-23
Added brew formula

robsalasco:
Hi!

I added the library to brew so the installation now is a lot easier!

Best, Roberto.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ sudo apt install clang-format cmake-curses-gui lcov doxygen
 First make sure you [have the developer tools installed](http://osxdaily.com/2014/02/12/install-command-line-tools-mac-os-x/) and then
 
 ```
-# Installing the bare build requirements
-brew install cmake
+# Installing the library
+brew install h3
 # Installing useful tools for development
 brew install clang-format lcov doxygen
 ```


README.md
```diff
@@ -44,8 +44,8 @@ sudo apt install clang-format cmake-curses-gui lcov doxygen
 First make sure you [have the developer tools installed](http://osxdaily.com/2014/02/12/install-command-line-tools-mac-os-x/) and then
 
 ```
-# Installing the bare build requirements
-brew install cmake
+# Installing the library
+brew install h3
```
nrabinowitz: This is awesome, but I'm not sure this is the right part of the README - maybe we ought to put this up higher, under `Installing`? The current "Building from source" section is really more developer-focused, and developers will probably want to make sure they have `cmake` installed. The new formula makes installation (particularly of the command-line tools) much easier, so it should probably go higher up.

nrabinowitz:
Thanks for the contribution! This is great.

@isaacbrodsky This will greatly reduce friction in using the CLI utils, implying that we ought to focus more on making sure they're stable/tested/supported.

isaacbrodsky:
> @isaacbrodsky This will greatly reduce friction in using the CLI utils, implying that we ought to focus more on making sure they're stable/tested/supported.

Yes, that is definitely something I'd like to improve further.

nrabinowitz:
@robsalasco - Can you please sign the Contributor License Agreement? We can't actually merge this unless that's signed. Thanks!

robsalasco:
@nrabinowitz signed!

nrabinowitz:
Merging as is, I'll address my comment in a subsequent PR. Thanks!

PR #198, opened 2019-02-27
Allocate memory for benchmarkPolyfill on the heap

### LINKED ISSUE #197: benchmarkPolyfill.exe crashing on windows
Other benchmarks compile run correctly, Polyfill crashing
Windows 10/64, Visual Studio 2017

isaacbrodsky:
Fixes #197 

It looks like it's failing as it allocates something on the stack (usually for procedure entry, in this case for stack allocation), pointing to a problem with `STACK_ARRAY_CALLOC`. I changed the benchmark to allocate on the heap instead. This does mean the benchmark now also covers allocating the memory on the heap, that could be removed from the benchmark by moving the call to `maxPolyfillSize` outside the benchmark.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The public API of this library consists of the functions declared in file
 [h3api.h.in](./src/h3lib/include/h3api.h.in).
 
 ## [Unreleased]
+### Fixed
+- `benchmarkPolyfill` allocates its memory on the heap (#198)
 
 ## [3.4.2] - 2019-02-21
 ### Changed
--- a/src/apps/benchmarks/benchmarkPolyfill.c
+++ b/src/apps/benchmarks/benchmarkPolyfill.c
@@ -16,7 +16,6 @@
 #include "algos.h"
 #include "benchmark.h"
 #include "h3api.h"
-#include "stackAlloc.h"
 
 // Fixtures
 GeoCoord sfVerts[] = {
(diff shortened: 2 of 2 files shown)


apiszcz:
I left a question on h3-py issues, is there planned windows support for the h3 python wrapper?

https://github.com/uber/h3-py/issues/27

dfellis:
> Fix looks fine - does this indicate a general problem with `STACK_ARRAY_CALLOC` that we need to address?

Allocating on a stack is always going to be a bit of a wild card -- you're eating into same memory you use to make function calls, so the amount you can safely allocate varies based on how deep in the stack you are. (Apparently 64-bit Windows consumes more of the stack before your program runs than 32-bit Windows and Unix-like OSes.) It's okay for small memory allocations because it can be faster than heap and is similar in a sense to just more arguments to your function, but for dynamic memory allocations it's not a great idea.

We already removed usage of `STACK_ARRAY_CALLOC` from `libh3`, but still use it in some of the benchmarks and other apps.

@isaacbrodsky I wasn't going to bring this up, but does this change remove the last calls to `STACK_ARRAY_CALLOC`?

dfellis:
> Great news, on replicating the error and change.
> 
> Separate but somewhat related question:
> I left a question on h3-py issues, is there planned windows support for the h3 python wrapper?
> 
> [uber/h3-py#27](https://github.com/uber/h3-py/issues/27)

That's more of a question for @TengHu but it's definitely doable.

I recently added Windows support to [h3-node](http://github.com/dfellis/h3-node) so if my free time permits I could tackle that.

isaacbrodsky:
> @isaacbrodsky I wasn't going to bring this up, but does this change remove the last calls to `STACK_ARRAY_CALLOC`?

It is not: it's still used in some tests. I think we could entirely remove it (and supporting build code), since that may help eliminate this type of issue in the future.

nrabinowitz:
> I think we could entirely remove it (and supporting build code), since that may help eliminate this type of issue in the future.

👍

Issue #212, opened 2019-03-27
Unexpected cell boundaries at resolution 7


billyauhk:
Found three strange cells and just verify these with ./h3ToGeoBoundary
Both cells are in Southern Atlantic and numerical error is suspected? All of them involve -52.010831568 10.356241923 which is way off from the intended resolution 7 size.

87dc6d364ffffff
{
   -52.013053368 -34.623293134
   -52.004115638 -34.609673316
   -51.992961023 -34.616515715
   -51.990741057 -34.636968000
   -51.999673873 -34.650589653
   -52.010831568 10.356241923
}

87dc6dacaffffff
{
   -52.033151282 -34.630078657
   -52.024211685 -34.616448895
   -52.013053368 -34.623293134
   -52.010831568 10.356241923
   -52.019766250 -34.657388784
   -52.030927647 -34.650554494
}

87dc6dacbffffff
{
   -52.019766250 -34.657388784
   -52.010831568 10.356241923
   -51.999673873 -34.650589653
   -51.997447785 -34.671043766
   -52.006377553 -34.684677180
   -52.017538323 -34.677854666
}

nrabinowitz:
Thanks for the report! Confirmed in C core and JS. It looks like the bad vertex is also shared with children at res 8, e.g. `87dc6dacaffffff`, `88dc6daca3fffff`, etc., so this is not limited to Class III resolutions.

nrabinowitz:
It looks like the bad vertex is shared - so e.g. `88dc6dacb5fffff`, `87dc6dacaffffff`, and `87dc6dacbffffff` all share one incorrect vertex. The error seems to be longitude only, at least in these cases.

isaacbrodsky:
It looks like this might be a bug in `_geoAzDistanceRads ` around line 252 https://github.com/uber/h3/blob/f9be0f9ffca923f50b96869188fcf0903c36b338/src/h3lib/lib/geoCoord.c#L252 For the vertex with the issue it hits that branch, for other vertices for the index `87dc6d364ffffff` it's not hit. I would assume it should be constraining `coslon` if `coslon > 1.0`.

isaacbrodsky:
Thanks @billyauhk for reporting this! The bug should be fixed in the next version of the H3 library released. Please let us know if you see any other unexpected cell vertices.

sahrk:
@billyauhk Thanks from me as well! This was a copy-and-paste from my DGGRID, where the same bug appears to have existed undiscovered for the last 15 years.

PR #213, opened 2019-03-27
Constrain `coslon` in `_geoAzDistanceRads`

isaacbrodsky:
#212 

This needs additional tests to cover all the branches in this function. Lcov doesn't report it, but the branches around lines 250~253 are uncovered. Lcov only reports lines, so we don't know about untested code like this. If we can't have the coverage tool we use report branch coverage, we should consider setting the formatter to put the branch condition always on a new line.

It looks like the vertex issue is likely a copy+paste error resulting in `coslon` being tested but `sinlon` being constrained. After fixing the index `87dc6d364ffffff` and it's neighbors look right.

Unrelated to this, I noticed that we are using long double constants in this file, which I don't think is necessary since the library was changed to use doubles. I changed the constants in this function but not the other places in the library.

### DIFF
--- a/src/h3lib/lib/geoCoord.c
+++ b/src/h3lib/lib/geoCoord.c
@@ -221,36 +221,36 @@ void _geoAzDistanceRads(const GeoCoord* p1, double az, double distance,
         if (fabs(p2->lat - M_PI_2) < EPSILON)  // north pole
         {
             p2->lat = M_PI_2;
-            p2->lon = 0.0L;
+            p2->lon = 0.0;
         } else if (fabs(p2->lat + M_PI_2) < EPSILON)  // south pole
         {
             p2->lat = -M_PI_2;
-            p2->lon = 0.0L;
+            p2->lon = 0.0;
         } else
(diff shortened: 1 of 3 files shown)


src/h3lib/lib/geoCoord.c
```diff
@@ -221,36 +221,36 @@ void _geoAzDistanceRads(const GeoCoord* p1, double az, double distance,
         if (fabs(p2->lat - M_PI_2) < EPSILON)  // north pole
         {
             p2->lat = M_PI_2;
-            p2->lon = 0.0L;
+            p2->lon = 0.0;
```
nrabinowitz: What if anything is the affect of removing the `L` suffix? Do we lose precision? I'm guessing no, just wanted to understand if this is expected to be a no-op.

isaacbrodsky: `p2->lon` is a double, so this should have the same result.

The comparisons using long doubles are different, I assume in those cases the double variables are promoted to long doubles for the purpose of comparison. I believe that could possibly have issues in edge cases.

Issue #214, opened 2019-03-28
Full branch test coverage


isaacbrodsky:
As mentioned in #213, the branches were not covered by our existing tests. Lcov (at least in the setting we use) reports only source code line coverage, but not branch coverage.

We should either have lcov or another tool report branch coverage, or if not possible have the formatter enforce that statements can't be on the same line as the if statement to enforce branch coverage.

PR #215, opened 2019-03-28
Add branch coverage

isaacbrodsky:
See #214

This change makes `lcov`/`genhtml` report the branch coverage. Currently it reports 92.0% (971 of 1056 branches) locally. From taking a look at `linkedGeo.c` it seems to be reporting non-hit assertions as uncovered, which may be something we want to fix up later. `geoCoord.c` correctly shows some of the reachable branches as uncovered.


isaacbrodsky:
For some reason, perhaps lcov/compiler differences, the coverage I got doesn't match what Travis shows: 92.9% (974 of 1048 branches)

PR #222, opened 2019-04-12
Add local IJ filters

isaacbrodsky:
This PR adds a new filter application, and adds a new option to `kRing` filter (to cover the `kRingDistances` function).

We should begin thinking about how to better handle command line arguments and help texts. We might want to consider either adding our own simple options library or depending on another one.

There's also a small code cleanup to not hard code the polar pentagon base cell identifiers.

### DIFF
--- a/src/apps/filters/h3ToLocalIj.c
+++ b/src/apps/filters/h3ToLocalIj.c
@@ -21,13 +21,13 @@
  *
  *  The program reads H3 indexes from stdin and outputs the corresponding
  *  IJ coordinates to stdout, until EOF is encountered. `NA` is printed if the
- * IJ coordinates could not be obtained.
+ *  IJ coordinates could not be obtained.
  *
  *  `origin` indicates the origin (or anchoring) index for the IJ coordinate
  *  space.
  *
  *  This program has the same limitations as the `experimentalH3ToLocalIj`
- * function.
+ *  function.
  */
 
 #include <inttypes.h>
--- a/src/apps/filters/kRing.c
+++ b/src/apps/filters/kRing.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 Uber Technologies, Inc.
+ * Copyright 2016-2017, 2019 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


src/apps/filters/h3ToLocalIj.c
```diff
@@ -20,9 +20,8 @@
  *  usage: `h3ToLocalIj [origin]`
  *
  *  The program reads H3 indexes from stdin and outputs the corresponding
- *  IJ coordinates to stdout, until EOF is encountered. The H3 indexes
- *  should be in integer form. `NA` is printed if the IJ coordinates
- *  could not be obtained.
+ *  IJ coordinates to stdout, until EOF is encountered. `NA` is printed if the
+ * IJ coordinates could not be obtained.
```
nrabinowitz: Nit: Indentation?

src/apps/filters/kRing.c
```diff
@@ -17,43 +17,57 @@
...
 
     int k = 0;
-    if (argc > 1) {
-        if (!sscanf(argv[1], "%d", &k)) error("k must be an integer");
+    if (!sscanf(argv[1], "%d", &k)) error("k must be an integer");
+
+    int printDistances = 0;
+    if (argc > 2) {
+        if (!sscanf(argv[2], "%d", &printDistances))
+            error("printDistances must be an integer");
```
nrabinowitz: This seems weird? It's a boolean flag, so ideally we'd want something like `--printDistances` or `-p`. This would be valid for values > 1 as well, which seems odd. Any reason not to use `strcmp` here and have more descriptive flags?

src/apps/filters/localIjToH3.c
```diff
@@ -0,0 +1,79 @@
...
+ */
+/** @file
+ * @brief stdin/stdout filter that converts from local IJ coordinates to
+ * H3 indexes. This is experimental.
+ *
+ *  usage: `localIjToH3 [origin]`
+ *
+ *  The program reads IJ coordinates from stdin and outputs the corresponding
+ *  H3 indexes to stdout, until EOF is encountered. `NA` is printed if the H3
+ * index could not be obtained.
```
nrabinowitz: Nit, here and L30: Indentation

src/apps/filters/localIjToH3.c
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/** @file
+ * @brief stdin/stdout filter that converts from local IJ coordinates to
+ * H3 indexes. This is experimental.
+ *
+ *  usage: `localIjToH3 [origin]`
+ *
+ *  The program reads IJ coordinates from stdin and outputs the corresponding
```
nrabinowitz: Should this specify the format you expect for `IJ` coords?

src/apps/filters/localIjToH3.c
```diff
@@ -0,0 +1,79 @@
...
+int main(int argc, char *argv[]) {
+    // check command line args
+    if (argc != 2) {
+        fprintf(stderr, "usage: %s [origin]\n", argv[0]);
+        exit(1);
+    }
+
+    H3Index origin;
+
+    if (!sscanf(argv[1], "%" PRIx64, &origin))
```
nrabinowitz: Can we use `H3_EXPORT(stringToH3)` here?

src/apps/filters/localIjToH3.c
```diff
@@ -0,0 +1,79 @@
...
+        if (!fgets(buff, BUFF_SIZE, stdin)) {
+            if (feof(stdin))
+                break;
+            else
+                error("reading IJ coordinates from stdin");
+        }
+
+        CoordIJ ij;
+        if (!sscanf(buff, "%d %d", &ij.i, &ij.j))
+            error("parsing IJ coordinates");
```
nrabinowitz: This should probably reiterate the expected input format.

src/apps/filters/kRing.c
isaacbrodsky: I'll address this in a second PR since adding that functionality is a little complicated. For the moment this is using the same pattern as selecting KML output with positional arguments.

Issue #42, opened 2018-04-15
naming "k-ring"


sahrk:
Internally H3 uses an IJK hex coordinate system, where each hex coordinate has a k-component. So it seems confusing to me to use "k-ring" to refer to an unrelated concept. Have you considered changing to a different label for that concept?

I have mostly dealt with the similar concept of the ring of hexagons at exactly the same metric distance from a central cell. I used to call this an "r-ring", for "radius". But r is too convenient for grid resolution, so I switched to "d-ring", for distance. Lately I've been moving towards referring to grid precision, rather than resolution, as I think that more precisely expresses the situation from the standpoint of the data structure. Then p for "precision" leaves r available for "r-ring". But alas I was still on the fence on resolution/precision when I worked on H3, so I used resolution in naming. But I offer up those alternatives for k-ring; I'd synchronize my terminology with H3 going forward if you chose something other than k-ring.

isaacbrodsky:
Could "radius ring" be simplified to "ring" (hollow ring) and "rings" (existing k-ring)? In function parameters, `k` could be replaced with `radius`.

If we do rename things, we should clear up the naming of the `kRing`, `hexRing`, and `hexRange` functions. `hexRing` is the only one of those that produces a hollow ring.

dfellis:
I will note that this would be a semver major version change. Would recommend keeping the old names as aliases and then dropping them when "enough" breaking changes have been proposed, or we'll outpace Chrome's versioning.

sahrk:
I had to read the documentation several times to clarify the differences between the different “Neighbor traversal functions”. First note that normally the hex ring at `k` distance from a central hex should have sides of length `k`. Assuming that’s the semantics of `hexRing` then it might make sense to align the description in the documentation with the other functions, which all are defined in terms of distance from the center.

@isaacbrodsky I’m OK with your suggestion of “ring” for the hollow ring (current `hexRing`), but I think “rings” is a better fit for `hexRange` than `kRing`, since it looks like `hexRange` is ordered by rings of increasing distance, while `kRing` doesn’t guarantee ring-ordering in the output. For `kRing` I’d suggest one of the following: `region`, `vicinity`, `neighborhood`, `zone`, or `sphere`.

[...]

PR #227, opened 2019-05-01
Argument parsing for h3ToGeo, h3ToGeoBoundary, geoToH3

isaacbrodsky:
Continues from #224 to rewrite argument parsing for these three filter applications.


src/apps/applib/lib/utility.c
```diff
@@ -68,12 +68,12 @@ int parseArgs(int argc, char* argv[], int numArgs, Arg* args[],
     const char* errorMessage = NULL;
     const char* errorDetails = NULL;
 
-    int failed = _parseArgsList(argc, argv, 4, args, helpArg, &errorMessage,
```
nrabinowitz: This was a bug from the last diff?

isaacbrodsky: Yes, this was a copy paste error when refactoring the code. I found it because it from a segfault when testing another filter.

src/apps/filters/geoToH3.c
```diff
@@ -34,40 +34,80 @@
...
-    if (!sscanf(argv[1], "%d", &res)) error("parsing resolution");
+    Arg helpArg = {.names = {"-h", "--help"},
+                   .helpText = "Show this help message."};
+    Arg resArg = {.names = {"-r", "--resolution"},
+                  .required = true,
+                  .scanFormat = "%d",
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText = "Resolution, 0-15 inclusive."};
+    Arg latArg = {.names = {"-lat", "--latitude"},
```
nrabinowitz: Nit: I'd have thought `--lat` and `--lon` would be the short forms - I've generally seen single-dash only for single-char args (partly so that if we wanted we could stack like `-xvf`).

One one-char option would be to use `-y` for lat and `-x` for long.

isaacbrodsky: Do you have a preference between adding `--lat`/`--lon` or `-y`/`-x`? I am leaning towards `--lat`/`--lon` since it does not have the ambiguity about which comes first.

src/apps/filters/h3ToGeo.c
```diff
@@ -72,46 +66,68 @@ void doCell(H3Index h, int isKmlOut) {
...
+                  "in degrees")) {
+        return helpArg.found ? 0 : 1;
+    }
 
-        if (argc > 2) {
-            kmlName = argv[2];
-            if (argc > 3) kmlDesc = argv[3];
-        }
+    if (kmlArg.found) {
+        char *kmlName = "geo from H3";
```
nrabinowitz: Nit: Might want a better default here, many folks might not provide a name or desc. Maybe
- Name: "H3 Geometry"
- Desc: "Generated by h3ToGeo"

src/apps/filters/h3ToGeo.c
```diff
@@ -72,46 +66,68 @@ void doCell(H3Index h, int isKmlOut) {
...
+        .value = &index,
+        .helpText =
+            "Index, or not specified to read indexes from standard in."};
+    Arg kmlArg = {.names = {"-k", "--kml"},
+                  .helpText = "Print output in KML format."};
+    Arg kmlNameArg = {.names = {"-kn", "--kml-name"},
+                      .scanFormat = "%255c",  // BUFF_SIZE - 1
+                      .valueName = "name",
+                      .value = &userKmlName,
+                      .helpText = "Name of the KML file."};
```
nrabinowitz: This isn't correct, right? Isn't this name used inside a tag in KML, rather than the file name?

src/apps/filters/h3ToGeo.c
```diff
@@ -17,43 +17,37 @@
...
+ *  respectively.
  *
  *  Examples:
  *
  *     `h3ToGeo < indexes.txt`
  *        - outputs plain text cell center points for the H3 indexes contained
  *          in the file `indexes.txt`
  *
- *     `h3ToGeo 1 "kml file" "h3 cells" < indexes.txt > cells.kml`
+ *     `h3ToGeo --kml --kml-name "kml file" --kml-description "h3 cells" <
```
nrabinowitz: As below, should this read `"kml file"` or `"My KML Title"`

src/apps/filters/h3ToGeo.c
```diff
@@ -17,43 +17,37 @@
...
- *       plain text output (the default) and 1 for KML output
- *
- *  `kmlName` indicates the string for the name tag in KML output (only used
- *       when `outputMode` == 1). The default is "geo from H3".
- *
- *  `kmlName` indicates the string for the desc tag in KML output (only used
- *       when `outputMode` == 1). The default is "generated by h3ToGeo".
+ *  `--kml` causes KML output to be printed. `--kml-name` and
+ *  `--kml-description` can be used to change the name and description in the
+ *  KML header, which default to "geo from H3" and "generated by h3ToGeo"
```
nrabinowitz: You'd need to update here if you take my suggestion below.

src/apps/filters/h3ToGeoBoundary.c
```diff
@@ -17,42 +17,36 @@
...
- *       plain text output (the default) and 1 for KML output
- *
- *  `kmlName` indicates the string for the name tag in KML output (only used
- *       when `outputMode` == 1). The default is "geo from H3".
- *
- *  `kmlName` indicates the string for the desc tag in KML output (only used
- *       when `outputMode` == 1). The default is "generated by h3ToGeoBoundary".
+ *  `--kml` causes KML output to be printed. `--kml-name` and
+ *  `--kml-description` can be used to change the name and description in the
+ *  KML header, which default to "geo from H3" and "generated by
```
nrabinowitz: Ditto

src/apps/filters/h3ToGeoBoundary.c
```diff
@@ -17,42 +17,36 @@
...
+ *  h3ToGeoBoundary" respectively.
  *
  *  Examples:
  *
  *     `h3ToGeoBoundary < indexes.txt`
  *        - outputs plain text cell boundaries for the H3 indexes contained
  *          in the file `indexes.txt`
  *
- *     `h3ToGeoBoundary 1 "kml file" "h3 cells" < indexes.txt > cells.kml`
+ *     `h3ToGeoBoundary --kml --kml-name "kml file" --kml-description "h3 cells"
```
nrabinowitz: Ditto on `"kml file"`

src/apps/filters/h3ToGeoBoundary.c
```diff
@@ -70,46 +64,68 @@ void doCell(H3Index h, int isKmlOut) {
...
+        .value = &index,
+        .helpText =
+            "Index, or not specified to read indexes from standard in."};
+    Arg kmlArg = {.names = {"-k", "--kml"},
+                  .helpText = "Print output in KML format."};
+    Arg kmlNameArg = {.names = {"-kn", "--kml-name"},
+                      .scanFormat = "%255c",  // BUFF_SIZE - 1
+                      .valueName = "name",
+                      .value = &userKmlName,
+                      .helpText = "Name of the KML file."};
```
nrabinowitz: Ditto

src/apps/filters/h3ToGeoBoundary.c
```diff
@@ -70,46 +64,68 @@ void doCell(H3Index h, int isKmlOut) {
...
 
-        if (argc > 2) {
-            kmlName = argv[2];
-            if (argc > 3) kmlDesc = argv[3];
-        }
+    if (kmlArg.found) {
+        char *kmlName = "geo from H3";
+        if (kmlNameArg.found) kmlName = userKmlName;
+
+        char *kmlDesc = "from h3ToGeoBoundary";
```
nrabinowitz: Same notes as above: `"H3 Geometry"` and `"Generated by h3ToGeoBoundary"` would be better IMO

src/apps/filters/kRing.c
```diff
@@ -59,23 +59,23 @@ int main(int argc, char* argv[]) {
...
     Arg helpArg = {.names = {"-h", "--help"},
-                   .helpText = "Show this help message"};
+                   .helpText = "Show this help message."};
     Arg kArg = {.names = {"-k", NULL},
                 .required = true,
                 .scanFormat = "%d",
                 .valueName = "k",
                 .value = &k,
-                .helpText = "Radius of hexagons"};
+                .helpText = "Radius of hexagons."};
```
nrabinowitz: Nit: Might be clearer as "Radius of ring in hexagons"

src/apps/filters/kRing.c
```diff
@@ -59,23 +59,23 @@ int main(int argc, char* argv[]) {
...
     Arg printDistancesArg = {
         .names = {"-d", "--print-distances"},
-        .helpText = "Print distance from origin after index"};
+        .helpText = "Print distance from origin after index."};
```
nrabinowitz: Nit: `...after each index.`?

src/apps/filters/h3ToGeo.c
isaacbrodsky: I had some difficulty with this because the tag is named "name" rather than "title", which I felt would be clearly not the name of the file. I think rewriting it to mention the tag and not refer to a "file" is a good idea.

src/apps/filters/kRing.c
isaacbrodsky: I'm not sure I understand that since this would produce the neighborhood around the index (cf #42) rather than a hollow ring.

src/apps/filters/geoToH3.c
sahrk: To me -y / -x sounds like either planar coordinates or two components of 3D ECEF coordinates. Both might be potential arguments to some future program (especially given that H3 can now output planar IJK grids).

src/apps/filters/h3ToGeo.c
```diff
@@ -72,46 +66,68 @@ void doCell(H3Index h, int isKmlOut) {
...
+    Arg indexArg = {
+        .names = {"-i", "--index"},
+        .scanFormat = "%" PRIx64,
+        .valueName = "index",
+        .value = &index,
+        .helpText =
+            "Index, or not specified to read indexes from standard in."};
+    Arg kmlArg = {.names = {"-k", "--kml"},
+                  .helpText = "Print output in KML format."};
+    Arg kmlNameArg = {.names = {"--kn", "--kml-name"},
```
nrabinowitz: Here I might have gone for `-n` and `-d`, but that's just bikeshedding.

src/apps/filters/kRing.c
nrabinowitz: Per offline: `Radius in hexagons.`

PR #231, opened 2019-05-06
Modified the Travis configuration file to enable automated FOSSA scans.

xinding33:
Hello, I'm from FOSSA and I'm working with the Uber OSPO to automate license scanning. In addition to the changes in this PR, we may also need to add an API key as an environmental variable in the build environment.

I noticed that this project is predominantly C. How are you vendoring third-party code for this project? Is all the third-party code placed in a subdirectory?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,7 @@ before_script:
   - mkdir build
   - cd build
   - cmake -DWARNINGS_AS_ERRORS=ON ..
+  - "curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | sudo bash"
 
 # Build and test the library, then build and test the examples.
 script:
@@ -79,3 +80,4 @@ script:
   - cmake ../../examples
   - make
   - make test
+  - fossa


dfellis:
There is currently 0 third party code in this project.

.travis.yml
```diff
@@ -68,6 +68,7 @@ before_script:
   - mkdir build
   - cd build
   - cmake -DWARNINGS_AS_ERRORS=ON ..
+  - "curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | sudo bash"
```
dfellis: I'm not a maintainer on this, anymore, so I'll leave the decision on this to Isaac, but `curl | sudo bash` is a huge code smell. Can you do this another way?

xinding33: Yes, absolutely. We can download the binary directly and integrate it into the build environment. We can essentially mimic what the install.sh script does: https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh.

dfellis: I would get @isaacbrodsky 's opinion on this, but personally I'd prefer the committed/vetted binary. `curl | sudo bash` might be safe to run within the access controls Travis-CI has been given, but I could see someone running it manually on some code they're still working on, and then if your team is hacked (no offense meant, every org could be) a malicious version of the executable could detect the elevated access and could patch in something malicious to the codebase.

That might be a bit paranoid, but I think it's worth setting a high bar for open source projects lots of people use and are interested in.

I'd also be okay with a `curl ... > install.sh && sha256sum -c checksumfile && sudo ./install.sh` though that would cause errors when you update the installer.

dfellis: (Also, someone could accidentally misconfigure the Travis-CI access and the vulnerability suddenly affects all Uber OSS projects, though that would affect any pull request, not requiring the extra hop through FOSSA's security. I only mention this because I don't think I realized just how critical the security of the CI system is until this PR.)

isaacbrodsky: > Yes, absolutely. We can download the binary directly and integrate it into the build environment. We can essentially mimic what the install.sh script does: https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh.

This means adding the `fossa` binary for Linux amd64 to this repo?

> ... a malicious version of the executable could detect the elevated access ...

I'm not sure I entirely understand your concern. Are you specifically concerned about the `sudo bash`? And/or about retrieving the artifact from `master` on `fossas/fossa-cli`?

Is this concern mitigated if we are not building distributed releases from within Travis CI?

> I'd also be okay with a `curl ... > install.sh && sha256sum -c checksumfile && sudo ./install.sh` though that would cause errors when you update the installer.

I like the idea to sha256 the installer but I'm not clear it would work as you expect, since the install.sh script is verifying a checksum retrieved from fossas/fossa-cli. This seems to move the part that could be modified from install.sh to a tag.

isaacbrodsky:
I think this is effected by the same issue as uber/h3-java#45, you may need to take some action to trigger a Travis CI build (maybe add a commit to the branch?)

.travis.yml
dfellis: > > Yes, absolutely. We can download the binary directly and integrate it into the build environment. We can essentially mimic what the install.sh script does: https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh.
> 
> This means adding the `fossa` binary for Linux amd64 to this repo?

That solution would be that, yes. It could be vetted before committing to make sure nothing malicious is in there.

> 
> > ... a malicious version of the executable could detect the elevated access ...
> 
> I'm not sure I entirely understand your concern. Are you specifically concerned about the `sudo bash`? And/or about retrieving the artifact from `master` on `fossas/fossa-cli`?

The `sudo` part has a potential impact on Travis-CI's security. The `curl | <execute>` part could impact the project's security. There's no vetting/validation that the program does what it's intended to do. If the FOSSA team's projects are ever compromised, we have no idea what the new executable could be doing.

> 
> Is this concern mitigated if we are not building distributed releases from within Travis CI?

Partially. If the build artifacts are created outside of Travis-CI and this fossa command is not installed and used in that other environment, it shouldn't be able to affect it. But I'm also thinking of a simple human error type situation where it's run with elevated privileges on a developer's personal laptop, for instance.

> 
> > I'd also be okay with a `curl ... > install.sh && sha256sum -c checksumfile && sudo ./install.sh` though that would cause errors when you update the installer.
> 
> I like the idea to sha256 the installer but I'm not clear it would work as you expect, since the install.sh script is verifying a checksum retrieved from fossas/fossa-cli. This seems to move the part that could be modified from install.sh to a tag.

Oh, I meant in this download-but-checksum situation the `checksumfile` is committed to the repo but the code is downloaded and installed as they normally do it. The checksum file would just look something like this:

```
7ce1848302585f53b60e8dd64342c42c0c460fbee4317d4e6495dcec0101bd0c  install.sh
```

The downside is that then you'll get FOSSA check failures if/when they update the installer, but the upside is that any change made can be manually vetted before pushing forward. This would be similar to the commit-fossa-binary proposal but with different failure modes (each change to the checker causes a visible failure, while with the binary commit one the checker may run out-of-date code for a while before FOSSA's backend service detects this and rejects it).

On a side note: Why does FOSSA even need to run an executable in the CI process to check for license issues? Why not a CI-like Github webhook on code changes that simply downloads the relevant code and performs the check on their own servers?

isaacbrodsky:
Replacing this PR with #291.

PR #234, opened 2019-05-15
Consistent pentagon local IJ coords

### LINKED ISSUE #184: h3Line algorithm has error cases crossing pentagons
This is probably a downstream effect of problems in `h3ToLocalIjk`. Certain `start, end` pairs that cross a pentagon will succeed, indicating that `h3Distance` and therefore `h3ToLocalIjk` succeeded, but the line cannot be drawn contiguously.

Error case: `H3_EXPORT(h3Line)(0x85285aa7fffffff, 0x851d9b1bfffffff, line)`

Output:

screen shot 2019-01-25 at 5 11 50 pm

isaacbrodsky:
Change failed directions so that more than one face cannot be crossed
when unfolding a pentagon. This should fix #184 by disallowing this kind of line.


src/h3lib/lib/localij.c
```diff
@@ -87,24 +87,28 @@ const int PENTAGON_ROTATIONS_REVERSE_POLAR[7][7] = {
...
+ * Prohibited directions when unfolding a pentagon.
+ *
+ * Indexes by two directions, both relative to the pentagon base cell. The first
+ * is the direction of the origin index and the second is the direction of the
+ * index to unfold. Direction refers to the direction from base cell to base
+ * cell if the indexes are on different base cells, or the leading digit if
+ * within the pentagon base cell.
+ *
+ * This previously included a Class II/Class III check but these were removed
+ * due failure cases. It's possible this could be restricted to a narrower set
```
dfellis: Nit: due _to_ failure cases.

src/h3lib/lib/localij.c
```diff
@@ -183,13 +187,7 @@ int h3ToLocalIjk(H3Index origin, H3Index h3, CoordIJK* out) {
...
-            // TODO: This previously included the Class III-based checks
-            // as in the index-on-pentagon case below, but these were
-            // removed due to some failure cases. It is possible that we
-            // could restrict this error to a narrower set of cases.
-            // https://github.com/uber/h3/issues/163
-            if (FAILED_DIRECTIONS_III[originLeadingDigit][dir] ||
-                FAILED_DIRECTIONS_II[originLeadingDigit][dir]) {
+            if (FAILED_DIRECTIONS[originLeadingDigit][dir]) {
                 // TODO this part of the pentagon might not be unfolded
                 // correctly.
```
nrabinowitz: Nit: Not a big deal, but reading this comment makes me think something is incorrect, rather than something *might* be incorrect and so we're throwing an error. Maybe `TODO: this part of the pentagon might not be unfolded correctly, so throw an error for now`?

src/h3lib/lib/localij.c
```diff
@@ -188,8 +188,8 @@ int h3ToLocalIjk(H3Index origin, H3Index h3, CoordIJK* out) {
             int originLeadingDigit = _h3LeadingNonZeroDigit(origin);
 
             if (FAILED_DIRECTIONS[originLeadingDigit][dir]) {
-                // TODO this part of the pentagon might not be unfolded
-                // correctly.
+                // TODO In this case this part of the pentagon might not be
```
dfellis: Nit: `In this case this part...` -> `In case this part...`

nrabinowitz: No, I think it's correct as is - "in case" means you don't know whether something might occur, but you want to anticipate it; "in this case" means you are clear that in the current context something will occur, and you are handling it.

dfellis: But we don't know if it's actually unfolding incorrectly or not, so in case that would happen we're doing something. The current phrasing is very confusing.

nrabinowitz: I requested this change, because every time I see this comment I think that `return 3` is thought to be possibly incorrect, instead of being a correct guard against possible issues. Open to other options. Maybe 
```
TODO: We may be unfolding the pentagon incorrectly in this case; return an error
code until this is guaranteed to be correct.
```
?

dfellis: That one is definitely clear.

isaacbrodsky: I'll update with the revised wording @nrabinowitz suggested

PR #238, opened 2019-05-17
Add argument parsing for other filter applications

isaacbrodsky:
Adds argument parsing for other applications, except for test/generator applications. This should bring argument parsing to all the existing filter applications.


src/apps/miscapps/h3ToGeoBoundaryHier.c
```diff
@@ -94,55 +97,90 @@ void recursiveH3IndexToGeo(H3Index h, int res, int isKmlOut) {
...
+                   .helpText = "Show this help message."};
+    Arg resArg = {.names = {"-r", "--resolution"},
+                  .scanFormat = "%d",
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, if less than the resolution of the prefix "
+                      "only the prefix is printed. Default 0."};
+    Arg prefixArg = {
+        .names = {"-p", "--prefix"},
```
dfellis: Why is this called `prefix` instead of `parent`?

isaacbrodsky: Not sure why I did that - I thought perhaps the term `prefix` had been used in these filters before but I see from the diff that isn't the case. I changed it to `parent`.

We might want to refactor these to `h3ToParent` and `h3ToChildren` to match the C API later.

dfellis: Technically all of these app changes could be considered a breaking change requiring a major version bump (if there's anyone out there who has a shell script dependent on them) but I don't think we've considered them part of the API before.

We could make that change if/when we decide to do that (and presumably after we make sure the app collection is capable of doing everything the C API can?)

isaacbrodsky: > Technically all of these app changes could be considered a breaking change requiring a major version bump (if there's anyone out there who has a shell script dependent on them) but I don't think we've considered them part of the API before.

We are not considering the filter applications part of the public API at this time.

> We could make that change if/when we decide to do that (and presumably after we make sure the app collection is capable of doing everything the C API can?)

Yes, I mention it to point out an existing terminology which isn't consistent with the C API.

src/apps/filters/h3ToComponents.c
```diff
@@ -100,33 +83,64 @@ void doCell(H3Index h, int verboseMode) {
...
+            printf("\n");
+        } else if (h3Mode == H3_UNIEDGE_MODE) {
+            printf("%d:%d:%d:%d:", h3Mode, H3_GET_RESERVED_BITS(h), h3Res,
+                   h3BaseCell);
+            for (int i = 1; i <= h3Res; i++) {
+                printf("%c", resDigitToChar(H3_GET_INDEX_DIGIT(h, i)));
+            }
+            printf("\n");
+        } else {
+            printf("INVALID INDEX\n");
```
nrabinowitz: Doesn't look like we check `h3IsValid` here, just the mode - should we check if the input is valid here? Not sure if that's necessary for all the filters, but almost certainly necessary for this one (plus this filter doesn't have perf concerns).

src/apps/filters/h3ToComponents.c
```diff
@@ -100,33 +83,64 @@ void doCell(H3Index h, int verboseMode) {
...
+                   .helpText = "Show this help message."};
+    Arg verboseArg = {.names = {"-v", "--verbose"},
+                      .helpText = "Verbose output mode."};
+    Arg indexArg = {
+        .names = {"-i", "--index"},
+        .scanFormat = "%" PRIx64,
+        .valueName = "index",
+        .value = &index,
+        .helpText =
+            "Index, or not specified to read indexes from standard in."};
```
nrabinowitz: Nit: `stdin` is actually clearer IMO. I think the long form is "standard input".

src/apps/filters/h3ToLocalIj.c
```diff
@@ -47,32 +47,56 @@ void doCell(H3Index h, H3Index origin) {
...
-    char buff[BUFF_SIZE];
-    while (1) {
-        // get an index from stdin
-        if (!fgets(buff, BUFF_SIZE, stdin)) {
-            if (feof(stdin))
-                break;
-            else
-                error("reading H3 index from stdin");
-        }
+    if (!H3_EXPORT(h3IsValid)(origin)) {
```
nrabinowitz: Same note - we should probably fail on invalid `index` or `stdin` input as well.

src/apps/filters/h3ToLocalIj.c
```diff
@@ -47,32 +47,56 @@ void doCell(H3Index h, H3Index origin) {
...
-    if (argc != 2) {
-        fprintf(stderr, "usage: %s [origin]\n", argv[0]);
-        exit(1);
-    }
+int main(int argc, char* argv[]) {
+    H3Index origin = 0;
+    H3Index index = 0;
 
-    H3Index origin;
+    Arg helpArg = {.names = {"-h", "--help"},
```
nrabinowitz: This is a total nit, but declaring each arg as a variable like this leads to inconsistent indentation, which makes the arg list far less legible.

I think this might read better as:
```
Arg* args = [
  {.names = {"-h", "--help"},
  ...
];
// Then name the args
Arg helpArg = args[0]; 
```
This avoids naming args you don't need to reference as vars, and keeps the indentation consistent.

src/apps/filters/h3ToLocalIj.c
```diff
@@ -47,32 +47,56 @@ void doCell(H3Index h, H3Index origin) {
...
+        .valueName = "origin",
+        .value = &origin,
+        .required = true,
+        .helpText =
+            "Origin (anchoring index) for the local coordinate system."};
+    Arg indexArg = {.names = {"-i", "--index"},
+                    .scanFormat = "%" PRIx64,
+                    .valueName = "index",
+                    .value = &index,
+                    .helpText = "Index to convert to IJ coordinates."};
```
nrabinowitz: Add same note about defaulting to `stdin` otherwise?

src/apps/filters/h3ToLocalIj.c
```diff
@@ -47,32 +47,56 @@ void doCell(H3Index h, H3Index origin) {
...
 
-    if (!sscanf(argv[1], "%" PRIx64, &origin))
-        error("origin could not be read");
+    Arg* args[] = {&helpArg, &originArg, &indexArg};
+    const int numArgs = 3;
+    const char* helpText = "Converts H3 indexes to local IJ coordinates";
```
nrabinowitz: It's worth noting that the real documentation here is mostly in the `@brief` comment at the top of the file, where it will be inaccessible to the end user. Should we put more of that content into the `helpText` or into some other place where a user could read it without access to the code? In this case especially, there's a lot of useful info in the header comment.

src/apps/filters/hexRange.c
```diff
@@ -50,29 +50,48 @@ void doCell(H3Index h, int k) {
...
+                .valueName = "k",
+                .value = &k,
+                .helpText = "Radius of hexagons."};
+    Arg originArg = {
+        .names = {"-o", "--origin"},
+        .scanFormat = "%" PRIx64,
+        .valueName = "origin",
+        .value = &origin,
+        .helpText =
+            "Origin, or not specified to read origins from standard in."};
```
nrabinowitz: Ditto: `stdin`

src/apps/filters/hexRange.c
```diff
@@ -50,29 +50,48 @@ void doCell(H3Index h, int k) {
...
+    H3Index origin = 0;
+
+    Arg helpArg = {.names = {"-h", "--help"},
+                   .helpText = "Show this help message."};
+    Arg kArg = {.names = {"-k", NULL},
+                .required = true,
+                .scanFormat = "%d",
+                .valueName = "k",
+                .value = &k,
+                .helpText = "Radius of hexagons."};
```
nrabinowitz: Nit: As in `kRing`, suggest `Radius in hexagons`

src/apps/filters/hexRange.c
```diff
@@ -50,29 +50,48 @@ void doCell(H3Index h, int k) {
...
-    char buff[BUFF_SIZE];
-    while (1) {
-        // get an index from stdin
-        if (!fgets(buff, BUFF_SIZE, stdin)) {
-            if (feof(stdin))
-                break;
-            else
-                error("reading H3 index from stdin");
-        }
+    if (originArg.found) {
```
nrabinowitz: Again, we should probably validate input.

src/apps/filters/localIjToH3.c
```diff
@@ -46,30 +46,71 @@ void doCell(const CoordIJ *ij, H3Index origin) {
...
+            // get coordinates from stdin
+            if (!fgets(buff, BUFF_SIZE, stdin)) {
+                if (feof(stdin))
+                    break;
+                else
+                    error("reading IJ coordinates from stdin");
+            }
+
+            if (!sscanf(buff, "%d %d", &ij.i, &ij.j))
+                error("Parsing IJ coordinates. Expected `i j`.");
```
nrabinowitz: Might want to specify that they must both be integers, e.g. `Expected '<int> <int>'`

src/apps/miscapps/h3ToGeoBoundaryHier.c
```diff
@@ -27,21 +28,23 @@
  *  `resolution` should be a positive integer. The default is 0 (i.e., only the
  *       specified cell H3Index would be processed).
  *
- *  `outputMode` indicates the type of output; currently the choices are 0 for
- *       plain text output (the default) and 1 for KML output.
+ *  `--kml` indicates KML output format; if not specified plain text output is
```
nrabinowitz: Again, this contains more detail than the actual help comment. We should think about how to expose this as documentation (maybe MAN pages? I have no idea how that works).

src/apps/miscapps/h3ToGeoBoundaryHier.c
```diff
@@ -17,7 +17,8 @@
  * @brief takes an H3 index and generates cell boundaries for all descendants
  * at a specified resolution.
  *
- *  usage: `h3ToGeoBoundaryHier H3Index [resolution outputMode]`
+ *  usage: `h3ToGeoBoundaryHier --parent parent [--resolution res] [--kml
```
nrabinowitz: Looks like we're somewhat inconsistent about these `usage` comments. I think the best option would be something like
```
See `h3ToGeoBoundaryHier --help` for usage.
```
to avoid getting out of sync with the argument defs.

src/apps/miscapps/h3ToGeoBoundaryHier.c
```diff
@@ -94,55 +97,90 @@ void recursiveH3IndexToGeo(H3Index h, int res, int isKmlOut) {
...
+        .helpText = "Print cell boundaries descendent from this index."};
+    Arg kmlArg = {.names = {"-k", "--kml"},
+                  .helpText = "Print output in KML format."};
+    Arg kmlNameArg = {.names = {"--kn", "--kml-name"},
+                      .scanFormat = "%255c",  // BUFF_SIZE - 1
+                      .valueName = "name",
+                      .value = &userKmlName,
+                      .helpText = "Text for the KML name tag."};
+    Arg kmlDescArg = {.names = {"--kd", "--kml-description"},
+                      .scanFormat = "%255c",  // BUFF_SIZE - 1
```
nrabinowitz: Worth assigning this to a constant or macro? Always needing to include the same explanatory comment is a bit of a code smell.

src/apps/miscapps/h3ToGeoBoundaryHier.c
```diff
@@ -94,55 +97,90 @@ void recursiveH3IndexToGeo(H3Index h, int res, int isKmlOut) {
...
+                      .valueName = "description",
+                      .value = &userKmlDesc,
+                      .helpText = "Text for the KML description tag."};
```
nrabinowitz: Worth noting `Assumes --kml` or similar in these arg descriptions?

src/apps/miscapps/h3ToGeoHier.c
```diff
@@ -95,55 +97,92 @@ void recursiveH3IndexToGeo(H3Index h, int res, int isKmlOut) {
...
+
+    Arg helpArg = {.names = {"-h", "--help"},
+                   .helpText = "Show this help message."};
+    Arg resArg = {.names = {"-r", "--resolution"},
+                  .scanFormat = "%d",
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, if less than the resolution of the parent "
+                      "only the parent is printed. Default 0."};
```
nrabinowitz: Nit: `Default 0` seems incorrect; the default is the res of the parent.

src/apps/miscapps/h3ToGeoHier.c
```diff
@@ -95,55 +97,92 @@ void recursiveH3IndexToGeo(H3Index h, int res, int isKmlOut) {
...
+    Arg parentArg = {
+        .names = {"-p", "--parent"},
+        .scanFormat = "%" PRIx64,
+        .valueName = "parent",
+        .value = &parentIndex,
+        .required = true,
+        .helpText = "Print cell centers descendent from this index."};
+    Arg kmlArg = {.names = {"-k", "--kml"},
```
nrabinowitz: Not required, but would it be worthwhile having common args defined in a util file and reused here? The KML args, the help arg, etc are pretty standard, and might be annoying to maintain across multiple files.

src/apps/miscapps/h3ToGeoHier.c
```diff
@@ -95,55 +97,92 @@ void recursiveH3IndexToGeo(H3Index h, int res, int isKmlOut) {
...
+    }
 
-        if (res > MAX_H3_RES)
-            error("specified resolution exceeds max resolution");
+    int rootRes = H3_GET_RESOLUTION(parentIndex);
 
-        if (argc > 3) {
-            if (!sscanf(argv[3], "%d", &isKmlOut))
-                error("outputMode must be an integer");
+    if (kmlArg.found) {
```
nrabinowitz: This block also seems like it could be defined in one place and reused.

src/apps/miscapps/h3ToGeoHier.c
isaacbrodsky: This program accepts coarser resolutions as valid input, so this is documenting that existing behavior.  From that perspective, 0 does seem to really be the default. I changed it to be a little clearer.

src/apps/filters/h3ToComponents.c
isaacbrodsky: This is a good point that all filters should handle invalid input, but I am going to defer this to a later change.

src/apps/filters/h3ToLocalIj.c
isaacbrodsky: I am not that convinced, I think that indexing into the args array could lead to errors (off by one, miscounting the number of arguments, etc).

src/apps/filters/h3ToComponents.c
```diff
@@ -100,33 +81,63 @@ void doCell(H3Index h, int verboseMode) {
...
+        .names = {"-i", "--index"},
+        .scanFormat = "%" PRIx64,
+        .valueName = "index",
+        .value = &index,
+        .helpText =
+            "Index, or not specified to read indexes from standard input."};
+
+    Arg* args[] = {&helpArg, &verboseArg, &indexArg};
+
+    if (parseArgs(argc, argv, 3, args, &helpArg,
```
nrabinowitz: Nit: Declare a var for `3`, for consistency?

src/apps/filters/h3ToGeo.c
```diff
@@ -67,30 +64,18 @@ void doCell(H3Index h, int isKmlOut) {
 
 int main(int argc, char *argv[]) {
     H3Index index = 0;
-    char userKmlName[BUFF_SIZE] = {0};
-    char userKmlDesc[BUFF_SIZE] = {0};
 
-    Arg helpArg = {.names = {"-h", "--help"},
-                   .helpText = "Show this help message."};
+    Arg helpArg = ARG_HELP;
     Arg indexArg = {
```
nrabinowitz: Not required, but `indexArg` seems like another candidate for a common macro.

src/apps/filters/hexRange.c
```diff
@@ -50,29 +51,47 @@ void doCell(H3Index h, int k) {
...
+        .names = {"-o", "--origin"},
+        .scanFormat = "%" PRIx64,
+        .valueName = "origin",
+        .value = &origin,
+        .helpText =
+            "Origin, or not specified to read origins from standard input."};
+
+    Arg* args[] = {&helpArg, &kArg, &originArg};
+
+    if (parseArgs(argc, argv, 3, args, &helpArg,
```
nrabinowitz: Nit: `numArgs`?

PR #246, opened 2019-06-07
Add release script to prompt for changelog entries

isaacbrodsky:
Change the release process to use this script, which prompts the user to review CHANGELOG.md and any merge commits, that is any PRs merged, before writing the new version number.

### DIFF
--- a/scripts/update_version.sh
+++ b/scripts/update_version.sh
@@ -16,7 +16,7 @@
 # This script is used interactively as part of the H3 release process 
 # (RELEASE.md) to update the version number in the VERSION file. Before
 # writing the new version to the file, changelog information is presented
-# for verification.
+# for verification. It is invoked by the make target `update-version`.
 
 set -eo pipefail
 
@@ -45,5 +45,6 @@ if [ "y" = "$CHANGELOG_OK" ] || [ "Y" = "$CHANGELOG_OK" ]; then
     echo "Wrote $NEXT_VERSION to the VERSION file"
 else
     echo "Cancelled - did not write VERSION file"
+    echo "Please update the CHANGELOG with the appropriate entries before bumping the version."
     exit 2
 fi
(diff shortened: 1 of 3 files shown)


scripts/update_version.sh
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+
+# This script is used interactively as part of the H3 release process 
+# (RELEASE.md) to update the version number in the VERSION file. Before
+# writing the new version to the file, changelog information is presented
+# for verification.
+
+set -eo pipefail
+
+CURRENT_VERSION=$(<VERSION)
```
nrabinowitz: For my info, this requires that we run the script from the root, right? It won't work if we run from inside `scripts`?

That might suggest it would be worthwhile to have a Make target for this.

scripts/update_version.sh
```diff
@@ -0,0 +1,49 @@
...
+echo -e "\n * Committed merges *"
+git log --merges --oneline $REVISION_RANGE
+
+echo
+read -p "Are all changes in the changelog [y/N]? " CHANGELOG_OK
+if [ "y" = "$CHANGELOG_OK" ] || [ "Y" = "$CHANGELOG_OK" ]; then
+    echo $NEXT_VERSION > VERSION
+    echo "Wrote $NEXT_VERSION to the VERSION file"
+else
+    echo "Cancelled - did not write VERSION file"
```
nrabinowitz: Nit: Might add `Please update the CHANGELOG with the appropriate entries before bumping the version.`

scripts/update_version.sh
isaacbrodsky: Yes, this expects to be run from the root of the repo, where the `VERSION` file is located.

Normally I'd run `make` from outside the repo directory, so I could add a parameter for the source directory that should be updated.

isaacbrodsky: Updated to include this.

PR #247, opened 2019-06-10
Add BUILD_BENCHMARKS, BUILD_FILTERS to CMake

isaacbrodsky:
This is intended to help with #243, so that just the core library and its tests can be built, e.g. by running CMake with `cmake -DBUILD_FILTERS=OFF -DBUILD_BENCHMARKS=OFF .`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The public API of this library consists of the functions declared in file
 
 ## [Unreleased]
 ### Added
-- CMake options for excluding filter applications or benchmarks from the build.
+- CMake options for excluding filter applications or benchmarks from the build. (#247)
 
 ## [3.4.4] - 2019-05-30
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The public API of this library consists of the functions declared in file
 [h3api.h.in](./src/h3lib/include/h3api.h.in).
 
 ## [Unreleased]
+### Added
+- CMake options for excluding filter applications or benchmarks from the build.
 
 ## [3.4.4] - 2019-05-30
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The public API of this library consists of the functions declared in file
 [h3api.h.in](./src/h3lib/include/h3api.h.in).
 
 ## [Unreleased]
+### Added
+- CMake options for excluding filter applications or benchmarks from the build.
```
nrabinowitz: Nit: Add PR # ?

As an aside, if we required PR numbers for all lines in the Changelog, it might make scripting the release check more reliable.

CMakeLists.txt
```diff
@@ -551,9 +557,14 @@ configure_package_config_file(
 #   * header location after install: <prefix>/include/h3/h3api.h
 #   * headers can be included by C++ code `#include <h3/h3api.h>`
 # Installing the library and filters system-wide.
+set(INSTALL_TARGETS h3)
+if(BUILD_FILTERS)
+   list(APPEND INSTALL_TARGETS geoToH3 h3ToComponents h3ToGeo h3ToGeoBoundary hexRange
```
nrabinowitz: Is that the full list? I don't see `h3ToLocalIj` or `localIjToH3`.

It also seems a little odd here to bundle the filters (which eventually should have the standing of a new binding) with the apps under `miscapps`. I think we should consider taking the "helper" apps under `miscapps` and treating them differently, e.g. with a different flag here. It also seems like `h3ToComponents` might properly live in `miscapps` not `filters` if we expect `filters` to eventually cover the API as a CLI binding.

CHANGELOG.md
isaacbrodsky: The trouble is I would need to know the PR number before creating the PR.

CMakeLists.txt
isaacbrodsky: We should probably refactor this so that when the application is registered it is added to the install targets list then, avoiding the possibility of it getting out of sync.

I agree that eventually the miscapps should be something else, but in the mean time they seem to be useful as part of the H3 binding.

PR #250, opened 2019-06-28
Commit package-lock.json to not have uncommited changes when building

isaacbrodsky:
Otherwise, the build-to-gh-pages.sh script fails because of uncommitted changes.


dfellis:
You can alternatively `.gitignore` the `package-lock.json` file to avoid this issue.

Since this is just for the GH Page build script, a hard dependency lock doesn't seem as necessary to me.

PR #253, opened 2019-07-15
Implement h3GetFaces

### LINKED ISSUE #236: Add h3GetFaces
Knowing that two cells are on different faces of the icosahedron can be useful for rendering, because they will have different shape distortions. This has come up in the deck.gl integration ([see e.g. this PR](https://github.com/uber/deck.gl/pull/3103)), and it should be trivial to add an util like `h3GetFace` to return the face number 0-20.

nrabinowitz:
Implements `h3GetFaces` per the discussion in #236, as well as the helper function `maxFaceCount` (using the same pattern as `maxKRingSize` etc). The `h3GetFaces` function takes an H3 index and an output array and fills the output array with all the icosahedron faces the index intersects. It returns 1 or 2 faces for hexagons and 5 faces for pentagons.

Implementing this required breaking up `_faceIjkToGeoBoundary` and `_faceIjkPentToGeoBoundary` in order to reuse the logic for getting the FaceIJK coords of the vertices - I've checked the benchmark locally and don't see any difference in `h3ToGeoBoundary` performance.

Also adds the test helper `iterateBaseCellIndexesAtRes` to iterate over a single specified base cell.

Closes #236

### DIFF
--- a/src/apps/testapps/testH3GetFaces.c
+++ b/src/apps/testapps/testH3GetFaces.c
@@ -97,6 +97,12 @@ SUITE(h3GetFaces) {
         assertPentagonFaces(pentagon);
     }
 
+    TEST(res15Pentagon) {
+        H3Index pentagon;
+        setH3Index(&pentagon, 15, 4, 0);
+        assertPentagonFaces(pentagon);
+    }
+
     TEST(baseCellHexagons) {
         int singleCount = 0;
         int multipleCount = 0;
--- a/src/apps/testapps/testH3GetFaces.c
+++ b/src/apps/testapps/testH3GetFaces.c
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2018 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/** @file
+ * @brief tests H3 distance function.
+ *
+ *  usage: `testH3Distance`
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include "baseCells.h"
+#include "h3Index.h"
+#include "h3api.h"
+#include "test.h"
+#include "utility.h"
+
+static int countFaces(h3, expectedMax) {
+    int sz = H3_EXPORT(maxFaceCount)(h3);
+    t_assert(sz == expectedMax, "got expected max face count");
+    int *faces = calloc(sz, sizeof(int));
+
+    H3_EXPORT(h3GetFaces)(h3, faces);
+
+    int validCount = 0;
+    for (int i = 0; i < sz; i++) {
+        if (faces[i] >= 0 && faces[i] <= 19) validCount++;
+    }
+
+    free(faces);
+    return validCount;
+}
+
+static void assertSingleHexFace(H3Index h3) {
+    int validCount = countFaces(h3, 2);
+    t_assert(validCount == 1, "got a single valid face");
+}
+
+static void assertMultipleHexFaces(H3Index h3) {
+    int validCount = countFaces(h3, 2);
+    t_assert(validCount == 2, "got multiple valid faces for a hexagon");
+}
+
+static void assertPentagonFaces(H3Index h3) {
+    int validCount = countFaces(h3, 5);
+    t_assert(validCount == 5, "got 5 valid faces for a pentagon");
+}
+
+SUITE(h3GetFaces) {
+    TEST(singleFaceHexes) {
+        // base cell 16 is at the center of an icosahedron face,
+        // so all children should have the same face
+        iterateBaseCellIndexesAtRes(2, assertSingleHexFace, 16);
+        iterateBaseCellIndexesAtRes(3, assertSingleHexFace, 16);
+    }
+
+    TEST(hexagonWithEdgeVertices) {
+        // Class II pentagon neighbor - one face, two adjacent vertices on edge
+        H3Index h3 = 0x821c37fffffffff;
+        assertSingleHexFace(h3);
+    }
+
+    TEST(hexagonWithDistortion) {
+        // Class III pentagon neighbor
+        H3Index h3 = 0x831c06fffffffff;
+        assertMultipleHexFaces(h3);
+    }
+
+    TEST(hexagonCrossingFaces) {
+        // Class II hex with two non-adjacent vertices on edge
+        H3Index h3 = 0x821ce7fffffffff;
+        assertMultipleHexFaces(h3);
+    }
+
+    TEST(hexagonCrossingFaces) {
+        // Class II hex with two vertices on edge
+        H3Index h3 = 0x821ce7fffffffff;
+        assertMultipleHexFaces(h3);
+    }
+
+    TEST(classIIIPentagon) {
+        H3Index pentagon;
+        setH3Index(&pentagon, 1, 4, 0);
+        assertPentagonFaces(pentagon);
+    }
+
+    // TEST(classIIPentagon) {
+    //     H3Index pentagon;
+    //     setH3Index(&pentagon, 2, 4, 0);
+    //     assertPentagonFaces(pentagon);
+    // }
+
+    TEST(baseCellHexagons) {
+        int singleCount = 0;
+        int multipleCount = 0;
+        for (int i = 0; i < NUM_BASE_CELLS; i++) {
+            if (!_isBaseCellPentagon(i)) {
+                // Make the base cell index
+                H3Index baseCell = H3_INIT;
+                H3_SET_MODE(baseCell, H3_HEXAGON_MODE);
+                H3_SET_BASE_CELL(baseCell, i);
+                int validCount = countFaces(baseCell, 2);
+                t_assert(validCount > 0, "got at least one face");
+                if (validCount == 1)
+                    singleCount++;
+                else
+                    multipleCount++;
+            }
+        }
+        t_assert(singleCount == 4 * 20,
+                 "got single face for 4 face-aligned hex base cells");
+        t_assert(multipleCount == 1.5 * 20,
+                 "got multiple faces for non-face-aligned hex base cells");
+    }
+
+    // TEST(baseCellPentagons) {
+    //     for (int i = 0; i < NUM_BASE_CELLS; i++) {
+    //         if (_isBaseCellPentagon(i)) {
+    //             // Make the base cell index
+    //             H3Index baseCell = H3_INIT;
+    //             H3_SET_MODE(baseCell, H3_HEXAGON_MODE);
+    //             H3_SET_BASE_CELL(baseCell, i);
+    //             assertPentagonFaces(baseCell);
+    //         }
+    //     }
+    // }
+}
--- a/src/apps/testapps/testH3GetFaces.c
+++ b/src/apps/testapps/testH3GetFaces.c
@@ -27,7 +27,7 @@
 #include "test.h"
 #include "utility.h"
 
-static int countFaces(h3, expectedMax) {
+static int countFaces(H3Index h3, int expectedMax) {
     int sz = H3_EXPORT(maxFaceCount)(h3);
     t_assert(sz == expectedMax, "got expected max face count");
     int *faces = calloc(sz, sizeof(int));
@@ -54,6 +54,7 @@ static void assertMultipleHexFaces(H3Index h3) {
 }
 
 static void assertPentagonFaces(H3Index h3) {
+    t_assert(H3_EXPORT(h3IsPentagon)(h3), "got a pentagon");
     int validCount = countFaces(h3, 5);
     t_assert(validCount == 5, "got 5 valid faces for a pentagon");
 }
@@ -73,14 +74,8 @@ SUITE(h3GetFaces) {
     }
 
     TEST(hexagonWithDistortion) {
-        // Class III pentagon neighbor
+        // Class III pentagon neighbor, distortion across faces
         H3Index h3 = 0x831c06fffffffff;
-        assertMultipleHexFaces(h3);
-    }
-
-    TEST(hexagonCrossingFaces) {
-        // Class II hex with two non-adjacent vertices on edge
-        H3Index h3 = 0x821ce7fffffffff;
         assertMultipleHexFaces(h3);
     }
 
@@ -96,11 +91,11 @@ SUITE(h3GetFaces) {
         assertPentagonFaces(pentagon);
     }
 
-    // TEST(classIIPentagon) {
-    //     H3Index pentagon;
-    //     setH3Index(&pentagon, 2, 4, 0);
-    //     assertPentagonFaces(pentagon);
-    // }
+    TEST(classIIPentagon) {
+        H3Index pentagon;
+        setH3Index(&pentagon, 2, 4, 0);
+        assertPentagonFaces(pentagon);
+    }
 
     TEST(baseCellHexagons) {
         int singleCount = 0;
@@ -120,20 +115,20 @@ SUITE(h3GetFaces) {
             }
         }
         t_assert(singleCount == 4 * 20,
-                 "got single face for 4 face-aligned hex base cells");
+                 "got single face for 4 aligned hex base cells per face");
         t_assert(multipleCount == 1.5 * 20,
-                 "got multiple faces for non-face-aligned hex base cells");
+                 "got multiple faces for non-aligned hex base cells");
     }
 
-    // TEST(baseCellPentagons) {
-    //     for (int i = 0; i < NUM_BASE_CELLS; i++) {
-    //         if (_isBaseCellPentagon(i)) {
-    //             // Make the base cell index
-    //             H3Index baseCell = H3_INIT;
-    //             H3_SET_MODE(baseCell, H3_HEXAGON_MODE);
-    //             H3_SET_BASE_CELL(baseCell, i);
-    //             assertPentagonFaces(baseCell);
-    //         }
-    //     }
-    // }
+    TEST(baseCellPentagons) {
+        for (int i = 0; i < NUM_BASE_CELLS; i++) {
+            if (_isBaseCellPentagon(i)) {
+                // Make the base cell index
+                H3Index baseCell = H3_INIT;
+                H3_SET_MODE(baseCell, H3_HEXAGON_MODE);
+                H3_SET_BASE_CELL(baseCell, i);
+                assertPentagonFaces(baseCell);
+            }
+        }
+    }
 }
--- a/src/apps/testapps/testH3GetFaces.c
+++ b/src/apps/testapps/testH3GetFaces.c
@@ -14,9 +14,7 @@
  * limitations under the License.
  */
 /** @file
- * @brief tests H3 distance function.
- *
- *  usage: `testH3Distance`
+ * @brief tests the h3GetFaces function
  */
 
 #include <stdio.h>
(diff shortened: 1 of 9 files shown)


src/h3lib/include/faceijk.h
```diff
@@ -50,17 +50,16 @@ typedef struct {
 extern const GeoCoord faceCenterGeo[NUM_ICOSA_FACES];
 
 // indexes for faceNeighbors table
-/** Invalid faceNeighbors table direction */
-#define INVALID -1
-/** Center faceNeighbors table direction */
-#define CENTER 0
```
nrabinowitz: Unused

src/h3lib/lib/faceijk.c
```diff
@@ -891,3 +906,21 @@ int _adjustOverageClassII(FaceIJK* fijk, int res, int pentLeading4,
...
+ * place so that the resulting cell address is relative to the correct
+ * icosahedral face.
+ *
+ * @param fijk The FaceIJK address of the cell.
+ * @param res The H3 resolution of the cell.
+ */
+int _adjustPentVertOverage(FaceIJK* fijk, int res) {
+    int pentLeading4 = 0;
+    int overage = _adjustOverageClassII(fijk, res, pentLeading4, 1);
+    while (overage == 2) {
```
nrabinowitz: Note that I cleaned up the logic here slightly to a single `while` loop.

src/h3lib/lib/h3Index.c
```diff
@@ -779,6 +779,86 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+ * @param out Output array. Must be of size maxFaceCount(h3).
+ */
+void H3_EXPORT(h3GetFaces)(H3Index h3, int* out) {
+    int res = H3_GET_RESOLUTION(h3);
+    int isPentagon = H3_EXPORT(h3IsPentagon)(h3);
+
+    // We can't use the vertex-based approach here for class II pentagons,
+    // because all their vertices are on the icosahedron edges. Their
+    // direct child pentagons cross the same faces, so use those instead.
+    if (isPentagon && !isResClassIII(res)) {
```
dfellis: Res 15 is a Class III grid as I recall, but can we add one test for a Res 15 pentagon to confirm that it doesn't accidentally try to get a child below Res 15 (which might break `makeDirectChild` is interesting ways)?

src/h3lib/lib/h3Index.c
```diff
@@ -779,6 +779,86 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+        H3_EXPORT(h3GetFaces)(childPentagon, out);
+        return;
+    }
+
+    // convert to FaceIJK
+    FaceIJK fijk;
+    _h3ToFaceIjk(h3, &fijk);
+
+    // Get all vertices as FaceIJK addresses. For simplicity, always
+    // initialize the array with 6 verts, ignoring the last one for pentagons
```
dfellis: Nit: This array isn't initialized, you've literally told the C compiler to leave the space for 6 structs directly allocated in the library for this function to use for this operation. (Which also makes this method not-parallel-computing-friendly.)

src/h3lib/lib/h3Index.c
```diff
@@ -779,6 +779,86 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+    // We may not use all of the slots in the output array,
+    // so fill with invalid values to indicate unused slots
+    int faceCount = H3_EXPORT(maxFaceCount)(h3);
+    for (int i = 0; i < faceCount; i++) {
+        out[i] = INVALID_FACE;
+    }
+
+    // add each vertex face, using the output array as a hash set
+    for (int i = 0; i < vertexCount; i++) {
+        FaceIJK vert = fijkVerts[i];
```
dfellis: Nit: This will copy the vertex definition and then perform the operations below, which probably isn't what you wanted. This could be `FaceIJK* vert = &fijkVerts[i];` then `_adjustPentVertOverage(vert, res);`, `_adjustOverageClassII(vert, res, 0, 1);`, and `int face = vert->face;`

src/h3lib/lib/h3Index.c
nrabinowitz: I did think of this, but you're right, I'll add a test.

src/h3lib/lib/h3Index.c
nrabinowitz: Sorry, array handling is still the most confusing part for me of working in C. You're saying I should use `malloc` here?

src/h3lib/lib/h3Index.c
nrabinowitz: Makes sense, will update.

src/h3lib/lib/h3Index.c
nrabinowitz: Or would this work with `FaceIJK fijkVerts[NUM_HEX_VERTS] = {0}`?

dfellis: The first one. The memory needs to be allocated either on the heap or on the stack. This syntax tells the compiler to put some dedicated space somewhere in the `.so/.dll` file and the function then references that space for all calls.

If multiple threads are running the same function at the "same time" (either actually at the same time on different cores, or one thread was running and a CPU interrupt happened partway through and the kernel scheduler spun up another thread and it continued in the same function) you could have very-hard-to-track bugs where the function gives you the wrong answer a vanishingly small amount of the time.

nrabinowitz: Got it. I was expecting the array allocation here to work the same as allocating a struct (which doesn't do what you're describing, right?).

isaacbrodsky: It looks like the original line was `FaceIJK fijkVerts[NUM_HEX_VERTS];` This looks to me like a stack allocation - if it were `static FaceIJK fijkVerts[NUM_HEX_VERTS];` then it would be in the `bss` section in the library.

I think `FaceIJK fijkVerts[NUM_HEX_VERTS] = {0};` should be OK.

dfellis: yes but no

When you have a function signature like: `StructName fnName();` your program is going to copy the value of the struct into a new memory location and not be affected after-the-fact, but a variable declaration within the function like `StructName structVar;` or `StructName structVar = { /* struct contents */ };` *is* going to also be a statically-allocated blob in the `.so/.dll` file that could suffer the same interrupt issue across multiple threads.

But if that struct is just internal to the function (not part of the output signature) sometimes the C compiler/optmizer will break it into its constituent fields as registers on the CPU, and CPU interrupts properly store and retrieve the registers for you and you luck out. (Also why basic `int`/`float`/`double`/etc declarations don't cause problems usually, the variables get completely converted into registers or stack allocations/deallocations.) C is unfortunately low-level enough where this matters but "high-level" enough that you can't clearly declare your intentions, here.

Allocating these things on the stack or the heap is the safest way to go for parallel-friendly functions. (This is probably why @isaachier pushed for so much stack allocation stuff, since it's faster than heap and handles this situation, but unfortunately some platforms don't allow stack allocation and some of our functions can allocate way too much memory.)

I'm thinking H3 should have some [valgrind](http://valgrind.org/) testing set up to check for memory leaks and parallelization issues.

dfellis: > It looks like the original line was `FaceIJK fijkVerts[NUM_HEX_VERTS];` This looks to me like a stack allocation - if it were `static FaceIJK fijkVerts[NUM_HEX_VERTS];` then it would be in the `bss` section in the library and the reuse across function calls would be an issue.
> 
> I think `FaceIJK fijkVerts[NUM_HEX_VERTS] = {0};` should be OK.

I thought stack allocation was disabled because it was causing problems for the ARM builds?

nrabinowitz: > I'm thinking H3 should have some valgrind testing set up to check for memory leaks and parallelization issues

+1, this is a good addition to our CI if possible.

isaacbrodsky: > I'm thinking H3 should have some [valgrind](http://valgrind.org/) testing set up to check for memory leaks and parallelization issues.

> +1, this is a good addition to our CI if possible.

We have an option for this in our CMake setup but we don't run it in CI right now. I think that would be a really good addition to automatically run it. 

> I thought stack allocation was disabled because it was causing problems for the ARM builds?

We removed dynamic stack allocations using [alloca](https://linux.die.net/man/3/alloca). We did not make any compiler option changes relating to how local variables are allocated AFAIK. I get the impression we might be thinking different compiler options are being applied here.

> When you have a function signature like: `StructName fnName();` your program is going to copy the value of the struct into a new memory location and not be affected after-the-fact, but a variable declaration within the function like `StructName structVar;` or `StructName structVar = { /* struct contents */ };` _is_ going to also be a statically-allocated blob in the `.so/.dll` file that could suffer the same interrupt issue across multiple threads.

I don't follow this. Could you send me an example program?

src/h3lib/lib/faceijk.c
isaacbrodsky: `_adjustOverageClassII` might be a good candidate for an enum. (Not a required change for this PR)

src/h3lib/lib/h3Index.c
```diff
@@ -779,6 +779,88 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+        // on another face
+        if (isPentagon) {
+            _adjustPentVertOverage(vert, res);
+        } else {
+            _adjustOverageClassII(vert, res, 0, 1);
+        }
+
+        // Save the face to the output array
+        int face = vert->face;
+        int pos = face % faceCount;
```
isaacbrodsky: Is this hashing approach being chosen for a specific reason?

src/h3lib/lib/h3Index.c
dfellis: > > I'm thinking H3 should have some [valgrind](http://valgrind.org/) testing set up to check for memory leaks and parallelization issues.
> 
> > +1, this is a good addition to our CI if possible.
> 
> We have an option for this in our CMake setup but we don't run it in CI right now. I think that would be a really good addition to automatically run it.
> 
> > I thought stack allocation was disabled because it was causing problems for the ARM builds?
> 
> We removed dynamic stack allocations using [alloca](https://linux.die.net/man/3/alloca). We did not make any compiler option changes relating to how local variables are allocated AFAIK. I get the impression we might be thinking different compiler options are being applied here.
> 

I was thinking of this issue: https://github.com/uber/h3/issues/233 so it was iOS-specific, not ARM-specific, but I remembered it involving a mobile platform. I thought we were using some of the compiler options to enforce that. I thought we were using something like `-mstack-increment=NUM` or `-fmax-stack-var-size` with a low value which would push these to become part of the data section and shared.

I was also confused because gcc says this about alloca:
```
       -msoft-stack
           Generate code that does not use ".local" memory directly for stack storage. Instead, a per-warp stack pointer is maintained explicitly. This enables variable-length stack allocation (with variable-length arrays or
           "alloca"), and when global memory is used for underlying storage, makes it possible to access automatic variables from other threads, or with atomic instructions. This code generation variant is used for OpenMP
           offloading, but the option is exposed on its own for the purpose of testing the compiler; to generate code suitable for linking into programs using OpenMP offloading, use option -mgomp.
```
Which implies that disabling variable-length stack allocation causes the stack to be a static block of shared memory.

> > When you have a function signature like: `StructName fnName();` your program is going to copy the value of the struct into a new memory location and not be affected after-the-fact, but a variable declaration within the function like `StructName structVar;` or `StructName structVar = { /* struct contents */ };` _is_ going to also be a statically-allocated blob in the `.so/.dll` file that could suffer the same interrupt issue across multiple threads.
> 
> I don't follow this. Could you send me an example program?

It would only apply if the earlier assumption is true: that the compiler was given a constraint on the amount of stack memory it can allocate for a given function and the struct in question was too large. It would have to switch from a stack allocation to fixed bytes in bss or data similar to how static variables work (and it would work fine for single-threaded applications).

nrabinowitz: Just checking - this conversation is largely academic, right? I switched to `malloc`, so unless we see a potential downside to using `malloc` here I think this should be fine now.

src/h3lib/lib/h3Index.c
nrabinowitz: It was easy? And I had a model for how to do it in the `kRing` implementation. Is there another option you'd prefer?

src/h3lib/lib/faceijk.c
nrabinowitz: Good idea, added.

src/h3lib/lib/h3Index.c
nrabinowitz: Test added.

src/h3lib/lib/h3Index.c
dfellis: I personally don't see a problem with it, but it's probably the same amount of compute effort (or slightly more for edge-intersecting hexagons) to just scan the array for the first `INVALID_FACE` to write the value, or abort the scan if the value is found in the array.

src/h3lib/lib/h3Index.c
dfellis: Mostly academic, yes, but if the compiler can do the allocation in stack space and not cause iOS to barf, that would be faster (and more predictable in speed) than malloc, which will sometimes have to call out to the kernel to get a new page of memory and cause an interrupt.

I would say the malloc is definitely safer with so many compile targets to worry about.

src/h3lib/lib/h3Index.c
nrabinowitz: Simplified the logic here to iterate over the array, removing modulo logic. In practice, the array is never longer than 5 elements, so it shouldn't make a difference.

src/h3lib/lib/faceijk.c
```diff
@@ -886,8 +901,26 @@ int _adjustOverageClassII(FaceIJK* fijk, int res, int pentLeading4,
...
+ * place so that the resulting cell address is relative to the correct
+ * icosahedral face.
+ *
+ * @param fijk The FaceIJK address of the cell.
+ * @param res The H3 resolution of the cell.
+ */
+Overage _adjustPentVertOverage(FaceIJK* fijk, int res) {
+    int pentLeading4 = 0;
+    Overage overage = _adjustOverageClassII(fijk, res, pentLeading4, 1);
+    while (overage == NEW_FACE) {
```
isaacbrodsky: nit: Any preference for or against a do/while loop?
```
Overage overage;
do {
    overage = _adjustOverageClassII(fijk, res, pentLeading4, 1);
} while(overage == NEW_FACE);
```

src/h3lib/lib/h3Index.c
```diff
@@ -739,12 +739,12 @@ void _h3ToFaceIjk(H3Index h, FaceIJK* fijk) {
...
     int pentLeading4 =
         (_isBaseCellPentagon(baseCell) && _h3LeadingNonZeroDigit(h) == 4);
-    if (_adjustOverageClassII(fijk, res, pentLeading4, 0)) {
+    if (_adjustOverageClassII(fijk, res, pentLeading4, 0) != NO_OVERAGE) {
         // if the base cell is a pentagon we have the potential for secondary
         // overages
         if (_isBaseCellPentagon(baseCell)) {
             while (1) {
-                if (!_adjustOverageClassII(fijk, res, 0, 0)) break;
+                if (_adjustOverageClassII(fijk, res, 0, 0) == NO_OVERAGE) break;
```
isaacbrodsky: Nit: seems like a good use for a do/while loop rather than `while(1)` (doesn't need to be changed since that wasn't added in this diff)

src/h3lib/lib/h3Index.c
isaacbrodsky: I don't think makeDirectChild is well defined for resolution 15 indexes. This call is overflowing the resolution bitfield, and creates a hexagon index at resolution 0 (so with invalid indexing digits) with an "edge/reserved" field of 1. I suspect what happens here is that it needs to recurse again to get to resolution 1, but didn't test. I don't think that code flow is intended, so it may be better to express this in a different way.

src/h3lib/lib/h3Index.c
```diff
@@ -779,6 +779,88 @@ void H3_EXPORT(h3ToGeoBoundary)(H3Index h3, GeoBoundary* gb) {
...
+/**
+ * Find all icosahedron faces intersected by a given  H3 index, represented
+ * as integers from 0-19. The array is sparse; since 0 is a valid value,
+ * invalid array values are represented as -1. It is the responsibility of
+ * the caller to filter out invalid values.
+ *
+ * @param h3 The H3 index
+ * @param out Output array. Must be of size maxFaceCount(h3).
+ */
+void H3_EXPORT(h3GetFaces)(H3Index h3, int* out) {
```
isaacbrodsky: Do we want this function to return `int` in order to return an error code on invalid `H3Index`s?

src/h3lib/lib/h3Index.c
isaacbrodsky: I don't believe that 6 FaceIJK structures would be an issue, so I'd suggest using stack allocation here. I don't think we need to be quite that careful about stack space.

nrabinowitz: Is `FaceIJK fijkVerts[NUM_HEX_VERTS] = {0}` using stack allocation, then? Is that the suggested change?

src/apps/testapps/testH3GetFaces.c
```diff
@@ -0,0 +1,140 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/** @file
+ * @brief tests H3 distance function.
+ *
+ *  usage: `testH3Distance`
```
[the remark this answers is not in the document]
nrabinowitz: Good catch, will fix.

src/apps/testapps/testH3GetFaces.c
```diff
@@ -0,0 +1,140 @@
...
+
+    TEST(baseCellHexagons) {
+        int singleCount = 0;
+        int multipleCount = 0;
+        for (int i = 0; i < NUM_BASE_CELLS; i++) {
+            if (!_isBaseCellPentagon(i)) {
+                // Make the base cell index
+                H3Index baseCell = H3_INIT;
+                H3_SET_MODE(baseCell, H3_HEXAGON_MODE);
+                H3_SET_BASE_CELL(baseCell, i);
```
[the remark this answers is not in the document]
nrabinowitz: True, will update.

src/h3lib/lib/faceijk.c
nrabinowitz: Good point, that's a bit cleaner. Will update.

src/h3lib/lib/h3Index.c
nrabinowitz: Sure, can update here as well.

src/h3lib/lib/h3Index.c
nrabinowitz: As noted before, we're really inconsistent on this - most of our similar functions, e.g. `h3ToGeoBoundary`, `h3ToGeo`, `h3GetBaseCell`, etc, do not validate input. I think it's fine to do the same here, for consistency.

My general feeling is that only the caller knows whether input validation is necessary, and they are empowered to add it with `h3IsValid` if checking input validity is more important than performance for their use case.

src/h3lib/lib/h3Index.c
nrabinowitz: This one's a little different - I think the most concise might be
```
while (_adjustOverageClassII(fijk, res, 0, 0) != NO_OVERAGE) continue;
```
I'm going to go with that, unless it seems too clever - using `do/while` requires assigning a variable to hold the overage output, which seems unnecessary.

src/h3lib/lib/h3Index.c
nrabinowitz: Updated to `FaceIJK fijkVerts[NUM_HEX_VERTS] = {0}` per @isaac's suggestion.

nrabinowitz: The previous formulation had compiler warnings, so I'm back to my original `FaceIJK fijkVerts[NUM_HEX_VERTS];` implementation. The internet seems to think this is a stack allocation, and @dfellis's original concern wasn't valid (unless this was declared at the file level and/or `static`). Please let me know if there are additional concerns here.

src/h3lib/lib/h3Index.c
nrabinowitz: > I don't think makeDirectChild is well defined for resolution 15 indexes. This call is overflowing the resolution bitfield

This is inside a `!isResClassIII(res)` condition; res 15 is Class III, so this will never be invoked for res 15 hexes. I can add a comment, just in case the condition is changed later.

nrabinowitz: Comment added.

isaacbrodsky: I might have misread when I mentioned about resolution 15. I agree this doesn't need a change since it won't be invoked for resolution 15.

Issue #252, opened 2019-07-11
Incorrect return code from `hexRing` on input pentagon


ajfriend:
The docstring of the `hexRing()` function says that it should `return 0 if no pentagonal distortion was encountered`. However, due to the "short-circuit" logic at the start of the function, it will always return `0` when `k==0`, no matter if the input `origin` is a pentagon or not:

https://github.com/uber/h3/blob/a6cd31b73aa26f0f7e4f5e5447b2c3bca20b17d6/src/h3lib/lib/algos.c#L549-L554

Do we want this behavior, or would it be better to return `HEX_RANGE_PENTAGON` in this case?

The fix should be easy, and only involve moving this check

https://github.com/uber/h3/blob/a6cd31b73aa26f0f7e4f5e5447b2c3bca20b17d6/src/h3lib/lib/algos.c#L560-L563

to run before the "short-circuit" logic.

dfellis:
Technically no pentagonal distortion is encountered in that case, since it never traverses from the pentagon to a neighbor.

ajfriend:
I was interpreting "pentagonal distortion" to mean "the algorithm ran into a pentagon at any point". Is that incorrect? What's the technical definition of pentagonal distortion?

ajfriend:
And actually, beyond technical correctness, I think most people would default to assuming the same interpretation. I wonder if fixing this would be more intuitive/safer.

dfellis:
> I was interpreting "pentagonal distortion" to mean "the algorithm ran into a pentagon at any point". Is that incorrect? What's the technical definition of pentagonal distortion?

A traversal involving a pentagon occurred, changing the number of valid IJK directions out of the pentagon, breaking the algorithm.

I don't think intentionally breaking the endpoint further is what users would want -- it's a solvable problem to correct the pentagonal distortion just as it is for crossing icosahedron faces, which also was originally broken for hexRing before H3 was open sourced. Basically, these are the limitations of the functions *right now* but we should strive for correctness *and* speed, so making this less correct seems like a step backwards.

(Side note: I believe there are 2 categories of pentagons: the two "polar" pentagons and the 10 "equatorial" pentagons, where the polar pentagons consist of the I direction of each face pointing to the center of the pentagon, while the equatorial pentagons have faces alternating between J and K pointing at the center of the pentagon. Figuring out which kind you're on and which face you're originating from should allow us to adjust the rotations appropriately for hexRing, but I don't really have the time anymore to focus on this problem.)

[...]

PR #256, opened 2019-07-18
changes to hexRing description

ajfriend:
Here's a stab at a change to the `hexRing` function description following #252. Let me know what you think!

### DIFF
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -539,12 +539,16 @@ int H3_EXPORT(hexRanges)(H3Index* h3Set, int length, int k, H3Index* out) {
 }
 
 /**
- * Returns the hollow hexagonal ring centered at origin with sides of length k.
+ * Returns the "hollow" ring of hexagons at exactly grid distance k from
+ * the origin hexagon. In particular, k=0 returns just the origin hexagon.
+ *
+ * A nonzero failure return code may happen, for example, if a pentagon is
+ * encountered. Failure cases may be fixed in future versions.
  *
  * @param origin Origin location.
  * @param k k >= 0
  * @param out Array which must be of size 6 * k (or 1 if k == 0)
- * @return 0 if no pentagonal distortion was encountered.
+ * @return 0 if successful; nonzero otherwise.
  */
 int H3_EXPORT(hexRing)(H3Index origin, int k, H3Index* out) {
     // Short-circuit on 'identity' ring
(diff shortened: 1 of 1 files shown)


src/h3lib/lib/algos.c
```diff
@@ -539,12 +539,16 @@ int H3_EXPORT(hexRanges)(H3Index* h3Set, int length, int k, H3Index* out) {
 }
 
 /**
- * Returns the hollow hexagonal ring centered at origin with sides of length k.
+ * Returns the "hollow" ring of hexagons at exactly grid distance k from
+ * the origin hexagon. In particular, k=0 returns just the origin hexagon.
+ *
+ * A nonzero failure return code may happen, for example, if a pentagon is
+ * encountered. Failure cases may be fixed in future versions.
```
isaacbrodsky: If we change the handling of failure cases, would that require bumping the version to 4.0.0? I believe in that case we may have rings where sides are not the same length. (See around line 611)

src/h3lib/lib/algos.c
```diff
@@ -539,12 +539,16 @@ int H3_EXPORT(hexRanges)(H3Index* h3Set, int length, int k, H3Index* out) {
 }
 
 /**
- * Returns the hollow hexagonal ring centered at origin with sides of length k.
+ * Returns the "hollow" ring of hexagons at exactly grid distance k from
+ * the origin hexagon. In particular, k=0 returns just the origin hexagon.
+ *
+ * A nonzero failure return code may happen, for example, if a pentagon is
```
nrabinowitz: Maybe `A non-zero failure code may be returned in some cases, for example if...`

src/h3lib/lib/algos.c
nrabinowitz: I guess this depends on which guarantees we consider to be a property of the major version. I think as @ajfriend has written it here, as long as all cells in the ring are `k` steps away from the origin, we have met the current API and no version change is needed. Adding additional failure cases is a grey area (it may cause existing code to throw errors, so may need a major bump), but removing failure cases should be fine IMO.

isaacbrodsky: I think the main user-facing change would be the need to handle the 0 values in the array, and possibly if a user made an assumption that `hexRing` always returns a hexagon shape.

ajfriend: Handling 0 values in the array is a good point. Do we want to mention that users currently **do not** need to filter out zeros? Or do we want to say that they should always be wary?

isaacbrodsky: I'd suggest maybe leaving it a little vague for what happens in the cases that currently fail. That would allow us to address this when we have a clearer idea of what the new behavior is.

ajfriend: sounds good. I'll land after making the change suggested by @nrabinowitz

PR #264, opened 2019-07-26
Add a test that runs Valgrind in Travis CI

isaacbrodsky:
Need to first check that Valgrind gets installed & run correctly in CI.

Valgrind should help us find issues with memory management, leaks, etc more readily.


nrabinowitz:
Code change looks fine - looks like the build timed out on the exhaustive tests for local IJ (this often takes ~1min locally, and the timeout is 10 min). It looks like valgrind is a lot slower, so maybe we ought to gate off exhaustive tests (`h3ToLocalIj`, `h3Line`, etc) into a separate suite and avoid running them here.

isaacbrodsky:
> Code change looks fine - looks like the build timed out on the exhaustive tests for local IJ (this often takes ~1min locally, and the timeout is 10 min). It looks like valgrind is a lot slower, so maybe we ought to gate off exhaustive tests (`h3ToLocalIj`, `h3Line`, etc) into a separate suite and avoid running them here.

Another option might be to not allocate memory for every index. What do you think about trying static, preallocated blocks of memory in testH3ToLocalIj around lines 133~134? Or maybe some other way of allocating for the test suite and passing the pointer in?

nrabinowitz:
> Another option might be to not allocate memory for every index. What do you think about trying static, preallocated blocks of memory in testH3ToLocalIj around lines 133~134? Or maybe some other way of allocating for the test suite and passing the pointer in?

I'm not sure that'll work for `h3Line` (size is pretty variable), and we may get the same issue there - my guess is that you'll spend a lot of time hunting down slow spots in the test suite, which doesn't seem worthwhile, and could have a chilling effect on writing exhaustive tests. I think it would simply be easier to take the slower tests and give them their own suite - this could also help running `make test` locally; we could have a `make fast-test` target that skipped the exhaustive suite.

CMakeLists.txt
```diff
@@ -517,6 +520,14 @@ if(BUILD_TESTING)
     add_h3_test_with_arg(testH3NeighborRotations src/apps/testapps/testH3NeighborRotations.c 0)
     add_h3_test_with_arg(testH3NeighborRotations src/apps/testapps/testH3NeighborRotations.c 1)
     add_h3_test_with_arg(testH3NeighborRotations src/apps/testapps/testH3NeighborRotations.c 2)
+
+    # The "Exhaustive" part of the test name is used by the Valgrind CI test to exclude
+    # these files in .travis.yml.
```
nrabinowitz: Nit: This comment is slightly too specific - I'd go with `The "Exhaustive" part of the test name is used to exclude the test from the test-fast target.`

Which makes me wonder whether this shouldn't just be `-Slow`, but admittedly `-Exhaustive` is more descriptive of what's currently there.

isaacbrodsky: I adjusted the comment to say that the "Exhaustive" part is for "test-fast" but I kept a comment that "test-fast" is there so that the Valgrind CI test doesn't timeout.

isaacbrodsky:
I used another branch to demonstrate that the build does fail when Valgrind detects an issue: https://github.com/isaacbrodsky/h3/commits/valgrind-in-travis-intentional-failure https://travis-ci.com/isaacbrodsky/h3/builds/121654600 (The Mac OSX build also failed because of warnings about the specific way I caused Valgrind to fail)

nrabinowitz:
> I used another branch to demonstrate that the build does fail when Valgrind detects an issue: https://github.com/isaacbrodsky/h3/commits/valgrind-in-travis-intentional-failure https://travis-ci.com/isaacbrodsky/h3/builds/121654600 (The Mac OSX build also failed because of warnings about the specific way I caused Valgrind to fail)

Oh, that's great. Though it looks like we don't get any useful output from valgrind here - how should the developer see valgrind output when this test fails?

isaacbrodsky:
> Oh, that's great. Though it looks like we don't get any useful output from valgrind here - how should the developer see valgrind output when this test fails?

This is a really good point. I found the `CTEST_OUTPUT_ON_FAILURE` option online and it looks like this does cause the Valgrind output to be printed: https://travis-ci.com/isaacbrodsky/h3/jobs/221833237 It does also print out of the `.`s, but hopefully for the fast tests this won't be overwhelming.

PR #265, opened 2019-07-29
Add tophap.com community projects using H3

dekmaster:


isaacbrodsky:
Hi, thanks for your PR. We really appreciate seeing companies and projects using H3!

Unfortunately we are not going to merge this request, as our intention for the Application page is to provide resources for open source projects and plugins that can be used to create, transform, visualize, etc H3 data. We'd also be open to posting links to articles or blog posts describing H3 use cases.

PR #267, opened 2019-08-01
Add h3ToCenterChild and getPentagonsAtRes

### LINKED ISSUE #255: add `pentagonsAtRes` function?
Shall we add a function like `pentagonsAtRes(int res, H3Index* out)` which would provide the user with the set of pentagon cells at any given resolution?

@nrabinowitz mentioned that we might be able to do this with something like

```
setH3Index(&myH3Index, res, <some_pentagon_base_cell_index>, 0);
```

where the pentaton base cells are: `4, 14, 24, 38, 49, 58, 63, 72, 83, 97, 107, 117`

### LINKED ISSUE #263: Adding h3ToCenterChild
I have a use-case where I need to find the center child of a given h3 index at a specified resolution.

I'm simply changing the four resolution bits, but I am wondering if this is worth adding as a helper function to the core library (so I can keep treating the index as an opaque datatype) or if it is too specific a use-case.

I'm willing to make a pull request if you think it would be useful.

zachasme:
Resolves #255 and #263.

### DIFF
--- a/docs/api/hierarchy.md
+++ b/docs/api/hierarchy.md
@@ -32,7 +32,7 @@ Returns the size of the array needed by `h3ToChildren` for these inputs.
 H3Index h3ToCenterChild(H3Index h, int childRes);
 ```
 
-Returns the center child (finer) index containing `h` at resolution `childRes`.
+Returns the center child (finer) index contained by `h` at resolution `childRes`.
 
 ## compact
 
--- a/docs/api/hierarchy.md
+++ b/docs/api/hierarchy.md
@@ -26,6 +26,22 @@ int maxH3ToChildrenSize(H3Index h, int childRes);
 
 Returns the size of the array needed by `h3ToChildren` for these inputs.
 
+## h3ToChildren
+
+```
+void h3ToChildren(H3Index h, int childRes, H3Index *children);
+```
+
+Populates `children` with the indexes contained by `h` at resolution `childRes`. `children` must be an array of at least size `maxH3ToChildrenSize(h, childRes)`.
+
+## h3ToCenterChild
+
+```
+H3Index h3ToCenterChild(H3Index h, int childRes);
+```
+
+Returns the center child (finer) index containing `h` at resolution `childRes`.
+
 ## compact
 
 ```
--- a/docs/api/hierarchy.md
+++ b/docs/api/hierarchy.md
@@ -26,14 +26,6 @@ int maxH3ToChildrenSize(H3Index h, int childRes);
 
 Returns the size of the array needed by `h3ToChildren` for these inputs.
 
-## h3ToChildren
-
-```
-void h3ToChildren(H3Index h, int childRes, H3Index *children);
-```
-
-Populates `children` with the indexes contained by `h` at resolution `childRes`. `children` must be an array of at least size `maxH3ToChildrenSize(h, childRes)`.
-
 ## h3ToCenterChild
 
 ```
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -73,3 +73,19 @@ int res0IndexCount();
 ```
 
 Number of resolution 0 **H3** indexes.
+
+## getPentagonsAtRes
+
+```
+void getPentagonsAtRes(int res, H3Index *out);
+```
+
+All the pentagon **H3** indexes at specified resolution.
+
+## pentagonIndexCount
+
+```
+int pentagonIndexCount();
+```
+
+Number of pentagon **H3** indexes.
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -74,10 +74,10 @@ int res0IndexCount();
 
 Number of resolution 0 **H3** indexes.
 
-## getPentagonsAtRes
+## getPentagonIndexes
 
 ```
-void getPentagonsAtRes(int res, H3Index *out);
+void getPentagonIndexes(int res, H3Index *out);
 ```
 
 All the pentagon **H3** indexes at specified resolution.
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -65,6 +65,7 @@ void getRes0Indexes(H3Index *out);
 ```
 
 All the resolution 0 **H3** indexes.
+`out` must be an array of at least size `res0IndexCount()`.
 
 ## res0IndexCount
 
@@ -80,7 +81,8 @@ Number of resolution 0 **H3** indexes.
 void getPentagonIndexes(int res, H3Index *out);
 ```
 
-All the pentagon **H3** indexes at specified resolution.
+All the pentagon **H3** indexes at the specified resolution.
+`out` must be an array of at least size `pentagonIndexCount()`.
 
 ## pentagonIndexCount
 
@@ -88,4 +90,4 @@ All the pentagon **H3** indexes at specified resolution.
 int pentagonIndexCount();
 ```
 
-Number of pentagon **H3** indexes.
+Number of pentagon **H3** indexes per resolution.
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -90,4 +90,4 @@ All the pentagon **H3** indexes at the specified resolution.
 int pentagonIndexCount();
 ```
 
-Number of pentagon **H3** indexes per resolution.
+Number of pentagon **H3** indexes per resolution. This is always 12, but provided as a convenience.
--- a/src/apps/testapps/testCompact.c
+++ b/src/apps/testapps/testCompact.c
@@ -190,7 +190,6 @@ SUITE(compact) {
         }
         uncompactResult = H3_EXPORT(uncompact)(
             someHexagons, numHex, uncompressed, numHex * 7, MAX_H3_RES + 1);
-        fprintf(stderr, "@@\n%d\n", uncompactResult);
         t_assert(uncompactResult != 0,
                  "uncompact fails when given resolutions beyond max");
     }
--- a/src/apps/testapps/testCompact.c
+++ b/src/apps/testapps/testCompact.c
@@ -165,6 +165,10 @@ SUITE(compact) {
         sizeResult = H3_EXPORT(maxUncompactSize)(someHexagons, numHex, -1);
         t_assert(sizeResult < 0,
                  "maxUncompactSize fails when given illegal resolutions");
+        sizeResult =
+            H3_EXPORT(maxUncompactSize)(someHexagons, numHex, MAX_H3_RES + 1);
+        t_assert(sizeResult < 0,
+                 "maxUncompactSize fails when given resolutions beyond max");
 
         H3Index uncompressed[] = {0, 0, 0};
         int uncompactResult =
@@ -180,6 +184,15 @@ SUITE(compact) {
         t_assert(
             uncompactResult != 0,
             "uncompact fails when given too little buffer (same resolution)");
+
+        for (int i = 0; i < numHex; i++) {
+            setH3Index(&someHexagons[i], MAX_H3_RES, i, 0);
+        }
+        uncompactResult = H3_EXPORT(uncompact)(
+            someHexagons, numHex, uncompressed, numHex * 7, MAX_H3_RES + 1);
+        fprintf(stderr, "@@\n%d\n", uncompactResult);
+        t_assert(uncompactResult != 0,
+                 "uncompact fails when given resolutions beyond max");
     }
 
     TEST(someHexagon) {
--- a/src/apps/testapps/testH3ToCenterChild.c
+++ b/src/apps/testapps/testH3ToCenterChild.c
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017-2019 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include "h3Index.h"
+#include "test.h"
+
+SUITE(h3ToCenterChild) {
+    H3Index hex8;
+    GeoCoord centroid;
+    setH3Index(&hex8, 8, 4, 2);
+    H3_EXPORT(h3ToGeo)(hex8, &centroid);
+
+    TEST(twoStepRes) {
+        H3Index centerChild = H3_EXPORT(h3ToCenterChild)(hex8, 10);
+        H3Index geoChild = H3_EXPORT(geoToH3)(&centroid, 10);
+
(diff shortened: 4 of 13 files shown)


zachasme:
What are your thoughts on naming?

The `atRes` suffix is not used anywhere else, perhaps `getPentagonalIndexes`/`pentgonalIndexCount` would be better?

src/apps/testapps/testH3ToCenterChild.c
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include "h3Index.h"
+#include "test.h"
+
+SUITE(h3ToCenterChild) {
```
nrabinowitz: It wouldn't hurt to add an exhaustive test here - something like (pseudocode):
```
for res in 0 to 14:
  for childRes in res + 1 to 15:
    h3Index = indexAtRes(res)
    centerChild = h3ToCenterChild(h3Index, childRes)
    // some property-based tests
    t_assert(h3GetResolution(centerChild) == childRes)
    t_assert(h3GetParent(centerChild, res) == h3Index)
```

src/apps/testapps/testPentagonsAtRes.c
```diff
@@ -0,0 +1,46 @@
...
+
+        for (int res = 0; res <= 15; res++) {
+            H3Index indexes[PADDED_COUNT] = {0};
+            H3_EXPORT(getPentagonsAtRes)(res, indexes);
+
+            int numFound = 0;
+
+            for (int i = 0; i < PADDED_COUNT; i++) {
+                H3Index h = indexes[i];
+                if (H3_EXPORT(h3IsValid)(h)) {
```
nrabinowitz: Nit: This might make more sense as 
```
if (h) {
  t_assert(H3_EXPORT(h3IsValid(h)), "index should be valid");
  // ... etc
}
```
This might help differentiate errors where we got an invalid index from cases where the array position wasn't filled.

src/apps/testapps/testPentagonsAtRes.c
```diff
@@ -0,0 +1,46 @@
...
+        int expectedCount = H3_EXPORT(pentagonIndexCount)();
+
+        for (int res = 0; res <= 15; res++) {
+            H3Index indexes[PADDED_COUNT] = {0};
+            H3_EXPORT(getPentagonsAtRes)(res, indexes);
+
+            int numFound = 0;
+
+            for (int i = 0; i < PADDED_COUNT; i++) {
+                H3Index h = indexes[i];
```
nrabinowitz: Nit: We're not very consistent here, but I think we generally use `h3` or `h3Index` as a conventional variable name.

src/h3lib/lib/h3Index.c
```diff
@@ -214,6 +214,31 @@ void H3_EXPORT(h3ToChildren)(H3Index h, int childRes, H3Index* children) {
...
+/**
+ * h3ToCenterChild produces the center child index for a given H3 index at
+ * the specified resolution
+ *
+ * @param h H3Index to find center child of
+ * @param childRes The resolution to switch to
+ *
+ * @return H3Index of the center child, or 0 if you actually asked for a parent
+ */
+H3Index H3_EXPORT(h3ToCenterChild)(H3Index parentH, int childRes) {
```
nrabinowitz: Nit: I'd prefer just `parent` or `h` to `parentH`. Similarly, below, I'd just go with `child` or `childIndex` - we don't use `-H` as a suffix anywhere else.

src/h3lib/lib/h3Index.c
```diff
@@ -214,6 +214,31 @@ void H3_EXPORT(h3ToChildren)(H3Index h, int childRes, H3Index* children) {
...
+    int parentRes = H3_GET_RESOLUTION(parentH);
+    if (childRes < parentRes) {
+        return H3_INVALID_INDEX;
+    } else if (childRes == parentRes) {
+        return parentH;
+    } else if (childRes < 0 || childRes > MAX_H3_RES) {
```
nrabinowitz: Nit 1: `childRes < 0` should not be possible if we've checked `childRes < parentRes`.
Nit 2: I'd combine this branch with the `if (childRes < parentRes)` branch above to consolidate validation.

More substantively - we need a similar check in `h3ToChildren` and `uncompact`, but it doesn't look like we check `MAX_H3_RES` there. Could we add a static function like `static bool isValidChildRes(int parentRes, int childRes)` and use in all three functions?

src/h3lib/lib/h3Index.c
```diff
@@ -860,6 +885,29 @@ void H3_EXPORT(h3GetFaces)(H3Index h3, int* out) {
...
+
+/**
+ * Generates all pentagons at the specified resolution
+ *
+ * @param res The resolution to produce pentagons at
+ */
+void H3_EXPORT(getPentagonsAtRes)(int res, H3Index* out) {
+    int i = 0;
+    for (int bc = 0; bc < NUM_BASE_CELLS; bc++) {
+        if (_isBaseCellPentagon(bc)) {
```
nrabinowitz: Would it be easier just hard-coding the indexes of the 12 base cells here? I believe they're `4, 14, 24, 38, 49, 58, 63, 72, 83, 97, 107, 117` - we could add a constant in `baseCells.h`. OTOH, I doubt performance is an issue here, so 🤷‍♂

src/h3lib/lib/h3Index.c
```diff
@@ -860,6 +885,29 @@ void H3_EXPORT(h3GetFaces)(H3Index h3, int* out) {
...
+            H3Index pentagonH;
```
nrabinowitz: Nit: As above, prefer `pentagon` to `pentagonH`

nrabinowitz:
> What are your thoughts on naming?
> 
> The `atRes` suffix is not used anywhere else, perhaps `getPentagonalIndexes`/`pentgonalIndexCount` would be better?

We also don't use `pentagonal` anywhere :). I don't mind `atRes` but could see just `getPentagons` or `getPentagonIndexes`. No strong preference.

docs/api/hierarchy.md
```diff
@@ -26,6 +26,22 @@ int maxH3ToChildrenSize(H3Index h, int childRes);
...
+
+Populates `children` with the indexes contained by `h` at resolution `childRes`. `children` must be an array of at least size `maxH3ToChildrenSize(h, childRes)`.
+
+## h3ToCenterChild
+
+```
+H3Index h3ToCenterChild(H3Index h, int childRes);
+```
+
+Returns the center child (finer) index containing `h` at resolution `childRes`.
```
isaacbrodsky: "contained by `h`"?

docs/api/misc.md
```diff
@@ -73,3 +73,19 @@ int res0IndexCount();
...
 
 Number of resolution 0 **H3** indexes.
+
+## getPentagonsAtRes
+
+```
+void getPentagonsAtRes(int res, H3Index *out);
+```
+
+All the pentagon **H3** indexes at specified resolution.
```
isaacbrodsky: nit: "at the specified resolution."

Please add a note that `out` must be of size `pentagonIndexCount()`

docs/api/misc.md
```diff
@@ -73,3 +73,19 @@ int res0IndexCount();
...
+
+## pentagonIndexCount
+
+```
+int pentagonIndexCount();
+```
+
+Number of pentagon **H3** indexes.
```
isaacbrodsky: "indexes per resolution."

src/apps/testapps/testH3ToCenterChild.c
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2017-2019 Uber Technologies, Inc.
```
isaacbrodsky: nit: could mark it as "Copyright 2019 Uber Technologies, Inc."

src/apps/testapps/testPentagonsAtRes.c
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2017-2019 Uber Technologies, Inc.
```
isaacbrodsky: Same comment about copyright header

src/apps/testapps/testH3ToCenterChild.c
isaacbrodsky: +1, I'd like to see a more exhaustive test for this function

src/apps/testapps/testPentagonsAtRes.c
```diff
@@ -0,0 +1,46 @@
...
+
+            int numFound = 0;
+
+            for (int i = 0; i < PADDED_COUNT; i++) {
+                H3Index h = indexes[i];
+                if (H3_EXPORT(h3IsValid)(h)) {
+                    t_assert(H3_EXPORT(h3IsPentagon(h)), "only pentagons");
+                    t_assert(H3_EXPORT(h3GetResolution(h)) == res,
+                             "only correct resolution");
+                    numFound++;
```
isaacbrodsky: Can we check that the found pentagon indexes are all unique?

src/h3lib/lib/h3Index.c
```diff
@@ -860,6 +885,29 @@ void H3_EXPORT(h3GetFaces)(H3Index h3, int* out) {
     }
 }
 
+/**
+ * pentagonIndexCount returns the number of pentagons (same at any resolution)
+ *
+ * @return int count of pentagon indexes
+ */
+int H3_EXPORT(pentagonIndexCount)() { return 12; }
```
isaacbrodsky: Nit: Move `#define NUM_PENTAGONS 12` from `generateNumHexagons.c` to `constants.h` and reuse it here?

src/h3lib/lib/h3Index.c
```diff
@@ -860,6 +885,29 @@ void H3_EXPORT(h3GetFaces)(H3Index h3, int* out) {
...
+
+/**
+ * Generates all pentagons at the specified resolution
+ *
+ * @param res The resolution to produce pentagons at
```
isaacbrodsky: Please add a comment for `out` indicating how large the array should be

src/apps/testapps/testH3ToCenterChild.c
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    TEST(sameRes) {
+        t_assert(H3_EXPORT(h3ToCenterChild)(hex8, 8) == hex8,
+                 "Same resolution returns self");
+    }
+
+    TEST(invalidInputs) {
+        t_assert(H3_EXPORT(h3ToCenterChild)(hex8, 6) == 0,
+                 "Lower resolution fails");
```
nrabinowitz: Nit: We try to avoid "lower"/"higher" as potentially confusing - when referring to resolution, please use "finer" or "coarser".

zachasme:
Thank you for all the feedback! I believe I've addressed it all. Please let me know if I've missed something.

I've added `static bool _isValidChildRes(int, int)` which is now used in `h3ToChildren`, `maxH3ToChildrenSize`, `h3ToCenterChild`, `uncompact` and `maxUncompactSize` as per your suggestion.

src/h3lib/lib/h3Index.c
zachasme: I did consider this but felt bad about adding a hard-coded list in addition to `BaseCellData` which already contains the information. What would you prefer?

src/apps/testapps/testH3ToCenterChild.c
zachasme: Good suggestion. I've also moved here the check that the child should equal indexed centroid of parent at child resolution. Are there more properties that asserted here?

src/h3lib/lib/h3Index.c
nrabinowitz: Probably best to leave it as is - still constant time and probably not worth over-optimizing on performance for this function. So nvm :).

src/apps/testapps/testH3ToCenterChild.c
isaacbrodsky: Only other check I can think to include is that the returned index passes `h3IsValid`

docs/api/misc.md
```diff
@@ -73,3 +74,20 @@ int res0IndexCount();
...
+All the pentagon **H3** indexes at the specified resolution.
+`out` must be an array of at least size `pentagonIndexCount()`.
+
+## pentagonIndexCount
+
+```
+int pentagonIndexCount();
+```
+
+Number of pentagon **H3** indexes per resolution.
```
nrabinowitz: Nit: Might be worth noting that this is always 12, but provided as a convenience. That particular number is fairly fundamental to the grid system, extremely unlikely to change, and included in the docs, so we probably don't need to pretend it's an implementation detail.

src/apps/testapps/testCompact.c
```diff
@@ -180,6 +184,15 @@ SUITE(compact) {
         t_assert(
             uncompactResult != 0,
             "uncompact fails when given too little buffer (same resolution)");
+
+        for (int i = 0; i < numHex; i++) {
+            setH3Index(&someHexagons[i], MAX_H3_RES, i, 0);
+        }
+        uncompactResult = H3_EXPORT(uncompact)(
+            someHexagons, numHex, uncompressed, numHex * 7, MAX_H3_RES + 1);
+        fprintf(stderr, "@@\n%d\n", uncompactResult);
```
nrabinowitz: Please remove the `fprint` statement.

src/h3lib/lib/h3Index.c
```diff
@@ -145,6 +145,21 @@ H3Index H3_EXPORT(h3ToParent)(H3Index h, int parentRes) {
     return parentH;
 }
 
+/**
+ * Determines whether a one resolution is a valid child resolution of another
```
isaacbrodsky: I think "a" should be removed here

src/apps/testapps/testH3ToCenterChild.c
nrabinowitz: The assertions here look good to me 👍

src/apps/testapps/testPentagonIndexes.c
```diff
@@ -0,0 +1,59 @@
...
+                    t_assert(H3_EXPORT(h3IsPentagon(h3Index)),
+                             "index should be pentagon");
+                    t_assert(H3_EXPORT(h3GetResolution(h3Index)) == res,
+                             "index should have correct resolution");
+
+                    // verify uniqueness
+                    int indexSeen = 0;
+                    for (int j = i + 1; j < PADDED_COUNT; j++) {
+                        if (h3Indexes[j] == h3Index) {
+                            indexSeen++;
```
nrabinowitz: Nit: You could also drop `indexSeen` and simply have `t_assert(false, "index should be seen only once")` if found

src/apps/testapps/testCompact.c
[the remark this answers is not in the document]
zachasme: Oops :blush:

src/h3lib/lib/h3Index.c
```diff
@@ -145,6 +145,21 @@ H3Index H3_EXPORT(h3ToParent)(H3Index h, int parentRes) {
     return parentH;
 }
 
+/**
+ * Determines whether one resolution is a valid child resolution of another
```
isaacbrodsky: nit: could also note that child being the same as parent is considered valid.

PR #269, opened 2019-08-07
Exclude assertions from branch coverage

isaacbrodsky:
This doesn't meaningfully change what is tested, but it does remove some branch coverage warnings which I don't believe should be present (mainly assertions.) eddyxu/cpp-coveralls hasn't been updated to support reporting branch coverage to Coveralls. I noticed that okkez/coveralls-lcov says it supports reporting branch coverage, so I'm trying switching to that.

I also adjusted the use of some integer constants to use the appropriate enums instead.

Before:
```
Overall coverage rate:
  lines......: 100.0% (1917 of 1917 lines)
  functions..: 100.0% (172 of 172 functions)
  branches...: 93.0% (999 of 1074 branches)
```

After:
```
Overall coverage rate:
  lines......: 100.0% (1917 of 1917 lines)
  functions..: 100.0% (172 of 172 functions)
  branches...: 95.9% (963 of 1004 branches)
```


isaacbrodsky:
Coveralls now shows the branch coverage for this PR: https://coveralls.io/jobs/51835082

src/h3lib/lib/faceijk.c
```diff
@@ -684,14 +684,14 @@ void _faceIjkToGeoBoundary(const FaceIJK* h, int res, int isPentagon,
     // edge-crossing vertices as needed
     g->numVerts = 0;
     int lastFace = -1;
-    int lastOverage = 0;  // 0: none; 1: edge; 2: overage
+    Overage lastOverage = NO_OVERAGE;
```
nrabinowitz: Huh, thought I changed these. Thanks!

src/h3lib/lib/h3Index.c
```diff
@@ -293,7 +293,7 @@ int H3_EXPORT(compact)(const H3Index* h3Set, H3Index* compactedSet,
                 int loc = (int)(parent % numRemainingHexes);
                 int loopCount = 0;
                 while (hashSetArray[loc] != 0) {
-                    if (loopCount > numRemainingHexes) {
+                    if (loopCount > numRemainingHexes) {  // LCOV_EXCL_BR_LINE
                         // LCOV_EXCL_START
```
nrabinowitz: Do we need this comment too?

src/h3lib/lib/h3Index.c
```diff
@@ -369,7 +369,7 @@ int H3_EXPORT(compact)(const H3Index* h3Set, H3Index* compactedSet,
                 int loopCount = 0;
                 bool isUncompactable = true;
                 do {
-                    if (loopCount > numRemainingHexes) {
+                    if (loopCount > numRemainingHexes) {  // LCOV_EXCL_BR_LINE
                         // LCOV_EXCL_START
```
nrabinowitz: Ditto

src/h3lib/lib/h3Index.c
isaacbrodsky: `LCOV_EXCL_BR_LINE` only excludes checking for branch coverage on the line it's on. Lcov doesn't have a concept of a set of lines being in a branch, so it doesn't exclude this block.

isaacbrodsky:
> LG, though I'm not sure why the CI coverage drops 2.2%, which doesn't line up with what you have in the PR description.

I'm not sure about all the causes of differences between my run and what Coveralls shows, but I did just notice that Coveralls is showing assert lines as having uncovered branches, so I think we should track down where differences are before merging.

isaacbrodsky:
I upgraded the CI job to Ubuntu Xenial to get the newer lcov version. Build seems to work the same as before, but I did note some compiler warnings/notes about `vertexGraph.c`. The branch coverage it reported improved in that `assert` lines are no longer being counted against branch coverage, but there are some unknown differences still.

I noticed there are a few other places where LCOV_EXCL_LINE has been used (so we're marking the branched code as unreachable), it looks to me like in all cases where we have that we should also exclude the branch if we're doing that. (I considered making assertions instead of unreachable branches, but I didn't feel convinced that was a better solution.)

isaacbrodsky:
I think some of the coverage difference is from difference compilers. I see that some of the differences is in whether function prototype lines are counted as covered, which could be explained by using Clang locally vs GCC in CI.

PR #271, opened 2019-08-08
Fix validity check for IJK+ coordinates

saprykin:


isaacbrodsky:
Thank you for catching this! Would you mind adding a test case to demonstrate this is fixed?

saprykin:
I added additional test cases when j and k components of IJK coordinates are out of range. With previous version new tests are not failing though they must. Hopefully I added it to right place.

PR #273, opened 2019-08-12
Version 3.6.0

isaacbrodsky:
Added changelog entry for #271, and updated the other ones to point to #267 instead of #263 (our changelog entries usually point to the PR rather than the issue)


CHANGELOG.md
```diff
@@ -6,9 +6,13 @@ The public API of this library consists of the functions declared in file
 [h3api.h.in](./src/h3lib/include/h3api.h.in).
 
 ## [Unreleased]
+
+## [3.5.1] - 2019-08-12
```
dfellis: With new features added, shouldn't this be 3.6.0 instead?

isaacbrodsky: Oops, yes I believe it should.

PR #278, opened 2019-09-03
Handle empty (zero length) input to `compact`

isaacbrodsky:
`compact`, `maxUncompactSize`, and `uncompact` should handle the
case where empty input is provided. In this case the arrays
should not be dereferenced, since the first index is already out
of range.

This also adds tests for some uncovered branches.

### DIFF
--- a/src/apps/testapps/testPolygon.c
+++ b/src/apps/testapps/testPolygon.c
@@ -576,11 +576,6 @@ SUITE(polygon) {
         t_assert(result == NORMALIZATION_ERR_UNASSIGNED_HOLES,
                  "Expected error code returned");
 
-        t_assert(countLinkedPolygons(&polygon) == 1, "Polygon count correct");
-        t_assert(countLinkedLoops(&polygon) == 1,
-                 "Loop count on first polygon correct");
-        t_assert(polygon.first == outer, "Got expected outer loop");
-
         H3_EXPORT(destroyLinkedPolygon)(&polygon);
     }
 }
--- a/src/apps/testapps/testPolygon.c
+++ b/src/apps/testapps/testPolygon.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 Uber Technologies, Inc.
+ * Copyright 2017-2019 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/apps/testapps/testPolygon.c
```diff
@@ -553,4 +553,34 @@ SUITE(polygon) {
...
+
+        int result = normalizeMultiPolygon(&polygon);
+
+        t_assert(result == NORMALIZATION_ERR_UNASSIGNED_HOLES,
+                 "Expected error code returned");
+
+        t_assert(countLinkedPolygons(&polygon) == 1, "Polygon count correct");
+        t_assert(countLinkedLoops(&polygon) == 1,
+                 "Loop count on first polygon correct");
+        t_assert(polygon.first == outer, "Got expected outer loop");
```
nrabinowitz: Is it worth asserting output here? My assumption would be that if we are returning error codes, we relinquish all guarantees about output when you get one - so no behavior after an error code is returned should be assertable, because it's all undefined.

src/h3lib/lib/h3Index.c
```diff
@@ -266,6 +266,9 @@ H3Index H3_EXPORT(h3ToCenterChild)(H3Index h, int childRes) {
  */
 int H3_EXPORT(compact)(const H3Index* h3Set, H3Index* compactedSet,
                        const int numHexes) {
+    if (numHexes == 0) {
```
nrabinowitz: Do we need similar handling on `uncompact`?

src/apps/testapps/testPolygon.c
isaacbrodsky: It may not be, since the polygon doesn't have all the features anymore.

src/h3lib/lib/h3Index.c
isaacbrodsky: `uncompact` and `maxUncompactSize` have for loops like `for (int i = 0; i < numHexes; i++)`, so they are already exiting when the number of hexagons is 0. I added tests in this PR to confirm this.

src/h3lib/lib/h3Index.c
```diff
@@ -266,6 +266,9 @@ H3Index H3_EXPORT(h3ToCenterChild)(H3Index h, int childRes) {
  */
```
ajfriend: Can we be explicit about what codes are returned in success and failure cases?

isaacbrodsky: So far we haven't done that in the public API, I think with the idea that the error codes might change in patch changes. The error codes have been used in some of the bindings to generate exception messages, though, so they are useful.

Issue #195, opened 2019-02-23
Improvement to h3Line algorithm: Support any great arc


nrabinowitz:
Improved algo for `h3Line` that can cross arbitrary base cells with better fidelity than the current implementation:
- Take the great arc between start and end
- Subdivide the great arc by the icosahedron edges
- Use the current implementation to draw lines for each line segment
- Union the results

This should allow any arc >= 180 degrees, and ensure fidelity to the true great arc.

TODO: Prove that this is also the shortest grid distance (`h3Distance`).

nrabinowitz:
Note that this should fix #184

PR #279, opened 2019-09-11
trivial fix to bboxHexRadius to guarantee containment

### LINKED ISSUE #136: polyfill not filling the complete polygon
Hello,

I just stumbled on an issue where the `polyfill` function is not able to fully fill polygons. The problem  affects long, diagonal polygons, like this:

bug-r13

The blue part is the polygon, the green shape are the generated hexagons. Here is a closeup of the lower end of the polygon:

bug-r13-close

The `polyfill` was done at resolution 13. Interestingly the result at resolution 12 looks much better:

bug-r12-close

The GeoJSON for my test polygon is

    {"type":"Polygon","coordinates":[[[5.76910652153309,51.1122595579185],[5.74843667987113,51.0844285870957],[5.74868565718131,51.0841156106137],[5.76939661923247,51.1120877871771],[5.76910652153309,51.1122595579185]]]}

To further investigate this issue I went ahead and removed the line of code which does the removal of heaxgons located outside the polygon in this file:

https://github.com/uber/h3/blob/96a0e86d6a80f0d5e83e7a9d1fde34147847af27/src/h3lib/lib/algos.c#L710

I took this approach because I wondered about the clean cut at the end of the filled part  and wanted to see in which radius H3 is generating it indexes.

Running `polyfill` on my test polygon again gave the following results:

Resolution 12:

hack-r12

Resolution 13:

hack-r13

So the filled `minK` hex radius seems to be a bit too small for this case. I gave it a try and changed the 1.5 multiplicator in

https://github.com/uber/h3/blob/96a0e86d6a80f0d5e83e7a9d1fde34147847af27/src/h3lib/lib/bbox.c#L120

to the - wildly guessed value - 1.35:

    return (int)ceil(bboxRadiusKm / (1.35 * centerHexRadiusKm));

This value is certainly not mathematically correct and needs to be validated. It also increases the number of indexes being generated, but gave me a polygon completely covered in indexes:

hack-r13-crude-fix

So I assume the issue is in the `bboxHexRadius` function in combination with filling shapes with a large number of indexes. This function seems to have some deficits dealing with long, thin, diagonal shapes. 
I did not notice any issues when using `polyfill` with rounder geometries.

kwmsmith:
Fixes #136 , at least in my usecases.

### DIFF
--- a/src/apps/testapps/testPolyfill.c
+++ b/src/apps/testapps/testPolyfill.c
@@ -64,10 +64,10 @@ SUITE(polyfill) {
 
     TEST(maxPolyfillSize) {
         int numHexagons = H3_EXPORT(maxPolyfillSize)(&sfGeoPolygon, 9);
-        t_assert(numHexagons == 3169, "got expected max polyfill size");
+        t_assert(numHexagons == 7057, "got expected max polyfill size");
 
         numHexagons = H3_EXPORT(maxPolyfillSize)(&holeGeoPolygon, 9);
-        t_assert(numHexagons == 3169, "got expected max polyfill size (hole)");
+        t_assert(numHexagons == 7057, "got expected max polyfill size (hole)");
 
         numHexagons = H3_EXPORT(maxPolyfillSize)(&emptyGeoPolygon, 9);
         t_assert(numHexagons == 1, "got expected max polyfill size (empty)");
--- a/src/apps/testapps/testPolyfill_GH136.c
+++ b/src/apps/testapps/testPolyfill_GH136.c
@@ -41,7 +41,6 @@ static int countActualHexagons(H3Index* hexagons, int numHexagons) {
 }
 
 SUITE(polyfill_gh136) {
-
     testPolygon.geofence = testGeofence;
     testPolygon.numHoles = 0;
 
--- a/src/apps/testapps/testPolyfill_GH136.c
+++ b/src/apps/testapps/testPolyfill_GH136.c
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017-2018 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include "algos.h"
+#include "constants.h"
+#include "geoCoord.h"
+#include "h3Index.h"
+#include "test.h"
+
+// https://github.com/uber/h3/issues/136
+
+static GeoCoord testVerts[] = {{0.10068990369902957, 0.8920772174196191},
+                               {0.10032914690616246, 0.8915914753447348},
+                               {0.10033349237998787, 0.8915860128746426},
+                               {0.10069496685903621, 0.8920742194546231}};
+static Geofence testGeofence = {.numVerts = 4, .verts = testVerts};
+static GeoPolygon testPolygon;
+
+static int countActualHexagons(H3Index* hexagons, int numHexagons) {
+    int actualNumHexagons = 0;
+    for (int i = 0; i < numHexagons; i++) {
+        if (hexagons[i] != 0) {
+            actualNumHexagons++;
+        }
+    }
+    return actualNumHexagons;
+}
+
+SUITE(polyfill_gh136) {
+
+    testPolygon.geofence = testGeofence;
+    testPolygon.numHoles = 0;
+
+    TEST(gh136) {
+        int res = 13;
+        int numHexagons = H3_EXPORT(maxPolyfillSize)(&testPolygon, res);
+        H3Index* hexagons = calloc(numHexagons, sizeof(H3Index));
+
+        H3_EXPORT(polyfill)(&testPolygon, res, hexagons);
+        int actualNumHexagons = countActualHexagons(hexagons, numHexagons);
+
+        t_assert(actualNumHexagons == 4353, "got expected polyfill size");
+        free(hexagons);
+    }
+}
--- a/src/apps/testapps/testPolyfill_GH136.c
+++ b/src/apps/testapps/testPolyfill_GH136.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 Uber Technologies, Inc.
+ * Copyright 2017-2019 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/h3lib/lib/bbox.c
+++ b/src/h3lib/lib/bbox.c
@@ -117,5 +117,5 @@ int bboxHexRadius(const BBox* bbox, int res) {
     // any orientation of the GeoJSON encased in a circle defined by the
     // bounding box radius and center, it is guaranteed to fit in this k-ring
     // Rounded *up* to guarantee containment
-    return (int)ceil(bboxRadiusKm / (1.5 * centerHexRadiusKm));
+    return (int)ceil(bboxRadiusKm / (1.0 * centerHexRadiusKm));
 }
--- a/src/h3lib/lib/bbox.c
+++ b/src/h3lib/lib/bbox.c
@@ -112,10 +112,14 @@ int bboxHexRadius(const BBox* bbox, int res) {
     // Determine the radius of the center hexagon
     double centerHexRadiusKm = _hexRadiusKm(H3_EXPORT(geoToH3)(&center, res));
 
-    // The closest point along a hexagon drawn through the center points
-    // of a k-ring aggregation is exactly 1.5 radii of the hexagon. For
-    // any orientation of the GeoJSON encased in a circle defined by the
-    // bounding box radius and center, it is guaranteed to fit in this k-ring
-    // Rounded *up* to guarantee containment
-    return (int)ceil(bboxRadiusKm / (1.0 * centerHexRadiusKm));
+    // We use centerHexRadiusKm un-scaled and rounded *up* to guarantee
+    // containment ot the bbox.  Ideal, undistorted hexagons could scale
+    // centerHexRadiusKm by a factor of up to 1.5, reducing bboxHexRadius.
+    // This is because the closest point along an undistorted hexagon drawn
+    // through the center points of a k-ring aggregation is exactly 1.5 radii
+    // of the hexagon.  But there is distortion near pentagons, and for those
+    // cases, the scaling needs to be less than 1.5.  Using the un-scaled value
+    // conservatively guarantees containment for all cases, at the expense of a
+    // larger bboxHexRadius.
+    return (int)ceil(bboxRadiusKm / centerHexRadiusKm);
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The public API of this library consists of the functions declared in file
 ## [Unreleased]
 ### Fixed
 - `compact` handles zero length input correctly. (#278)
+- `bboxHexRadius` scaling factor adjusted to guarantee containment for `polyfill`. (#279)
 
 ## [3.6.0] - 2019-08-12
 ### Added
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,7 @@ set(OTHER_SOURCE_FILES
     src/apps/testapps/testVertexGraph.c
     src/apps/testapps/testCompact.c
     src/apps/testapps/testPolyfill.c
+    src/apps/testapps/testPolyfill_GH136.c
     src/apps/testapps/testPentagonIndexes.c
     src/apps/testapps/testKRing.c
     src/apps/testapps/testH3ToGeoBoundary.c
@@ -506,6 +507,7 @@ if(BUILD_TESTING)
     add_h3_test(testH3SetToVertexGraph src/apps/testapps/testH3SetToVertexGraph.c)
     add_h3_test(testLinkedGeo src/apps/testapps/testLinkedGeo.c)
     add_h3_test(testPolyfill src/apps/testapps/testPolyfill.c)
+    add_h3_test(testPolyfill_GH136 src/apps/testapps/testPolyfill_GH136.c)
     add_h3_test(testVertexGraph src/apps/testapps/testVertexGraph.c)
     add_h3_test(testH3UniEdge src/apps/testapps/testH3UniEdge.c)
     add_h3_test(testGeoCoord src/apps/testapps/testGeoCoord.c)


nrabinowitz:
This looks like it should be fine, but fails the polyfill tests...

dfellis:
It's possible that the test itself encoded a bad fill and was not noticed.

kwmsmith:
@nrabinowitz @dfellis tests pass, changelog updated.  Anything more?  If not, can this be merged?

src/apps/testapps/testPolyfill.c
```diff
@@ -64,10 +64,10 @@ SUITE(polyfill) {
 
     TEST(maxPolyfillSize) {
         int numHexagons = H3_EXPORT(maxPolyfillSize)(&sfGeoPolygon, 9);
-        t_assert(numHexagons == 3169, "got expected max polyfill size");
+        t_assert(numHexagons == 7057, "got expected max polyfill size");
```
nrabinowitz: This is quite a substantial change. Were we that far off previously?

I'm not sure if we can do this easily in the C library, but this is the kind of thing that would really benefit from visual snapshot testing. I'm going to take a look at rendering the two results here to verify - otherwise I'm really not sure how we can have any confidence in either number :/.

nrabinowitz: Oh... Sorry, nevermind. I read this as the `polyfill` output, not the `maxPolyfillSize` output, which of course is expected to change substantially in this diff.

The comment about visual snapshot testing still applies, but not to this specific change.

src/h3lib/lib/bbox.c
```diff
@@ -117,5 +117,5 @@ int bboxHexRadius(const BBox* bbox, int res) {
     // any orientation of the GeoJSON encased in a circle defined by the
     // bounding box radius and center, it is guaranteed to fit in this k-ring
     // Rounded *up* to guarantee containment
-    return (int)ceil(bboxRadiusKm / (1.5 * centerHexRadiusKm));
+    return (int)ceil(bboxRadiusKm / (1.0 * centerHexRadiusKm));
```
isaacbrodsky: The comment above describes the purpose of the 1.5 constant, but I think there's some missing reasoning that gets from that comment to the revised code. It might be good to update the comment to reflect the conversation on Github about why/how we'r choosing to estimate this.

minor: Seems like this could be `(int)ceil(bboxRadiusKm / centerHexRadiusKm)`

src/apps/testapps/testPolyfill.c
isaacbrodsky: I'd appreciate adding a test case which fails for the `1.5` constant but works with `1.0`.

kwmsmith:
@nrabinowitz output from `make benchmarks`:

```
Scanning dependencies of target bench_benchmarkH3Line
        -- h3LineNear: 37.550600 microseconds per iteration (10000 iterations)
        -- h3LineFar: 1777.783000 microseconds per iteration (1000 iterations)
[ 45%] Built target bench_benchmarkH3Line
[ 54%] Built target benchmarkH3Api
Scanning dependencies of target bench_benchmarkH3Api
        -- geoToH3: 1.534800 microseconds per iteration (10000 iterations)
        -- h3ToGeo: 0.836200 microseconds per iteration (10000 iterations)
        -- h3ToGeoBoundary: 3.015400 microseconds per iteration (10000 iterations)
[ 54%] Built target bench_benchmarkH3Api
[ 63%] Built target benchmarkPolygon
Scanning dependencies of target bench_benchmarkPolygon
        -- pointInsideGeofenceSmall: 0.078360 microseconds per iteration (100000 iterations)
        -- pointInsideGeofenceLarge: 0.700870 microseconds per iteration (100000 iterations)
        -- bboxFromGeofenceSmall: 0.063780 microseconds per iteration (100000 iterations)
        -- bboxFromGeofenceLarge: 0.914230 microseconds per iteration (100000 iterations)
[ 63%] Built target bench_benchmarkPolygon
[ 72%] Built target benchmarkH3SetToLinkedGeo
Scanning dependencies of target bench_benchmarkH3SetToLinkedGeo
        -- h3SetToLinkedGeoRing2: 84.195600 microseconds per iteration (10000 iterations)
        -- h3SetToLinkedGeoDonut: 32.966100 microseconds per iteration (10000 iterations)
        -- h3SetToLinkedGeoNestedDonuts: 134.701200 microseconds per iteration (10000 iterations)
[ 72%] Built target bench_benchmarkH3SetToLinkedGeo
[ 90%] Built target benchmarkKRing
Scanning dependencies of target bench_benchmarkKRing
        -- kRing10: 15.334900 microseconds per iteration (10000 iterations)
        -- kRing20: 60.004300 microseconds per iteration (10000 iterations)
        -- kRing30: 133.516100 microseconds per iteration (10000 iterations)
        -- kRing40: 236.836000 microseconds per iteration (10000 iterations)
        -- kRingPentagon10: 449.734000 microseconds per iteration (500 iterations)
        -- kRingPentagon20: 3745.942000 microseconds per iteration (500 iterations)
        -- kRingPentagon30: 11868.120000 microseconds per iteration (50 iterations)
        -- kRingPentagon40: 28947.200000 microseconds per iteration (10 iterations)
[ 90%] Built target bench_benchmarkKRing
[100%] Built target benchmarkPolyfill
Scanning dependencies of target bench_benchmarkPolyfill
        -- polyfillSF: 6377.710000 microseconds per iteration (500 iterations)
        -- polyfillAlameda: 10470.600000 microseconds per iteration (500 iterations)
        -- polyfillSouthernExpansion: 479781.800000 microseconds per iteration (10 iterations)
```

kwmsmith:
# Scaling 1.5
resolution 13
num hexagons: 3945
polyfill-1.5

## Scaling 1.5 detail
detail-1.5

# Scaling 1.0
resolution 13
num hexagons: 4353
polyfill-1.0

## Scaling 1.0 detail
detail-1.0

src/h3lib/lib/bbox.c
kwmsmith: See edited comment and cleaned up code.

src/apps/testapps/testPolyfill.c
kwmsmith: @isaacbrodsky test added:

`src/apps/testapps/testPolyfill_GH136.c`

kwmsmith:
I think all comments have been addressed:
* see above for benchmark comparison
* regression test added: `src/apps/testapps/testPolyfill_GH136.c`
* Visual comparison between old / new for one case.

Anything else?

nrabinowitz:
@kwmsmith Thanks so much for the follow-up here. I feel like ideally we'd be able to have an actual mathematically derived constant here, since if we could choose, say, `1.25` and only get 1.5x perf cost, that would be better. I'm unfortunately not sure how we'd calculate it, though. An empirical option would be to test on a large sample of rectangular polygons, and verify that the cells at each corner of each polygon were included in the polyfill, though in this case the issue is determining a large enough sample set.

All that said, I'd rather be correct than fast, and I think we should probably land this as is and address performance later.

kwmsmith:
> All that said, I'd rather be correct than fast, and I think we should probably land this as is and address performance later.

That's my strong preference -- I can swallow a 2x perf degradation.  But I can't swallow incorrect polyfill results :)

@dfellis mentioned re-implementing polyfill using `H3Line` or somesuch -- when would that better and faster implementation land?

nrabinowitz:
> @dfellis mentioned re-implementing polyfill using `H3Line` or somesuch -- when would that better and faster implementation land?

This is referring to adapting  the implementation we had in a prior version of H3, when we had a simpler hex grid. Honestly, we're a bit out from it, and I suspect this is a v4 feature.
- It depends on #195, which itself is a bit of work
- It may have weird cases around pentagons that need more investigation
- It would result in a *spherical* polyfill, where now we use a Cartesian polyfill. This is both a potentially breaking API change, and we'd need to implement spherical point-in-poly.

dfellis:
> > All that said, I'd rather be correct than fast, and I think we should probably land this as is and address performance later.
> 
> That's my strong preference -- I can swallow a 2x perf degradation. But I can't swallow incorrect polyfill results :)
> 
> @dfellis mentioned re-implementing polyfill using `H3Line` or somesuch -- when would that better and faster implementation land?

I don't really know when we'll get that done. I only have my spare time to work on this, anymore, since I'm not at Uber.

The dependency graph on that is:
[ ] Get H3Line draw lines across icosahedron edges (possible solution, detect icosahedron edge intersections, segment the line drawing for each icosahedron, and detect/eliminate the duplicate hexagon ID)
[ ] Implement the 2D area filling algorithm using [one of the algorithms described here](https://www.tutorialspoint.com/computer_graphics/polygon_filling_algorithm.htm) after tracing the polygon with H3Line and converting them to the 2D IJ coordinates. (The IJ coordinates may also require split across icosahedrons and then merging the results, I don't remember where we got with the icosahedron unfolding concept.)

So not impossible, and if I could dedicate all of my time on this, I could probably get it in a week, but since I can't, I don't really know what the ETA on that would be.

src/apps/testapps/testPolyfill_GH136.c
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2017-2018 Uber Technologies, Inc.
```
isaacbrodsky: nit: 2019

nrabinowitz:
FYI @dfellis and @kwmsmith - I'm looking into an alternate algorithm, prototyped here: https://observablehq.com/@nrabinowitz/spidering-polyfill

This doesn't depend on `h3Line`, considers a far smaller set of cells than the giant k-ring approach, and should have the same (or better) guarantees of correctness. Would be interested in any thoughts or comments.

dfellis:
> FYI @dfellis and @kwmsmith - I'm looking into an alternate algorithm, prototyped here: https://observablehq.com/@nrabinowitz/spidering-polyfill
> 
> This doesn't depend on `h3Line`, considers a far smaller set of cells than the giant k-ring approach, and should have the same (or better) guarantees of correctness. Would be interested in any thoughts or comments.

My thoughts:

That does seem like a mostly improved approach versus what we have now, assuming we fix an issue with it I'll go into more depth later.

It's probably slower than the speed H3Line + IJ-based filling could do (because of the repeated k-ring generation), but I'm not sure if it would be significantly slower.

The `maxPolyfillSize` could be reduced to taking the area of the bounding box and dividing it by the minimum area of a hexagon (or just a pentagon area to be safe).

The current implementation doesn't seem like anything but the first starting point is used, because it recurses in depth from that point (though I suppose it's possible that holes in the geometry could produce fully-isolated geometries and that it would be necessary to iterate them all), but I think a breadth-based search algorithm would be better: A while loop of searching for neighbors in the geofence and add them to the set, then repeat the process only for the neighbors found to be in the polygon and add their matching neighbors to the set and repeat. The recursion depth would be lower (as each iteration finds more of the interior hexagons) making a stack overflow less likely (actually, it should be possible to implement without any recursive function calls, just three `Set`s, one for all matching hexagons and one for the hexagons to test their neighbors on, and the third for the newly found neighbors, until the newly found neighbors set is empty).

Now that I'm thinking about it more, the recursion-less breadth-based search should also be faster, because the two sets for finding neighbors and the previous matches that should be tested can be allocated once at the beginning, and then at the end of the loop just reassigning the the new matches memory as the old matches variable and zeroing the old matches and reassigning to the new matches variable should work perfectly, while the recursive version has to allocate small chunks of memory on each step.

@nrabinowitz do you mind if I try to implement that? I think this is a great 80/20 rule type situation versus the "best" H3Line + IJ fill approach and I think it would only take me a day, which is much more motivating to me. :) Unless you want to do it, or you disagree with my comments above?

isaacbrodsky:
@dfellis I presume the sets are fixed length arrays? Are they just the size of `maxPolyfillSize`?

dfellis:
> @dfellis I presume the sets are fixed length arrays? Are they just the size of `maxPolyfillSize`?

I think that would be safest. If the polygon was guaranteed to be convex you could cut it down to just the number of hexagons needed just for the outer rim of the bounding box (as the guaranteed maximum, but a concavity in the polygon means instead of the average of 3 hexagons could be up to the whole 6 hexagons in the k-ring, and enough back-and-forth could cause an intermediate set to exceed the bounding box rim set.

nrabinowitz:
@dfellis I'm torn here, because I'd like to implement this myself but also realize that I won't have time to do so in the next few weeks. So I think it would be great if you want to take a shot at it. A few notes:

- There's now an iterative version at https://observablehq.com/@nrabinowitz/spidering-polyfill, which might be worth looking at. @isaacbrodsky and I had discussed using this implementation with a linked-list stack, which would allow us to allocate only necessary memory for the to-be-evaluated hexagons. You may be right, though, that allocating one list item at a time like this would be slower than a big array at the beginning, so up to you. Without a real stack, though, it seems like you'd need to iterate through the array each time to find the new candidates (unless I'm missing something)? The linked-list would allow constant-time `pop` and `push` for the candidate stack.

- I hadn't thought through `maxPolyfillSize`, but I like the bbox/average hex area approach. It's possible you'd undercount, though, e.g. if the input was a rectangle in a portion of the grid with larger-than-average hex area. So a scaling factor should probably be applied here.

- @isaacbrodsky and I spent some time thinking through this yesterday, and you need the initial set to include the 1-rings for *all* vertices and all *hole* vertices. Without this, we were able to identify shapes with non-traversable "isthmuses" (think e.g. an hourglass) that would prevent full polyfill. This isn't a big deal, as the point-in-poly logic already guarantees that the number of vertices is a factor in the time complexity.

Let me know if you want to discuss it any further. If you don't get to this relatively soon, I might find time to pick it up, but I'd ping you first.

dfellis:
> @dfellis I'm torn here, because I'd like to implement this myself but also realize that I won't have time to do so in the next few weeks. So I think it would be great if you want to take a shot at it. A few notes:

I'll have some time tomorrow to take a stab at it. I'll let you know how far along I get.

> 
>     * There's now an iterative version at https://observablehq.com/@nrabinowitz/spidering-polyfill, which might be worth looking at. @isaacbrodsky and I had discussed using this implementation with a linked-list stack, which would allow us to allocate only necessary memory for the to-be-evaluated hexagons. You may be right, though, that allocating one list item at a time like this would be slower than a big array at the beginning, so up to you. Without a real stack, though, it seems like you'd need to iterate through the array each time to find the new candidates (unless I'm missing something)? The linked-list would allow constant-time `pop` and `push` for the candidate stack.
> 

First point, thank you for that. That looks better, but still not quite what I mentioned above (that one's going to expand outward from the first point as far as it can, then it will go to the next point and try to expand, etc. I'm imagining something that takes all the points first, expands each of them and checks them, then takes that new set and expands them outward, etc). I'm not sure if there's any performance benefit on the way I was envisioning it and this one, though.

As to your question about a chunk of memory versus linked lists, that's assuming that we're going to use the same hash function as we've used before. I think here an algorithm that's more expensive to insert but much cheaper to iterate is the best idea, since the vast majority of the time I expect the number of output cells to be roughly equal to the number of input cells, which are based on the number of points defining the polygon (in the beginning I bet output ~= 3x input, but it will eventually level off and the last few iterations will have an output significantly smaller than the input as the k-rings keep running into already-found hexagons).

The final output set should probably be hashed normally since it's focused on `insert` and `has` checking but these two intermediate sets seem like it would be best to use an algorithm that takes the highest and lowest hexagon ID (by numeric value) seen in the prior set and tries to insert new hexagon IDs in quasi-sorted order, but if it runs into an existing hexagon there goes to the next slot and fills it in, using a "guess" on how many hexagons it's going to find in the next iteration to determine the proper spacing from low to high. That should produce a "mostly compact" set, and if we also have helper variables keeping a count of how many were inserted in each set, we can iterate the set until we find them all but not have to iterate over the entire chunk of memory allocated. `insert` will be slightly slower (collisions will be higher) and `has` would require a full iteration of the memory, but `iterate` would be very fast like the linked-list approach without the constant allocations.

>     * I hadn't thought through `maxPolyfillSize`, but I like the bbox/average hex area approach. It's possible you'd undercount, though, e.g. if the input was a rectangle in a portion of the grid with larger-than-average hex area. So a scaling factor should probably be applied here.
> 

That's why I proposed `bbox / pentagon area` -- it will overestimate, but it should never underestimate (I'm also weary of my ability to come up with a constant bigger than that since this whole conversation started from my not determining a good constant here).

>     * @isaacbrodsky and I spent some time thinking through this yesterday, and you need the initial set to include the 1-rings for _all_ vertices and all _hole_ vertices. Without this, we were able to identify shapes with non-traversable "isthmuses" (think e.g. an hourglass) that would prevent full polyfill. This isn't a big deal, as the point-in-poly logic already guarantees that the number of vertices is a factor in the time complexity.
>

Including all vertices seems perfectly fine.

> 
> Let me know if you want to discuss it any further. If you don't get to this relatively soon, I might find time to pick it up, but I'd ping you first.

Here's my discussion. :) Going to stop here for the day, though, and I'll try to get to prototyping something out tomorrow.

nrabinowitz:
@ajfriend helped identify a failure case here:

Screen Shot 2019-09-19 at 5 36 02 PM

Because you can't traverse between contained cells, you wouldn't get them with this algo :(.

dfellis:
I'm pretty sure a k = 2 approach to neighbor traversal will solve this problem. The algorithm would find the first and last hexagons, and at k = 2, we definitely catch these (because they are X, Y grid aligned and since this issue only occurs if the polygon is narrow and very straight in an X, Y coordinate system, k = 2 should be what's needed to find it).

Can you get me the polygon here so I can add a test case to my branch? I would hate to lose the 3-4x performance improvement because of this.

nrabinowitz:
> I'm pretty sure a k = 2 approach to neighbor traversal will solve this problem. The algorithm would find the first and last hexagons, and at k = 2, we definitely catch these (because they are X, Y grid aligned and since this issue only occurs if the polygon is narrow and very straight in an X, Y coordinate system, k = 2 should be what's needed to find it).
> 
> Can you get me the polygon here so I can add a test case to my branch? I would hate to lose the 3-4x performance improvement because of this.

This one was literally sketched on top of a screenshot, but I could come up with a real polygon.

I think k=2 won't do it, sadly. I think if you take a very long, very narrow triangle, center it on a given hex so that the hex center is contained, then rotate it at different angles, you can find angles w/r/t the grid where you'd have an arbitrarily large buffer of non-traversable cells.

I'm also bummed by this, and not sure whether we should care about these edge cases or not. The challenge is not only "Are we correct for a single weird polygon?", but "Are we correct for a set of contiguous polygons?" - in the latter case, poor handling of slivers could end up with "dead" cells in the set, clearly within the boundary of the set, but not assigned to any polygon.

dfellis:
> > I'm pretty sure a k = 2 approach to neighbor traversal will solve this problem. The algorithm would find the first and last hexagons, and at k = 2, we definitely catch these (because they are X, Y grid aligned and since this issue only occurs if the polygon is narrow and very straight in an X, Y coordinate system, k = 2 should be what's needed to find it).
> > Can you get me the polygon here so I can add a test case to my branch? I would hate to lose the 3-4x performance improvement because of this.
> 
> This one was literally sketched on top of a screenshot, but I could come up with a real polygon.
> 
> I think k=2 won't do it, sadly. I think if you take a very long, very narrow triangle, center it on a given hex so that the hex center is contained, then rotate it at different angles, you can find angles w/r/t the grid where you'd have an arbitrarily large buffer of non-traversable cells.
> 
> I'm also bummed by this, and not sure whether we should care about these edge cases or not. The challenge is not only "Are we correct for a single weird polygon?", but "Are we correct for a set of contiguous polygons?" - in the latter case, poor handling of slivers could end up with "dead" cells in the set, clearly within the boundary of the set, but not assigned to any polygon.

So I see what you're getting at now. This also indicts the H3Line approach outlined above since this polygon can't even draw a line between the points and maintain the contiguous-polygons-produce-contiguous-H3-Sets behavior.

But I do still have an idea for a solution here, and it involves H3Line. The algo uses the polygon's points to get likely starting hexagons to "spider" from, but there's no reason why that has to be the starting set. We could execute H3Line operations between the points in the polygons to get a larger set of hexagons to test from. This would let us get all of these hexagons because while we can't traverse from the points, we *can* traverse from places along the line segments.

It would only fail where a line segment is crossing an icosahedron (we can fall back to just the points for now and fix up H3Line later), and there would be problems with polygons that are large enough that a single segment is affected by the spherical vs cartesian projection differences (but that would also require that the polygon is one of these degenerate polygons as well at that large of a scale, which seems very unlikely and unworthy of supporting).

I would need to see how putting H3Line into the "spider" algorithm affects performance. I expect it to hurt it some, but I'm not sure by how much.

PR #291, opened 2019-11-27
Add FOSSA step to Travis CI

isaacbrodsky:
This should add automated scanning of dependencies and licenses used in the repo, similar to uber/h3-java#52. The core library itself doesn't have dependencies outside of the C standard, but this should pick up some dependencies from the website (not sure if it will pick up all because we do not `npm install` the website in CI) 

Replaces #231.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ addons:
 matrix:
   include:
       # Check that clang-format doesn't detect some files are not formatted.
-    - env: NAME="Formatting check"
+    - name: "Formatting check"
       compiler: clang
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


dfellis:
> not sure if it will pick up all because we do not `npm install` the website in CI

Why not add the install to that one? It shouldn't affect the other jobs and the CI already takes a godawful amount of time, so I don't think it would significantly slow things down.

isaacbrodsky:
> > not sure if it will pick up all because we do not `npm install` the website in CI
> 
> Why not add the install to that one? It shouldn't affect the other jobs and the CI already takes a godawful amount of time, so I don't think it would significantly slow things down.

Sure, added it to the Coveralls + FOSSA job.

PR #295, opened 2019-12-20
Additional tests for `polyfill`

isaacbrodsky:
There are some tests in this PR which we currently expect to fail.


src/apps/applib/lib/utility.c
```diff
@@ -233,3 +233,16 @@ void randomGeo(GeoCoord* g) {
         (((float)rand() / (float)(RAND_MAX)) * 180.0) - 90.0);
     g->lon = H3_EXPORT(degsToRads)((float)rand() / (float)(RAND_MAX)) * 360.0;
 }
+
+/**
+ * Returns the number of non-invalid indexes in the array.
+ */
+int countActualHexagons(H3Index* hexagons, int numHexagons) {
```
nrabinowitz: Would `countValidHexagons` be more accurate?

src/apps/testapps/testPolyfill.c
```diff
@@ -41,14 +42,68 @@ static GeoCoord emptyVerts[] = {{0.659966917655, -2.1364398519394},
...
+}
+
+static void fillIndex_assertions(H3Index h) {
+    if (isTransmeridianCell(h)) {
+        // TODO: these do not work correctly
+        return;
+    }
+
+    int currentRes = H3_EXPORT(h3GetResolution)(h);
+    for (int nextRes = currentRes; nextRes <= currentRes + 1; nextRes++) {
```
nrabinowitz: This only goes down one resolution, right? Should `1` be a clearly-named const like `testDepth`? If not, should this be a loop at all?

src/apps/testapps/testPolyfill.c
```diff
@@ -41,14 +42,68 @@ static GeoCoord emptyVerts[] = {{0.659966917655, -2.1364398519394},
...
+        if (bndry.verts[i].lon < minLon) minLon = bndry.verts[i].lon;
+        if (bndry.verts[i].lon > maxLon) maxLon = bndry.verts[i].lon;
+    }
+
+    return maxLon - minLon > M_PI - (M_PI / 4);
+}
+
+static void fillIndex_assertions(H3Index h) {
+    if (isTransmeridianCell(h)) {
+        // TODO: these do not work correctly
```
nrabinowitz: Is this true for all transmeridian cells, or just polar cells? I thought transmeridian cells in general should work correctly.

isaacbrodsky: I recall this was an issue for not only the cells at the poles.

nrabinowitz: If so, that's a problem. The `polyfill` algo is supposed to work correctly for transmeridian polygons, and we should have tests to confirm.

src/apps/applib/lib/utility.c
isaacbrodsky: They're not necessarily validated with `h3IsValid` - just checked for not being 0, which I take to mean "null" or "missing".

src/apps/testapps/testPolyfill.c
isaacbrodsky: I tried replacing `isTransmeridianCell` with this body, and the test suite failed:
```
    int res = H3_GET_RESOLUTION(h);
    GeoCoord northPoleCoords = {.lat = M_PI_2, .lon = 0};
    H3Index northPoleIndex = H3_EXPORT(geoToH3)(&northPoleCoords, res);
    if (northPoleIndex == h) {
        return true;
    }
    GeoCoord southPoleCoords = {.lat = -M_PI_2, .lon = 0};
    H3Index southPoleIndex = H3_EXPORT(geoToH3)(&southPoleCoords, res);
    if (southPoleIndex == h) {
        return true;
    }
    return false;
```

src/apps/testapps/testPolyfill.c
isaacbrodsky: I didn't test 2 (or more) because the assertion fails. Added a comment noting this.

Issue #298, opened 2020-01-05
`compact` behavior on duplicate input


ajfriend:
I'm seeing inconsistent behavior when I input a collection of duplicated hexagons to the `compact()` function.

Using the `cython` branch of `h3-py` (with a few extra `print` debugging statements):

```python
import numpy as np
import h3.api.memview_int as h3

def foo(n):
    h = h3.geo_to_h3(37, -122, 10)
    hexes = np.array([h] * n, dtype='uint64')

    return np.array(h3.compact(hexes))
```

When I run `foo(8)`, I get

```python
len(hu):  8
flag:  0
array([622204016658513919, 622204016658513919, 622204016658513919,
       622204016658513919, 622204016658513919, 622204016658513919,
       622204016658513919, 622204016658513919], dtype=uint64)
```

This is *mostly* OK, as some of the wrappers will dedupe the output with a `set` container. However, **should** the C `compact` function actually raise an error on any duplicate input?

When I run `foo(9)`, I get

```
len(hu):  9
flag:  -2
```
and an `H3ValueError`, due to the `-2` return flag.

It seems like the C code is detecting the duplicate hexagons in this case, but not the case where `n <= 8`.

- Should we try to catch this duplicate entry error consistently?
- Can we check for duplicate hexagons in all cases, or do we expect the user to input an array with no duplicates? If the latter, should we add that to the documentation of the function?
- Alternatively, should the `compact` function just handle and ignore duplicated hexagons? (We can "compact" any duplicates into a single hexagon.)

ajfriend:
Also, what about inputs to `compact` that contain hexagons of different resolutions. This seems to return without error, but perhaps we should detect this and raise an error?

For example, the code

```python
h1 = h3.geo_to_h3(37, -122, 10)
h2 = h3.geo_to_h3(37, -122, 9)

hexes = np.array([h1,h2], dtype='uint64')

np.array(h3.compact(hexes))
```

returns

```python
len(hu):  2
flag:  0
array([622204016658513919, 617700417031372799], dtype=uint64)
```

and the code

```python
h = h3.geo_to_h3(37, -122, 10)

hexes = [h] + list(h3.h3_to_children(h))
hexes = np.array(hexes, dtype='uint64')

np.array(h3.compact(hexes))
```

returns

```python
len(hu):  8
flag:  0
array([622204016658513919, 626707616285855743, 626707616285859839,
       626707616285863935, 626707616285868031, 626707616285872127,
[...]

PR #299, opened 2020-01-07
Detect and fail on duplicate input to `compact`

isaacbrodsky:
See #298.

There was an off by one error in detecting duplicate inputs to `compact`. All the comparisons assumed one-indexed counts, but only the second comparison (checking for whether all children were present in order to replace the parent) added one to the count. The first comparison (checking for duplicates) used the count directly from the array, which was zero-indexed.

### DIFF
--- a/src/apps/testapps/testCompact.c
+++ b/src/apps/testapps/testCompact.c
@@ -152,6 +152,52 @@ SUITE(compact) {
                  "compact fails on duplicate input");
     }
 
+    TEST(compact_duplicateMinimum) {
+        // Test that the minimum number of duplicate hexagons causes failure
+        H3Index h3;
+        // Arbitrary index
+        setH3Index(&h3, 5, 0, 2);
+
+        int arrSize = H3_EXPORT(maxH3ToChildrenSize)(h3, 11) + 1;
+        H3Index* children = calloc(arrSize, sizeof(H3Index));
+
+        H3_EXPORT(h3ToChildren)(h3, 11, children);
+        // duplicate one index
+        children[arrSize - 1] = children[0];
+
+        H3Index* output = calloc(arrSize, sizeof(H3Index));
+
+        int compactResult = H3_EXPORT(compact)(children, output, arrSize);
+        t_assert(compactResult != 0,
+                 "compact fails on duplicate input (single duplicate)");
+
+        free(output);
+        free(children);
+    }
+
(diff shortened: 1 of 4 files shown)


src/apps/testapps/testCompact.c
```diff
@@ -152,6 +152,52 @@ SUITE(compact) {
...
+        H3Index* children = calloc(arrSize, sizeof(H3Index));
+
+        H3_EXPORT(h3ToChildren)(h3, 11, children);
+        // duplicate one index
+        children[arrSize - 1] = children[0];
+
+        H3Index* output = calloc(arrSize, sizeof(H3Index));
+
+        int compactResult = H3_EXPORT(compact)(children, output, arrSize);
+        t_assert(compactResult != 0,
```
nrabinowitz: Should we assert specific error codes here? Should we have named constants for the error codes?

src/h3lib/lib/h3Index.c
```diff
@@ -314,7 +314,15 @@ int H3_EXPORT(compact)(const H3Index* h3Set, H3Index* compactedSet,
                         hashSetArray[loc] & H3_RESERVED_MASK_NEGATIVE;
                     if (tempIndex == parent) {
                         int count = H3_GET_RESERVED_BITS(hashSetArray[loc]) + 1;
-                        if (count > 7) {
+                        int limitCount = 7;
+                        if (H3_EXPORT(h3IsPentagon)(
```
dfellis: I suspect this will have a negative impact on the performance of `compact`. I wonder how much, though.

ajfriend:
Can we document the meaning of the different error codes, `-1`, `-2`?

ajfriend:
Does this guarantee to catch **any** duplicated cells? Or just a subset of cases that are "easy" to detect?

isaacbrodsky:
> Does this guarantee to catch **any** duplicated cells? Or just a subset of cases that are "easy" to detect?

I expect this will only detect the "easy" case that there are too many children of a parent cell. Duplicating a cell will be counted as another child of the parent.

src/apps/testapps/testCompact.c
isaacbrodsky: I've updated the PR to have that for the new tests (not updating the existing ones in this PR.) I think this is reasonable when the return codes are named.

Issue #297, opened 2020-01-01
Non-unique "valid" H3Index representations


ajfriend:
You can get some weird results by messing with the "reserved" bits in an H3Index. They seem to be ignored by most functions, so by changing them, you can get a **different** H3Index (in that the `uint64` values do not evaluate as equal, even though the correspond to the same H3 cell.

For example,

```python
def poison(h):
    'Set bits (0,5,6,7) to 1'
    ...

h = h3.geo_to_h3(0,0,9)
p = poison(h)

print('', h)
print(p)
```
returns
```
 89754e64993ffff
8f9754e64993ffff
```

Note that `h3.h3_is_valid(p) is True` and `p != h`. However, despite this lack of equality, they correspond to the same hex:
- `h3.h3_get_resolution(h) == h3.h3_get_resolution(p)`
- `h3.h3_to_geo(h) == h3.h3_to_geo(p)`

Note that the "poisoning" seems to carry through when using some common functions:

```python
print('', h3.h3_to_parent(h))
print(h3.h3_to_parent(p))
```
returns
```
 88754e6499fffff
8f8754e6499fffff
```

Even if all functions worked with these "poisoned" hexes, I think we would still have a problem: this breaks a user's expectation that `h1 != h2` would imply that `h1` and `h2` are different hexagons.

However, the "poisoning" does indeed seem to break some functions. For example, for normal hexes, we see the expected `compact` behavior:

```python
>>> hexes = h3.k_ring(h, 4)
>>> len(hexes)
61
>>> len(h3.compact(hexes))
31
```

However, after "poisoning" the same set of hexagons, compact doesn't seem to reduce the size of the hexagon set:

```python
>>> len(h3.compact(map(poison, hexes)))
61
```

Should we change the definition of a "valid" H3Index to be that the "reserved" bits must be zero?

dfellis:
I definitely agree that you've found a bug in h3IsValid. I don't think the other C functions should validate the inputs, though. The bindings could have a slower safe mode that calls it for each bound method, but if the user has prevalidated data in a database letting them reduce the runtime of their loop is worthwhile.

ajfriend:
So does everyone agree that for an H3Index to be valid, its "reserved" bits should all be 0?

If so, I guess the next steps would be to add a test for that in both `h3IsValid` (hexagons and pentagons) and `h3UnidirectionalEdgeIsValid` (unidirectional edges).

isaacbrodsky:
[...]

PR #300, opened 2020-01-07
Clarify that the reserved bits must be 0 for cell (mode 1) indexes.

isaacbrodsky:
Clarify the documentation and validation functions that the reserved
bits must be 0 for a cell index to be valid. This is done to prevent
valid indexes being different but refering to the same cell.

The high bit of an H3Index is also treated as a reserved bit
(previously documentation was silent on it.)

See #297.

### DIFF
--- a/docs/core-library/h3indexing.md
+++ b/docs/core-library/h3indexing.md
@@ -18,27 +18,27 @@ H3Index Representation
 
 The **H3Index** is the integer representation of an **H3** index, which can be placed into multiple modes to indicate the kind of concept being indexed. Mode 1 is an **H3** Cell (Hexagon) Index, mode 2 is an **H3** Unidirectional Edge (Hexagon A -> Hexagon B) Index, mode 3 is planned to be a bidirectional edge (Hexagon A <-> Hexagon B). Mode 0 is reserved and indicates an invalid **H3** index.
 
-The components of the **H3** cell index (mode 1) are packed into the lowest order 63 bits of a 64-bit integer in order as follows:
+The components of the **H3** cell index (mode 1) are packed into a 64-bit integer in order, highest bit first, as follows:
 
+* 1 bit reserved and set to 0,
 * 4 bits to indicate the index mode,
 * 3 bits reserved and set to 0,
 * 4 bits to indicate the cell resolution 0-15,
 * 7 bits to indicate the base cell 0-121, and
-* 3 bits to indicate each subsequent digit 0-6 from resolution 1 up to the resolution of the cell (45 bits total is reserved for resolutions 1-15)
+* 3 bits to indicate each subsequent digit 0-6 from resolution 1 up to the resolution of the cell (45 bits total are reserved for resolutions 1-15)
 
-The components of the **H3** unidirectional edge index (mode 2) are packed into the lowest order 63 bits of a 64-bit integer in order as follows:
+The components of the **H3** unidirectional edge index (mode 2) are packed into a 64-bit integer in order, highest bit first, as follows:
 
+* 1 bit reserved and set to 0,
 * 4 bits to indicate the index mode,
 * 3 bits to indicate the edge 1-6 of the cell to traverse,
 * 4 bits to indicate the cell resolution 0-15,
 * 7 bits to indicate the base cell 0-121, and
-* 3 bits to indicate each subsequent digit 0-6 from resolution 1 up to the resolution of the cell (45 bits total is reserved for resolutions 1-15)
+* 3 bits to indicate each subsequent digit 0-6 from resolution 1 up to the resolution of the cell (45 bits total are reserved for resolutions 1-15)
 
-The canonical string representation of an **H3Index** is the hexadecimal representation of the integer, using lowercase letters.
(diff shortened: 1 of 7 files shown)


docs/core-library/h3indexing.md
```diff
@@ -34,10 +34,12 @@ The components of the **H3** unidirectional edge index (mode 2) are packed into
 * 7 bits to indicate the base cell 0-121, and
 * 3 bits to indicate each subsequent digit 0-6 from resolution 1 up to the resolution of the cell (45 bits total is reserved for resolutions 1-15)
```
nrabinowitz: Nit while you're here: "45 bits total **are** reserved..."

docs/core-library/h3indexing.md
```diff
@@ -34,10 +34,12 @@ The components of the **H3** unidirectional edge index (mode 2) are packed into
...
 
-The canonical string representation of an **H3Index** is the hexadecimal representation of the integer.
+The canonical string representation of an **H3Index** is the hexadecimal representation of the integer, using lowercase letters.
 
 The three bits for each unused digit are set to 7.
 
+The highest order bit of an **H3Index** must be set to 0.
```
nrabinowitz: Worth adding something about how this ensures a positive integer?

dfellis: Unless it's considered a `uint64` where it'd be positive no matter what. :)

I think we did it just to reduce the amount of crap we'd have to deal with in Java.

src/h3lib/include/h3Index.h
```diff
@@ -46,6 +46,12 @@
 /** The number of bits in a single H3 resolution digit. */
 #define H3_PER_DIGIT_OFFSET 3
 
+/** 1 in the highest bit, 0's everywhere else. */
+#define H3_HIGH_BIT_MASK ((uint64_t)(1) << H3_MAX_OFFSET)
+
+/** 1 in the highest bit, 0's everywhere else. */
```
dfellis: This should be 0 in the highest bit, 1's everywhere else.

docs/core-library/h3indexing.md
```diff
@@ -34,10 +34,12 @@ The components of the **H3** unidirectional edge index (mode 2) are packed into
 * 7 bits to indicate the base cell 0-121, and
 * 3 bits to indicate each subsequent digit 0-6 from resolution 1 up to the resolution of the cell (45 bits total is reserved for resolutions 1-15)
 
-The canonical string representation of an **H3Index** is the hexadecimal representation of the integer.
+The canonical string representation of an **H3Index** is the hexadecimal representation of the integer, using lowercase letters.
```
ajfriend: Should we also specify that leading zeros should be excluded? For example, cells can be represented with 15 hex digits, like `89754e64993ffff`. I could imagine a user assuming that the hex representation would **always** be 16 hex digits, and might pad with a leading zero: `089754e64993ffff`.

This would correspond to the same cell, but the strings would not match.

docs/core-library/h3indexing.md
ajfriend: This feels weird to have stated at the end here. What about rewording the "the lowest order 63 bits of a 64-bit integer" stuff, and add "The highest order bit of an **H3Index** must be set to 0" to the beginning of the list above? That would keep everything in order.

src/h3lib/lib/h3UniEdge.c
```diff
@@ -179,7 +179,7 @@ int H3_EXPORT(h3UnidirectionalEdgeIsValid)(H3Index edge) {
         return 0;
     }
 
-    return H3_EXPORT(h3IsValid)(origin);
+    return H3_EXPORT(h3IsValid)(origin & H3_RESERVED_MASK_NEGATIVE);
```
ajfriend: Why do we need this change? `getOriginH3IndexFromUnidirectionalEdge` should already return a valid `origin` hex with the reserved bits properly set to zero.

dfellis: I think @isaacbrodsky meant to run this on the original `edge` index passed in, instead of on the generated `origin`.

ajfriend:
Do we want the property that `h3IsPentagon(h) is True` implies `h3IsValid(h) is True`? It doesn't look like this is currently the case, as `h3IsPentagon` doesn't check the reserved bits, for example.

Same question for `h3IsResClassIII`.

If not, it may be worth documenting that a user must run both `h3IsPentagon` and `h3IsValid` to ensure an H3Index is a valid pentagon.

docs/core-library/h3indexing.md
isaacbrodsky: @ajfriend good suggestion, better to rework that text to avoid ambiguity about only using some of the bits.

src/h3lib/lib/h3UniEdge.c
isaacbrodsky: Yes, this is actually a no-op. Removed the mask, and rely on a test to ensure the highest bit is passed through and checked in `h3IsValid`.

ajfriend:
Also, for both `h3IsPentagon` and `h3IsResClassIII`, I think they'll return `True` for edges whose origin passes the test. Should we check the mode of the H3Index in these functions?

PR #308, opened 2020-02-03
RFC: Names for H3 Concepts, Types, and Functions

ajfriend:
Starting a draft on the naming RFC.

Question: As an as-little-as-I-can-get-away-with Git user, how do I avoid picking up these already landed changes from @nrabinowitz?


dfellis:
> Starting a draft on the naming RFC.
> 
> Question: As an as-little-as-I-can-get-away-with Git user, how do I avoid picking up these already landed changes from @nrabinowitz?

`git rebase -i master` and then delete the line with @nrabinowitz 's commit.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -31,8 +33,28 @@ Reaching a consensus on the precise, technical language used when discussing H3
 
 ### Notes
 
-We concede that "hexagon" will still be used informally to refer to the concept of "cell" (As the core team, we do it ourselves!). This should be expected in casual, informal discussions of H3. When **precision** is required, however, we advise the use of strict technical terms like "index", "cell", "hexagon", "pentagon", etc.
+We realize that "hexagon" will still be used informally to refer to the concept of "cell" (As the development team, we do it ourselves!). This should be expected in casual, informal discussions of H3. When **precision** is required, however, we advise the use of strict technical terms like "index", "cell", "hexagon", "pentagon", etc.
 
 ## Use of "Topological"
 
-todo: define "topological"
+We should clarify in the documentation that H3 works with **topological** hexagons and pentagons, rather than true **geometric** hexagons and pentagons.
```
dfellis: Can we also clarify here that the vast majority of hexagons (and half the pentagons) are both topological and geometric hexagons (or pentagons)? It's only on icosahedron boundaries that extra points are added to deal with the two gnomonic projections adjacent to each other.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -31,8 +33,28 @@ Reaching a consensus on the precise, technical language used when discussing H3
...
+
+Instead of `UnidirectionalEdge`, use the term `DirectedEdge` or `DiEdge`.
+
+For a future undirected edge mode, use the term `UndirectedEdge` or `Edge`.
+
+Function name changes:
+
+|                  Current name                 |          Proposed name           |
+|-----------------------------------------------|----------------------------------|
+| `getH3UnidirectionalEdge`                     | `getH3DiEdge`                    |
```
dfellis: I understand this is shorter, but `DiEdge` sounds weird to me, and would require lookup in the documentation, while `DirectedEdge` is still clear.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -31,8 +33,28 @@ Reaching a consensus on the precise, technical language used when discussing H3
...
+| `h3UnidirectionalEdgeIsValid`                 | `h3DiEdgeIsValid`                |
+| `getOriginH3IndexFromUnidirectionalEdge`      | `getOriginH3CellFromDiEdge`      |
+| `getDestinationH3IndexFromUnidirectionalEdge` | `getDestinationH3CellFromDiEdge` |
+| `getH3IndexesFromUnidirectionalEdge`          | `getH3CellsFromDiEdge`           |
+| `getH3UnidirectionalEdgesFromHexagon`         | `getH3DiEdgesFromCell`           |
```
dfellis: Can we put some documentation above in the `Concepts` section on what a `Cell` is versus an `Index`?

Index is any 64-bit integer representing an H3 object, Cells are hexagons or pentagons, and DirectedEdges are traversals from an origin cell to an adjacent destination cell.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,60 @@
...
+Reaching a consensus on the precise, technical language used when discussing H3 concepts will clarify future library discussions and improve end-user documentation.
+
+## Proposals
+
+## Concepts
+
+The following technical terms should be used in the documentation, the H3 codebase, and precise technical discussions of the library.
+
+- `H3Index`, "H3 index", or "index"
+    - an unsigned 64-bit integer representing **any** H3 object (hexagon, pengaton, unidirectional edge, ...)
```
nrabinowitz: Nit: `pentagon`

nrabinowitz: More substantively - we currently refer to this in function names as `h3`, e.g. `h3ToGeo` etc. It would be good to have an opinion on whether we should keep this convention in function names, or whether this RFC would suggest updating to `cellToGeo`, etc.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
nrabinowitz: +1 for `DirectedEdge`. Much more obvious. Compare e.g. Googling ["diedge"](https://www.google.com/search?q=diedge) vs Googling ["directed edge"](https://www.google.com/search?q=directed+edge) - the former doesn't even put you in the right topical area, while "directed edge" gives you a clear definition immediately.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,60 @@
...
+For a future undirected edge mode, use the term `UndirectedEdge` or `Edge`.
+
+Function name changes:
+
+|                  Current name                 |          Proposed name           |
+|-----------------------------------------------|----------------------------------|
+| `getH3UnidirectionalEdge`                     | `getH3DiEdge`                    |
+| `h3UnidirectionalEdgeIsValid`                 | `h3DiEdgeIsValid`                |
+| `getOriginH3IndexFromUnidirectionalEdge`      | `getOriginH3CellFromDiEdge`      |
+| `getDestinationH3IndexFromUnidirectionalEdge` | `getDestinationH3CellFromDiEdge` |
```
nrabinowitz: Per my comment above, I don't think our use of `H3` in these function names is clear or useful. I'm not even sure in the two above whether we need to specify the output type - `getDestinationFromDirectedEdge` seems perfectly clear if you understand a `DirectedEdge` as connecting two cells (and if you don't, you probably don't need this function).

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,60 @@
...
+| `getH3IndexesFromUnidirectionalEdge`          | `getH3CellsFromDiEdge`           |
+| `getH3UnidirectionalEdgesFromHexagon`         | `getH3DiEdgesFromCell`           |
```
nrabinowitz: As per above, including `H3` here seems unnecessary and bloats the name. Suggest `getDirectedEdgesFromCell` or just `getDirectedEdges`.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,60 @@
...
+
+For a future undirected edge mode, use the term `UndirectedEdge` or `Edge`.
+
+Function name changes:
+
+|                  Current name                 |          Proposed name           |
+|-----------------------------------------------|----------------------------------|
+| `getH3UnidirectionalEdge`                     | `getH3DiEdge`                    |
+| `h3UnidirectionalEdgeIsValid`                 | `h3DiEdgeIsValid`                |
+| `getOriginH3IndexFromUnidirectionalEdge`      | `getOriginH3CellFromDiEdge`      |
```
nrabinowitz: Is the `from` clause really necessary in these names? It's clear, but is it worth the name bloat? I'm not wholly opposed, but I think most of these could work with simpler names like `getOriginCell` and `getDirectedEdges`

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,60 @@
...
+## Proposals
+
+## Concepts
+
+The following technical terms should be used in the documentation, the H3 codebase, and precise technical discussions of the library.
+
+- `H3Index`, "H3 index", or "index"
+    - an unsigned 64-bit integer representing **any** H3 object (hexagon, pengaton, unidirectional edge, ...)
+    - often represented as a 15-character (or 16-character) hexadecimal string, like `'8928308280fffff'`
+- cell
```
nrabinowitz: Nit: Should `H3 cell` be a preferred form? One question here and above is when to use `H3 index|cell` and when to use the short form, particularly in function names.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: Fair enough. `DirectedEdge` it shall be!

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: I think for a directed edge, `from` does indicate that the given cell would be the origin for all the returned hexes.

For `getDirectedEdges`, I think it's unclear what the argument should be. Get the directed edges corresponding to an undirected edge? Return all 12 directed edges associated to a cell? Return 6 directed edges starting from the cell?

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: How's it look now?

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: Oh, I'm all for doing something like `cellToGeo`!

I was thinking of doing the function renaming (`k_ring`, etc.) in a separate RFC, but I think there's enough overlap that I can just turn this into an all-encompassing "naming" RFC.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: Agreed. I tried out `getDirectedEdgeDestination`. Let me know what you think!

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,96 @@
...
+| `hexRing`                     | `ring`                | hollow ring                             |
+| `numHexagons`                 | `numCells`            |                                         |
+| `getRes0Indexes`              | `getRes0Cells`        |                                         |
+| `getPentagonIndexes`          | `getPentagons`        |                                         |
+| `h3GetBaseCell`               | `getBaseCellNumber`   |                                         |
+| `h3Distance`                  | `gridDistance`        | no `get` prefix?                        |
+| `h3Line`                      | `cellPath`            | future: `edgePath`, `directedEdgePath`? |
+
+
+For the `kRing`, `hexRange`, `hexRing`, etc. family of functions, should we come up with some standard prefix or suffix to denote that the function will fail if it encounters a pentagon?
```
nrabinowitz: That's only true for `hexRange`, right? Should we even expose that, or just expose `neighbors` and `ring`?

dfellis:
Just going to leave one meta-comment here: While all of the officially supported bindings are in languages where an object with the name `h3` effectively namespaces these functions, in C there is no such thing, and I believe that is the source of the original naming conventions: to prevent accidental collisions with other C functions pulled into the same project. (@sahrk can confirm or deny that reasoning. Also I do believe he was interested in cleaning up the naming conventions and had some ideas on that, as well. Since we're finally planning the breaking change to the library to do that, we should ask him, too. :) )

Because of that, it may make sense to keep the `h3` prefix on the C functions and have the bindings strip that back off?

nrabinowitz:
> Because of that, it may make sense to keep the `h3` prefix on the C functions and have the bindings strip that back off?

Would this mean prefixing _every_ public function with `h3`? I'm not opposed, if there's a good C reason to do this, though I'm not excited about having different function names in C and the bindings.

nrabinowitz:
> > Because of that, it may make sense to keep the `h3` prefix on the C functions and have the bindings strip that back off?
> 
> Would this mean prefixing _every_ public function with `h3`? I'm not opposed, if there's a good C reason to do this, though I'm not excited about having different function names in C and the bindings.

Also I thought that we had the `H3_EXPORT` macro specifically to address this functional concern in C?

dfellis:
> > > Because of that, it may make sense to keep the `h3` prefix on the C functions and have the bindings strip that back off?
> > 
> > 
> > Would this mean prefixing _every_ public function with `h3`? I'm not opposed, if there's a good C reason to do this, though I'm not excited about having different function names in C and the bindings.
> 
> Also I thought that we had the `H3_EXPORT` macro specifically to address this functional concern in C?

We do, but I don't think it does any adjustment to the function name by default? Users of the library could recompile it with their own naming scheme, but I just mean that the official bindings (excluding h3-js) could break when the code is dynamically loaded into the languages' runtime. I'm just putting out there that maybe we should have `H3_EXPORT` default to add a prefix instead of not?

We also haven't heard any reports of issues of this sort, so perhaps the names are unique enough, already? (But the proposed names like `ring` and `neighbors` seem general enough that that might not be the case in the future.)

ajfriend:
> > > > Because of that, it may make sense to keep the `h3` prefix on the C functions and have the bindings strip that back off?
> > > 
> > > 
> > > Would this mean prefixing _every_ public function with `h3`? I'm not opposed, if there's a good C reason to do this, though I'm not excited about having different function names in C and the bindings.
> > 
> > 
> > Also I thought that we had the `H3_EXPORT` macro specifically to address this functional concern in C?
> 
> We do, but I don't think it does any adjustment to the function name by default? Users of the library could recompile it with their own naming scheme, but I just mean that the official bindings (excluding h3-js) could break when the code is dynamically loaded into the languages' runtime. I'm just putting out there that maybe we should have `H3_EXPORT` default to add a prefix instead of not?
> 
> We also haven't heard any reports of issues of this sort, so perhaps the names are unique enough, already? (But the proposed names like `ring` and `neighbors` seem general enough that that might not be the case in the future.)

We should definitely figure out a standard way to do this. 

Also, if we add some h3 prefix, would we also want to do that for current names like `degsToRads`, `compact`, `polyfill`, `uncompact`?

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,96 @@
...
+| (doesn't exist)               | `isValidIndex`        |                                         |
+| `h3IsValid`                   | `isValidCell`         |                                         |
+| `h3UnidirectionalEdgeIsValid` | `isValidDirectedEdge` |                                         |
+| `h3IsPentagon`                | `isPentagon`          |                                         |
+| `h3ToParent`                  | `cellToParent`        |                                         |
+| `h3ToChildren`                | `cellToChildren`      |                                         |
+| `h3IndexesAreNeighbors`       | `cellsAreNeighbors`   |                                         |
+| `geoToH3`                     | `geoToCell`           |                                         |
+| `h3ToGeo`                     | `cellToGeo`           |                                         |
+| `kRing`                       | `disk`                | filled-in disk                          |
```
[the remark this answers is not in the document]
ajfriend: Using `neighbors` for `disk` would conflict with the use of `neighbors` in `cellsAreNeighbors`.

I think `cellsWithinRadius` and `cellsAtRadius` would be more descriptive than the names we have now, but I do like the concision of `ring` and `disk`.

Also, thinking about how many verbal conversations I already have around "k-rings", it would be less of a mouthful to say `ring` and `disk` than the `cells*Radius` names. Thinking about how many times I've said, "Do you mean a hollow ring, or a filled-in ring?", it'd be great to have short, precise names for these concepts.

dfellis:
> > > > > Because of that, it may make sense to keep the `h3` prefix on the C functions and have the bindings strip that back off?
> > > > 
> > > > 
> > > > Would this mean prefixing _every_ public function with `h3`? I'm not opposed, if there's a good C reason to do this, though I'm not excited about having different function names in C and the bindings.
> > > 
> > > 
> > > Also I thought that we had the `H3_EXPORT` macro specifically to address this functional concern in C?
> > 
> > 
> > We do, but I don't think it does any adjustment to the function name by default? Users of the library could recompile it with their own naming scheme, but I just mean that the official bindings (excluding h3-js) could break when the code is dynamically loaded into the languages' runtime. I'm just putting out there that maybe we should have `H3_EXPORT` default to add a prefix instead of not?
> > We also haven't heard any reports of issues of this sort, so perhaps the names are unique enough, already? (But the proposed names like `ring` and `neighbors` seem general enough that that might not be the case in the future.)
> 
> We should definitely figure out a standard way to do this.
> 
> Also, if we add some h3 prefix, would we also want to do that for current names like `degsToRads`, `compact`, `polyfill`, `uncompact`?

So the `H3_EXPORT` macro is the standard way, it'll apply to all of them equally.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,96 @@
...
+The following technical terms should be used in the documentation, the H3 codebase, and precise technical discussions of the library.
+
+- `Index` or `H3Index`:
+    - an unsigned 64-bit integer representing **any** H3 object (hexagon, pentagon, directed edge, ...)
+    - often represented as a 15-character (or 16-character) hexadecimal string, like `'8928308280fffff'`
+- `Cell`:
+    - an H3 index of `mode 1` (hexagon or pentagon)
+    - for functions that can handle either hexagons or pentagons, the more general term "cell" should be used whenever possible
+- hexagon:
+    - an H3 cell which is a **topological** hexagon
```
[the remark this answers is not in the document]
isaacbrodsky: I appreciate the specificity of this RFC, but in common usage the term _hexagon_ gets used a lot - could we confirm with users that this naming will be readable and searchable without going through a glossary? I can think of some alternatives which are a little less specific (_cell_ becomes _hexagon_ and _hexagon_ becomes something like _strict hexagon_).

sahrk:
FWIW, here is a little digression on the way in which I use this terminology... To me a `cell` is the atomic unit of location in a DGGS. Each `cell` has a representation in multiple location reference systems; as a point or polygon on the projection plane (`hex2d`) or on the sphere (`geo`), as an `H3` index, and as a `FaceIJK` index, and other representations might be created later for specific use cases (like the now extinct `H3Fat` index). Each of these are different materializations of the same `cell`. Imagine you are processing a list of `cell`'s; those `cell`'s might start out as `H3` indexes, but at some point you might transform them into a list of `FaceIJK` indexes to accomplish some task on the plane, and then transform them back to `H3` indexes. At each step I would say you are working with a list of `cell`'s, though not always with a list of `H3` indexes. So in my usage `cell` is not synonymous with `H3 Index`. 

This terminology developed in the context of my `DGGRID`, which had the specific design goal of abstracting the idea of location away from a specific representation. But that's a lot easier to pull off efficiently in `C++` than in `C`, and it was never a design goal of `H3`, where the notion of a `cell` seems tightly coupled to its `H3Index`. So I can see the argument for using the terms as if they are synonyms, though semantic imprecision may reduce your naming flexibility in unexpected ways down the road.

With that background, @dfellis, when I used an `h3` prefix in naming functions it was to designate the particular representation of the cell that is taken as input. For example, `h3IsValid` as opposed to `h3FatIsValid`. Most of the functions are transformation functions, as in `h3ToGeo`, which converts the cell from an `H3 Index` representation to a representation as a geographic point.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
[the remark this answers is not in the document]
ajfriend: I tried to capture this in the "Concepts" -> "Notes" section.

```
We realize that "hexagon" will still be used informally to refer to the concept of "cell" (As the development team, we do it ourselves!).
This should be expected in casual, informal discussions of H3.
When **precision** is required, however, we advise the use of strict technical terms like "index", "cell", "hexagon", "pentagon", etc.
In the codebase and in the documentation, strictly correct terminology should always be used.
```

Let me know what you think of that. Maybe we should make sure this is more prominent. 

I think the hexagon/cell/strict hexagon stuff could be handled by having this explained early and loudly in the documentation. Users shouldn't need to look at the whole glossary, but this one concept is important enough that we can lead with it, making it very hard for a user to miss.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,96 @@
...
+| `geoToH3`                     | `geoToCell`           |                                         |
+| `h3ToGeo`                     | `cellToGeo`           |                                         |
+| `kRing`                       | `disk`                | filled-in disk                          |
+| `hexRing`                     | `ring`                | hollow ring                             |
+| `numHexagons`                 | `numCells`            |                                         |
+| `getRes0Indexes`              | `getRes0Cells`        |                                         |
+| `getPentagonIndexes`          | `getPentagons`        |                                         |
+| `h3GetBaseCell`               | `getBaseCellNumber`   |                                         |
+| `h3Distance`                  | `gridDistance`        | no `get` prefix?                        |
+| `h3Line`                      | `cellPath`            | future: `edgePath`, `directedEdgePath`? |
```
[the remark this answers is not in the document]
ajfriend: Agreed.

How about `gridDistance` and `gridPath`?

**Or**, should it be `gridPathCell`, anticipating possible future functions like `gridPathEdge` and `gridPathDirectedEdge`?

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,96 @@
...
+| `h3ToChildren`                | `cellToChildren`      |                                         |
+| `h3IndexesAreNeighbors`       | `cellsAreNeighbors`   |                                         |
+| `geoToH3`                     | `geoToCell`           |                                         |
+| `h3ToGeo`                     | `cellToGeo`           |                                         |
+| `kRing`                       | `disk`                | filled-in disk                          |
+| `hexRing`                     | `ring`                | hollow ring                             |
+| `numHexagons`                 | `numCells`            |                                         |
+| `getRes0Indexes`              | `getRes0Cells`        |                                         |
+| `getPentagonIndexes`          | `getPentagons`        |                                         |
+| `h3GetBaseCell`               | `getBaseCellNumber`   |                                         |
```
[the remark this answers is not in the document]
ajfriend: How about this:

```
- `H3Index`:
    - an unsigned 64-bit integer representing **any** H3 object (hexagon, pentagon, directed edge, ...)
    - often represented as a 15-character (or 16-character) hexadecimal string, like `'8928308280fffff'`
    - the full term "H3 index" should be used to avoid confusion with other common uses of "index"; when a "traditional" index is needed, prefer using "number", "pos", or another term to avoid confusion
```

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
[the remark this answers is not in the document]
nrabinowitz: I think `grid` here implies `cell` (if not, we should use `cellDistance` and `cellPath`). If we ever wanted to provide an edge path, I'd go with e.g. `directedEdgePath`.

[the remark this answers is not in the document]
ajfriend: I was thinking that `grid` implies the cell/edge **graph**, but not specifically the `cell` (the nodes in the graph).

`cellDistance` and `cellPath` don't make me think of the graph, necessarily. 

The same "grid path" between two cells could be represented a number of ways:

- an ordered sequence of cells (`gridPath` or `gridPathCell`)
- a collection of directed edges (`gridPathDirectedEdge`)
- a collection of undirected edges (`gridPathEdge`)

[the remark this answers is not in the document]
nrabinowitz: I guess I don't see where the addition of the term `grid` for these functions adds any clarity beyond what's implied by `cellPath`, `edgePath`, etc. Cells, edges, and directed edges are already in a grid by definition.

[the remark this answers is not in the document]
ajfriend: Above, I thought you were concerned about "I feel like `h3Line` and `h3Distance` are closely related, but this naming doesn't reflect that.", so I was trying to find a naming that made that connection clear, but also allowed for the different potential path representations.

`cellPath` doesn't evoke the grid in my mind. As the fundamental unit of DGGS, that feels like it stands on its own. That is, `grid` implies `cell`, `edge`, and `path`, but the converse doesn't seem natural to me.

Specifically, `cellDistance` could be interpreted as the Haversine distance between cell centroids, but `gridDistance` eliminates that ambiguity for me.

I kind of like `gridDistance`, `cellPath`, `edgePath`, and `directedEdgePath`, but that loses the connection between distance and path, as you noted above.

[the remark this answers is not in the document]
ajfriend: I also like having related functions change faster in the right-most characters, which is what drew me to `___PathCell`, `___PathDirectedEdge`, `___PathEdge`. But that's more of a personal preference :)

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+## H3 Grid Functions
+
+| Current name |     Proposed name      |
+|--------------|------------------------|
+| `h3Distance` | `gridDistance`         |
+| `h3Line`     | `gridPathCell`         |
+| *DNE*        | `gridPathEdge`         |
+| *DNE*        | `gridPathDirectedEdge` |
```
dfellis: I'm curious what this one would do, since the only logical value that I can think of for a valid directed edge index is `1`.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,131 @@
...
+
+| Current name | Proposed name |     Notes      |
+|--------------|---------------|----------------|
+| `kRing`      | `disk`        | filled-in disk |
+| `hexRing`    | `ring`        | hollow ring    |
+
+
+todo: For the `kRing`, `hexRange`, `hexRing`, etc. family of functions, should we come up with some standard prefix or suffix to denote that the function will fail if it encounters a pentagon?
+
+I'm anticipating that, at least in the wrappers, we'd probably just expose users to the "works in all cases" version of the function.
```
dfellis: I don't think that's a valid assumption for the Java and Go bindings. Performance can matter on large datasets, especially if being done "online" in the execution time of a user's request.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,131 @@
...
+
+|  Current name  |    Proposed name     |                   Notes                    |
+|----------------|----------------------|--------------------------------------------|
+| `hexAreaKm2`   | `hexAreaAvgKm2`      | todo: `hexAreaMaxKm2` and `hexAreaMinKm2`? |
+| `hexAreaM2`    | `hexAreaAvgM2`       | todo: add min/max version?                 |
+| `edgeLengthKm` | `hexEdgeLengthAvgKm` | todo: add min/max version?                 |
+| `edgeLengthM`  | `hexEdgeLengthAvgM`  | todo: add min/max version?                 |
+| *DNE*          | `pentagonAreaAvgKm2` | plus others                                |
+
+todo: Could we add functions with signatures like `float cellAreaKm2(H3Index h)` to compute the actual area of a specific H3 cell?
```
dfellis: I think this is a great idea. I also don't think this blocks 4.0 release, though, so maybe it should be proposed in a separate document?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: I was thinking that `gridPathEdge` would return a collection of undirected edges connecting the passed-in origin and destination cells.

`gridPathDirectedEdge` would return a collection of directed edges connecting the origin and destination cells.

`gridPathCell` returns a collection of cells.

They're just three different ways of describing the same path between origin and destination cells, except for maybe the direction part.

dfellis: Ahh, that makes more sense. I saw it grouped with `gridDistance` and that affected my interpretation of what this would do.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: That's true. A better way to say that would be that we'd probably *direct* new users to the "always works" functions, and let the fast versions have a possibly more cumbersome name like `<func_name>_hexonly` or something.

The point is just to make it obvious to the end user when two functions do essentially the same thing, but one only works on hexagons.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Good point. Made an issue! #312

ajfriend:
@sahrk, I agree we should make our language precise. We definitely shouldn't be using cell as synonymous with `H3Index`.

It seems like 3 things are going on with the names:

1. Avoid naming collisions in C.
    - However, this seems to already be handled by `H3_EXPORT` if needed.
    - Thus, it sounds like we do not need to add any prefix to the functions for this purpose.
2. The abstract concept: cell vs hexagon vs pentagon vs edge
3. Representation of that concept: `H3Index`, lat/lng, FaceIJK, ...

The proposed change `h3IsValid` to `isValidCell`, for example, is to address ambiguity around point 2. It is unclear if it is testing validity for cells or for edges, for example. That said, point 3 is left unspecified (I guess I was thinking of it as being implied.).

Like you mentioned, it would be nice to have a single function name for each operation determined by the abstract concept, and have that function dispatch to different implementations based on the specific representation. Not so easy to do in C tho :(

I suppose we could use a prefix on function names to specify point 3. However, it seems like the majority of the functions are currently under the "H3 Index namespace", with just a few functions for converting to other formats.

Given that focus, would it make sense to usually leave out any prefix for point 3 and assume that `H3Index` is implied? When we do need a translation function, we could specify the type explicitly.

Do we expect to expand the library to have many more FaceIJK functions, for example, or do we expect that it will always focus on the `H3Index` representation?

Alternatively, could we do something to emulate a namespace for point 3?

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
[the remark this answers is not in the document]
ajfriend: Added some new ideas!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,130 @@
...
+    - an H3 cell which is a **topological** hexagon
+    - todo: advise use of the term "strict" for functions that can only handle hexagons?
+- pentagon:
+    - an H3 cell which is a **topological** pentagon
+- `DirectedEdge`
+    - an H3 index of `mode 2`
+    - representing a traversal from an origin cell to an adjacent destination cell
+- grid:
+    - the graph with nodes corresponding to H3 cells, and edges given by pairs of adjacent cells
+    - for example, `gridDistance` is the minimal number of edges in a graph path connecting two cells
```
ajfriend: @sahrk Would you use "grid" or "graph" to describe the topological connections between adjacent cells? Or is there another term that's better?

dfellis:
>     1. Avoid naming collisions in C.
>        
>        * However, this seems to already be handled by `H3_EXPORT` if needed.
>        * Thus, it sounds like we do not need to add any prefix to the functions for this purpose.

I personally don't agree with this statement. I don't think we ran into issues before when we were accidentally making almost every function unique with redundant `h3` text embedded in them. With the new naming strategy proposed, we could run into issues with people using the bindings (excluding h3-js since that's actually a transpilation) in their own real-world applications because of collisions that didn't happen before.

`H3_EXPORT` does solve it, but I think we need to explicitly call out that while the original C code can get away with `H3_EXPORT` doing nothing by default and anyone using it directly can add a prefix if they want, the bindings in 4.0.0 and up _should_ define such a prefix to make collisions unlikely for any users of the bindings, since they cannot easily set that prefix themselves.

Does that make my position on this point clear?

ajfriend:
@dfellis Got it. So we *should* use `H3_EXPORT` to apply a prefix like `h3_` to all functions and types by default? (Starting in 4.0)

dfellis:
Yes. I think so. It only matters for the bindings, not the original C library or h3-js, but it would likely be important for h3-py, h3-java, h3-go, and h3-nodr.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Notes
+
+We realize that "hexagon" will still be used informally to refer to the concept of "cell" (As the development team, we do it ourselves!).
+This should be expected in casual, informal discussions of H3.
+However, when **precision** is required, we advise the use of strict technical terms like "index", "cell", "hexagon", "pentagon", etc.
+In the codebase and in the documentation, strictly correct terminology should **always** be used.
+
```
nrabinowitz: Add section about `H3_EXPORT` / C naming collisions

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,130 @@
...
+## Use of "Topological"
+
+We should clarify in the documentation that H3 works with **topological** hexagons and pentagons, rather than true **geometric** hexagons and pentagons.
+
+See, for example: https://github.com/uber/h3-js/issues/53
+
+
+## General Function Names
+
```
ajfriend: `resolution` vs `getResolution`

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,130 @@
...
+| `h3IsPentagon`                | `isPentagon`          |                   |
+| `h3ToParent`                  | `cellToParent`        | or `getParent`?   |
+| `h3ToChildren`                | `cellToChildren`      | or `getChildren`? |
+| `h3IndexesAreNeighbors`       | `cellsAreNeighbors`   |                   |
+| `geoToH3`                     | `geoToCell`           |                   |
+| `h3ToGeo`                     | `cellToGeo`           |                   |
+| `numHexagons`                 | `numCells`            |                   |
+| `getRes0Indexes`              | `getRes0Cells`        |                   |
+| `getPentagonIndexes`          | `getPentagons`        |                   |
+| `h3GetBaseCell`               | `getBaseCellNumber`   |                   |
```
nrabinowitz: Possible rules of thumb for `get` prefix:
- Use for property getters (`getResolution`, `getBaseCellNumber`)
- Use for constant getters (`getNumCells`, `getRes0Cells`)
- But don't use for: transformations, aggregations, projections (`geoToCell`, etc)

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,130 @@
...
+## General Function Names
+
+|          Current name         |     Proposed name     |       Notes       |
+|-------------------------------|-----------------------|-------------------|
+| *Does Not Exist (DNE)*        | `isValidIndex`        |                   |
+| `h3IsValid`                   | `isValidCell`         |                   |
+| `h3UnidirectionalEdgeIsValid` | `isValidDirectedEdge` |                   |
+| `h3IsPentagon`                | `isPentagon`          |                   |
+| `h3ToParent`                  | `cellToParent`        | or `getParent`?   |
+| `h3ToChildren`                | `cellToChildren`      | or `getChildren`? |
```
nrabinowitz: Suggest `cellParent` and `cellChildren`?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Grid! It's right there in the name. Got it, got it. Cool. Moving along... :man_facepalming:

sahrk:
> Given that focus, would it make sense to usually leave out any prefix for point 3 and assume that `H3Index` is implied? When we do need a translation function, we could specify the type explicitly.

Yes, I think that the function names using `cell` are much cleaner for the external API, which I think is looking very pretty compared to the clunkiness of my original naming.

> Do we expect to expand the library to have many more FaceIJK functions, for example, or do we expect that it will always focus on the `H3Index` representation?

Nothing we're discussing precludes using different representations internally in the `H3` library to whatever degree is required. I was uncomfortable when the planar `IJ` coordinates were exposed in the public API, even though I agree it was the right decision from a utility standpoint. I think that using `cell` in the function names (and only using other representations for translation functions) helps make it clear that when a user transforms cells into a representation other than `H3Index` they are no longer in Kansas; they aren't working with "H3 cells" and have left the guarantees of the `H3` library behind and are on their own.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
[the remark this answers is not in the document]
sahrk: While I've always used `gridDistance` as you do here, maybe it would make sense to tie the distance directly to the rest of the terminology by using `cellPathDistance`.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
sahrk: > @sahrk Would you use "grid" or "graph" to describe the topological connections between adjacent cells? Or is there another term that's better?

I have always seen and used "grid".

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: I like it. Updating!

ajfriend:
What does everyone think about:

## Cell Neighborhood Functions

| Current name |    Proposed names    |     Notes      |
|--------------|----------------------|----------------|
| `kRing`      | `disk` or `gridDisk` | filled-in disk |
| `hexRing`    | `ring` or `gridRing` | hollow ring    |

ajfriend:
In terms of what @sahrk was calling out about different type representations of the same concept, and how it might cause problems down the road, I was thinking of one potential example.

@nrabinowitz, if you were to add some kind of vertex mode (which, i'm assuming, would represent **topological** vertices between cells), then there might be confusion around functions like:

- `h3ToGeoBoundary`
- `getH3UnidirectionalEdgeBoundary`

The ambiguity would be around wether the output would represent **topological** `H3Vertex` objects, or geographical lat/lng points (for which pentagons and hexagons might have more than 5 or 6 vertices).

Is there an obvious naming convention that would help clarify this? Or do we need to start thinking about specifying the types of outputs?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Cool. Case closed!

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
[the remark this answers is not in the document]
ajfriend: I'm now proposing:

| Current name |     Proposed name      |
|--------------|------------------------|
| `h3Distance` | `gridDistance`         |
| `h3Line`     | `gridPathCell`         |
| *DNE*        | `gridPathEdge`         |
| *DNE*        | `gridPathDirectedEdge` |

I think I'm happy with `gridDistance` instead of `cellPathDistance`, as the concept of "distance" is the same for the following three representations of the same path.

nrabinowitz:
> In terms of what @sahrk was calling out about different type representations of the same concept, and how it might cause problems down the road, I was thinking of one potential example.
> 
> @nrabinowitz, if you were to add some kind of vertex mode (which, i'm assuming, would represent **topological** vertices between cells), then there might be confusion around functions like:
> 
> * `h3ToGeoBoundary`
> * `getH3UnidirectionalEdgeBoundary`
> 
> The ambiguity would be around wether the output would represent **topological** `H3Vertex` objects, or geographical lat/lng points (for which pentagons and hexagons might have more than 5 or 6 vertices).
> 
> Is there an obvious naming convention that would help clarify this? Or do we need to start thinking about specifying the types of outputs?

I think it's fine to say that `Boundary` or `GeoBoundary` means lat/long points, and `Vertex` means the topological points. So we'd have e.g. `cellToGeoBoundary` and `cellToVertexes`. This is clearest if we use `Geo`/`GeoPoint`/`GeoCoord` and `Boundary`/`GeoBoundary` consistently (not sure if your current naming does so).

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,142 @@
...
+|-------------------------------|-----------------------|
+| *Does Not Exist (DNE)*        | `isValidIndex`        |
+| `h3IsValid`                   | `isValidCell`         |
+| `h3UnidirectionalEdgeIsValid` | `isValidDirectedEdge` |
+| `h3IsPentagon`                | `isPentagon`          |
+| `h3ToParent`                  | `cellParent`          |
+| `h3ToChildren`                | `cellChildren`        |
+| `h3IndexesAreNeighbors`       | `cellsAreNeighbors`   |
+| `geoToH3`                     | `geoToCell`           |
+| `h3ToGeo`                     | `cellToGeo`           |
```
nrabinowitz: Per my last comment, using `cellToGeo` and `geoToCell` implies that `geo` is our standard naming convention for lat/lon points. Just to confirm, is this what we want?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,142 @@
...
+| `numHexagons`                 | `numCells`            |
```
nrabinowitz: It seems like this should be `getNumCells`? We're not very consistent with `get`, but I don't see a reason why this function should not use it if the ones below do.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,142 @@
...
+
+|                  Current name                 |        Proposed name         |
+|-----------------------------------------------|------------------------------|
+| `getH3UnidirectionalEdge`                     | `getDirectedEdge`            |
+| `h3UnidirectionalEdgeIsValid`                 | `isValidDirectedEdge`        |
+| `getOriginH3IndexFromUnidirectionalEdge`      | `getDirectedEdgeOrigin`      |
+| `getDestinationH3IndexFromUnidirectionalEdge` | `getDirectedEdgeDestination` |
+| `getH3IndexesFromUnidirectionalEdge`          | `getDirectedEdgeCells`       |
+| `getH3UnidirectionalEdgesFromHexagon`         | `getDirectedEdgesFromCell`   |
+| `getH3UnidirectionalEdgeBoundary`             | `getDirectedEdgeBoundary`    |
```
nrabinowitz: As per above, `get` seems inconsistent here. Why `cellToGeoBoundary` but not `directedEdgeToGeoBoundary`? (Also, if we think `GeoBoundary` is too long, we should standardize on `Boundary`.)

Since we're spending so much time on naming, it would be good to have a guideline for when to use a `get-` prefix and when not to - e.g.
- Use `get` when retrieving constant data
- Use `get` when retrieving a subset of an object, or metadata about the object (e.g. `getEdgeOrigin`, `getCellResolution`)
- Use `to` when transforming between different representations of the same object, or when doing a lossy transformation to a new object

In this case, that might mean `cellToDirectedEdges` and `directedEdgeToCells` and possible `cellsToDirectedEdge`. Just spitballing to some extent, but it would be good to have a consistent line of reasoning for using `get` and `to`.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Area/Length Functions
+
+|  Current name  |    Proposed name     |                   Notes                    |
+|----------------|----------------------|--------------------------------------------|
+| `hexAreaKm2`   | `hexAreaAvgKm2`      | todo: `hexAreaMaxKm2` and `hexAreaMinKm2`? |
+| `hexAreaM2`    | `hexAreaAvgM2`       | todo: add min/max version?                 |
+| `edgeLengthKm` | `hexEdgeLengthAvgKm` | todo: add min/max version?                 |
+| `edgeLengthM`  | `hexEdgeLengthAvgM`  | todo: add min/max version?                 |
+| *DNE*          | `pentagonAreaAvgKm2` | plus others                                |
```
nrabinowitz: Per the reasoning above, these would all have `get` prefixes (as they're retrieving library constants).

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,142 @@
...
+| `h3ToChildren`                | `cellChildren`        |
+| `h3IndexesAreNeighbors`       | `cellsAreNeighbors`   |
+| `geoToH3`                     | `geoToCell`           |
+| `h3ToGeo`                     | `cellToGeo`           |
+| `numHexagons`                 | `numCells`            |
+| `getRes0Indexes`              | `getRes0Cells`        |
+| `getPentagonIndexes`          | `getPentagons`        |
+| `h3GetBaseCell`               | `getBaseCellNumber`   |
+
+### naming note
```
isaacbrodsky: nit: capitalize "Naming"

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+## H3 Grid Functions
+
+| Current name |     Proposed name      |
+|--------------|------------------------|
+| `h3Distance` | `gridDistance`         |
+| `h3Line`     | `gridPathCell`         |
+| *DNE*        | `gridPathEdge`         |
+| *DNE*        | `gridPathDirectedEdge` |
```
isaacbrodsky: It is a little difficult for me to mentally parse the name `gridPathCell`. Perhaps `cellPath` (as previously discussed), or maybe make the last word plural (`gridPathCells`, `gridPathEdges`, ...), would that better communicate it is a path made up of cells?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,142 @@
...
+| *DNE*          | `pentagonAreaAvgKm2` | plus others                                |
+
+todo: Could we add functions with signatures like `float cellAreaKm2(H3Index h)` to compute the actual area of a specific H3 cell?
+
+## Cell Neighborhood Functions
+
+| Current name |    Proposed names    |     Notes      |
+|--------------|----------------------|----------------|
+| `kRing`      | `disk` or `gridDisk` | filled-in disk |
+| `hexRing`    | `ring` or `gridRing` | hollow ring    |
```
isaacbrodsky: Note that `hexRing` is one of the hex-only functions, and I don't believe has a fallback in the C library.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: I like pluralizing.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: I'll add `geo` to the definitions up top. Are there any better alternatives to `geo`?

ajfriend: maybe `coord`?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Makes sense.

What about the `disk`, `ring`, and the distance and path functions? Do we want a `get` prefix for all of those, too?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
dfellis: I honestly like `coord` better since all of these things are geographic in nature.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,149 @@
...
+I'm anticipating that, at least in the wrappers, we'd probably just expose users to the "works in all cases" version of the function.
+
+
+## H3 Edge Types
+
+Instead of `UnidirectionalEdge`, use the term `DirectedEdge`.
+
+For a future undirected edge mode, use the term `Edge`.
+
+|                  Current name                 |    Proposed name (`get`)     |         Alternate (`to`)         |
```
isaacbrodsky: I find the `to` form to be clearer.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: For future reference, I think we all agreed on `point` in the meeting today.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: updating based on our meeting discussion today!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: For reference, we agreed that `get` was not needed for "computation" functions like `disk` and `ring`.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: updating based on our meeting discussion today!

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: adding!

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: I've added some new notes to the terminology section. Let me know if you think that covers it. I'll resolve this for now.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: adding

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: I think we've resolved this. Let me know if you see any issues with the latest version!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -118,19 +118,67 @@ There is some ambiguity between property, transform, and computation, so use you
...
+|--------------|---------------|----------|----------------|
+| `hexRanges`  | (remove)      | hex only | N x `hexRange` |
+
+
+#### Todo
+- maybe there's a better suffix than `_fast`/`_slow`?
+- what's going on with these distance allocations (or lack of)?
+
+```C
+// unnecessary malloc?
```
dfellis: Not unnecessary, if you look at the `kRingDistances` implementation it needs the "scratch space" to work.

ajfriend: Interesting. So `kRingDistances` depends on `hexRangeDistances` and `_kRingInternal`. From what I can tell, only `_kRingInternal` needs the "scratch space", but `hexRangeDistances` does not.

That means the `malloc` for `kRingDistances` is not needed for the "fast path" that only calls `hexRangeDistances`. If that's the majority of cases, should we re-factor this code to skip the `malloc`?

dfellis: if we push the `malloc` into `kRingDistances` or `_kRingInternal` but I'd need to look at the function signatures to see if that's even possible (is the size value available at that point in the call stack?)

ajfriend: Instead of pushing it in to those functions, we could pull it out into `kRing`: #319

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: noted!

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
```diff
@@ -0,0 +1,96 @@
...
+| `h3IsValid`                   | `isValidCell`         |                                         |
+| `h3UnidirectionalEdgeIsValid` | `isValidDirectedEdge` |                                         |
+| `h3IsPentagon`                | `isPentagon`          |                                         |
+| `h3ToParent`                  | `cellToParent`        |                                         |
+| `h3ToChildren`                | `cellToChildren`      |                                         |
+| `h3IndexesAreNeighbors`       | `cellsAreNeighbors`   |                                         |
+| `geoToH3`                     | `geoToCell`           |                                         |
+| `h3ToGeo`                     | `cellToGeo`           |                                         |
+| `kRing`                       | `disk`                | filled-in disk                          |
+| `hexRing`                     | `ring`                | hollow ring                             |
```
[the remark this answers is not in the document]
ajfriend: Agreed. That's reflected in the current version's table.

dev-docs/RFCs/v4.0.0/terms_for_concepts_and_datatypes.md
ajfriend: This should be reflected in the latest pass.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -11,74 +13,73 @@ This can cause confusion, for example, when a function might only work on strict
...
-| `compact`                     |                                                        |
-| `uncompact`                   |                                                        |
-| `polyfill`                    |                                                        |
+- use `get` prefix for
+    + constant data (`getNumCells`, `getRes0Cells`)
+    + object properties (`getResolution`, `getBaseCellNumber`)
+- use `to` to denote transforms
+    + different representations of the same object
+    + when doing a lossy transformation to a new object (`cellToParent`, `pointToCell`)
+- do not use `get` or `two` for *computations*
```
isaacbrodsky: typo: two -> to

dfellis: He's technically right, you should not use `two` as a prefix to computations ;)

ajfriend: haha, thanks

ajfriend: fixed.

ajfriend:
This RFC is almost done! I'll add my remaining questions below.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+|--------------|------------------------------------------|--------------------------------------------------|
+| `hexRange`   | `gridDiskUnsafe`                         | `hexRangeDistances`, does not allocate distances |
+| *DNE*        | `gridDiskSafe`                           |                                                  |
+| `kRing`      | `gridDisk`                               | `kRingDistances`, allocates and drops distances  |
+| `hexRanges`  | `gridDiskMultiUnsafe`, `gridDisksUnsafe` | N x `hexRange`                                   |
+
+- **TODO**: Do we *really* want to keep `hexRanges` in the API? It sounds like it currently has applications, but how
+  hard would it really be for a user to reproduce this functionality by just calling `hexRange` multiple times, and moving
+  the pointer before each call? It is also weird that we only provide the "unsafe" version of this function in the API.
+- **Idea**: leave it out initially, and see if anyone complains? It would be very easy to add back in.
```
ajfriend: Do we have consensus on what we want to do with `hexRanges`?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+| `h3UnidirectionalEdgeIsValid`                 | `isValidDirectedEdge`        |
+| `getH3UnidirectionalEdge`                     | `cellsToDirectedEdge`        |
+| `getH3IndexesFromUnidirectionalEdge`          | `directedEdgeToCells`        |
+| `getH3UnidirectionalEdgesFromHexagon`         | `originToDirectedEdges`      |
+| *DNE*                                         | `destinationToDirectedEdges` |
+| `getH3UnidirectionalEdgeBoundary`             | `directedEdgeToLine`         |
+| `getOriginH3IndexFromUnidirectionalEdge`      | `getDirectedEdgeOrigin`      |
+| `getDestinationH3IndexFromUnidirectionalEdge` | `getDirectedEdgeDestination` |
+
+todo: happy with `directedEdgeToLine`?
```
ajfriend: Are we happy with the name `directedEdgeToLine`? Is it clear that it would correspond to a GeoJSON `LineString`?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+|  Current name  |        Proposed name        |         Notes         |
+|----------------|-----------------------------|-----------------------|
+| `hexAreaKm2`   | `getHexagonAreaAvgKm2`      |                       |
+| `hexAreaM2`    | `getHexagonAreaAvgM2`       |                       |
+| `edgeLengthKm` | `getHexagonEdgeLengthAvgKm` |                       |
+| `edgeLengthM`  | `getHexagonEdgeLengthAvgM`  |                       |
+| *DNE*          | `getPentagonAreaAvg*`       |                       |
+| *DNE*          | `getPentagonEdgeLengthAvg*` |                       |
+| *DNE*          | `cellAreaKm2`               | area of specific cell |
+| *DNE*          | `cellAreaM2`                | area of specific cell |
```
ajfriend: Is everyone happy with these changes to the area/length functions?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
nrabinowitz: If we *do* keep it, I don't mind calling it `gridDisksUnsafe` and only exposing an unsafe version. But I'm on team "leave it out" - the small perf gain doesn't seem worth the added complexity in either the library or the bindings, and there's no particular reason (other than Uber's internal needs for one or two applications, which may no longer even apply) why we'd expose a batch function for k-ring but not for e.g. `pointToCell` or `isValidCell`

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+  hard would it really be for a user to reproduce this functionality by just calling `hexRange` multiple times, and moving
+  the pointer before each call? It is also weird that we only provide the "unsafe" version of this function in the API.
+- **Idea**: leave it out initially, and see if anyone complains? It would be very easy to add back in.
+
+#### Hollow Ring
+
+| Current name |  Proposed name   | Calls |
+|--------------|------------------|-------|
+| `hexRing`    | `gridRingUnsafe` | NONE  |
+| *DNE*        | `gridRingSafe`   |       |
```
nrabinowitz: Brute force approach for this - call `gridDiskDistances` and then filter out the inside rings

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
nrabinowitz: Would `Polyline` make more sense here? That might align a bit better with `Polygon`

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+| `h3ToParent`                  | `cellToParent`        |
+| `h3ToChildren`                | `cellToChildren`      |
+| `numHexagons`                 | `getNumCells`         |
+| `getRes0Indexes`              | `getRes0Cells`        |
+| `getPentagonIndexes`          | `getPentagons`        |
+| `h3GetBaseCell`               | `getBaseCellNumber`   |
+| `h3GetResolution`             | `getResolution`       |
+| `geoToH3`                     | `pointToCell`         |
+| `h3ToGeo`                     | `cellToPoint`         |
+| `h3ToGeoBoundary`             | `cellToPoly`          |
```
nrabinowitz: Would `cellToPolygon` be better?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
nrabinowitz: 👍

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: What does the `poly` mean in `Polyline`? If anything, this is a line segment and so might be called something like `UniLine`.

In GeoJSON, there's no Polyline, just `LineString`. If we did call it `Polyline`, I wonder if users would confuse it with the GeoJSON `MultiLineString`. Thoughts?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Agreed. I'll update.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Updated the text.

dfellis:
Btw @ajfriend before you merge this, remove the `[WIP]` from the title!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Ahh, I think I misunderstood what this function does. I was thinking it was representing a line as a pari of points from the lat/lng of the origin to the lat/lng of the destination.

Instead, it is representing the polygon edge(s) that are shared by the cells. (That's where the `Poly` comes from!)

Does this raise a question about how we might need to be careful about using the term "edge". There are "edges" of each polygon cell, and there are "edges" in the sense of the grid, connecting two adjacent cells. Depending on which one you're thinking of, the representations would be perpendicular.

ajfriend:
Will do!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+- **mode**:
+    - an integer describing the type of object being represented by an H3 index
+    - this integer is encoded in the `H3Index`
+- **cell** or **H3 cell**:
+    - a geometric polygon corresponding to an `H3Index` of `mode 1` (hexagon or pentagon)
+    - for functions that can handle either hexagons or pentagons, the more general term "cell" should be used whenever possible
+- **hexagon**:
+    - an H3 **cell** which is a **topological** hexagon
+    - below, we explain that functions which *only* work with **hexagons** have an `Unsafe` suffix;
+      these functions are paried with ones having a `Safe` suffix, meaning they can handle **pentagons**, but are slower
```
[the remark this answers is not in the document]
ajfriend: thanks!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+- **pentagon**:
+    - an H3 **cell** which is a **topological** pentagon
+- **directed edge**:
+    - represents a traversal from an origin **cell** to an adjacent destination **cell**
+    - corresponds to an `H3Index` of `mode 2`
+- **grid**:
+    - the graph with nodes corresponding to H3 cells, and edges given by pairs of adjacent cells
+    - for example, `gridDistance` is the minimal number of edges in a graph path connecting two cells
+- **point**:
+    - a representation of a geometric point in terms of a latitude/longitude pair
```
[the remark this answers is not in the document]
ajfriend: done!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,225 @@
...
+    - for functions that can handle either hexagons or pentagons, the more general term "cell" should be used whenever possible
+- **hexagon**:
+    - an H3 **cell** which is a **topological** hexagon
+    - below, we explain that functions which *only* work with **hexagons** have an `Unsafe` suffix;
+      these functions are paried with ones having a `Safe` suffix, meaning they can handle **pentagons**, but are slower
+- **pentagon**:
+    - an H3 **cell** which is a **topological** pentagon
+- **directed edge**:
+    - represents a traversal from an origin **cell** to an adjacent destination **cell**
+    - corresponds to an `H3Index` of `mode 2`
```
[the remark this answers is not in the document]
ajfriend: I'm not sure what you mean here. I updated the **cell** definition a bit with this wording. Is that what you meant, or were you referring to the edge definition?

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Also, this makes me wonder if we should reconsider the `Geo*` struct names to align with the changes

```
| `geoToH3`         | `pointToCell`   |
| `h3ToGeo`         | `cellToPoint`   |
| `h3ToGeoBoundary` | `cellToPolygon` |
```

That is,

- `GeoCoord`
- `GeoBoundary`
- `Geofence`
- `GeoPolygon`
- `GeoMultiPolygon`
- `LinkedGeoCoord`
- `LinkedGeoLoop`
- `LinkedGeoPolygon`
- `CoordIJ`

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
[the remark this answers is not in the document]
nrabinowitz: Oh, sorry, for diff layout reasons I read this as a definition for `cell`. Stet!

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
nrabinowitz: I think it would be a good idea to align the internal data structure names with the API. Should be a straightforward find/replace.

`Polyline` is used in some other contexts (e.g. SVG) to describe a line with multiple segments, which may be true for a directed edge if it has a distortion vertex.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: Agreed. Updated it to:

```
| Current name |  Proposed name   |              Calls               |
|--------------|------------------|----------------------------------|
| `hexRing`    | `gridRingUnsafe` | NONE                             |
| *DNE*        | `gridRingSafe`   | `gridDiskDistancesSafe`          |
| *DNE*        | `gridRing`       | `gridRingUnsafe`, `gridRingSafe` |
```

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: I took a stab at this with the latest diff. Let me know how it looks!

ajfriend:
Interested if anyone has thoughts on these changes:

```
| Current name |      Proposed name       |
|--------------|--------------------------|
| `compact`    | same or `compactCells`   |
| `uncompact`  | same or `uncompactCells` |
| `polyfill`   | same or `polygonToCells` |
| *DNE*        | `multiPolygonToCells`    |
```

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -0,0 +1,260 @@
...
+| `getRes0Indexes`              | `getRes0Cells`         |
+| `getPentagonIndexes`          | `getPentagons`         |
+| `h3GetBaseCell`               | `getBaseCellNumber`    |
+| `h3GetResolution`             | `getResolution`        |
+| `geoToH3`                     | `pointToCell`          |
+| `h3ToGeo`                     | `cellToPoint`          |
+| `compact`                     | same, `compactCells`   |
+| `uncompact`                   | same, `uncompactCells` |
+| *DNE*                         | `getMode`              |
+| `polyfill`                    | same, `polygonToCells` |
```
dfellis: We need to come to a final conclusion before merging this RFC on these three. Do we rename them with the `Cells` nomenclature, or leave them as is?

ajfriend: Resolved in the meeting today. Thanks!

ajfriend:
> Thanks for such great suggestions for new names! I would suggest cellRing & cellDisk instead of gridRing & gridDisk. Thanks!

Thanks @kashuk! We went back and forth between the `cell*` and `grid*` names previously.

One reason for going with `grid*` was to get it to line up with `gridDistance`. The other was to indicate to the user that these are really rings and disks that should be thought of in terms of the graph distances in the H3 cell grid/graph.

Issue #241, opened 2019-05-29
Add configurable memory management routines


coconut2015:
I briefly looked some of the h3lib code via web interface, and they use malloc / free.  Forgive me if h3lib can already configure to be used with other memory management routines.

I am wondering if it is possible to add configurable memory management routines other than malloc / free.  I could certainly modify the code myself, but then it would create an upgrade / merge issue.

Basically, I am interested in using H3 as embedded functions in databases.  PostgreSQL for instance would have its own set of routines.

https://www.postgresql.org/docs/current/spi-memory.html

Other databases such as DB2 that support C UDF could face the similar issue.

Thanks.

dfellis:
H3's API is designed such that the vast majority of memory usage is passed in by the consumer of the API. See for instance [maxH3ToChildrenSize](https://github.com/uber/h3/blob/master/src/h3lib/lib/h3Index.c#L158) and [h3ToChildren](https://github.com/uber/h3/blob/master/src/h3lib/lib/h3Index.c#L192), which are paired functions to first estimate how much memory you need to allocate, and then given a chunk of allocated memory, perform the operation.

The vast majority of the time when `malloc`/`free` is used, it's internal only and the memory is freed before the return value provided. The only exception I'm aware of is the [h3SetToLinkedGeo](https://github.com/uber/h3/blob/master/src/h3lib/lib/algos.c#L799) function, but cleanup logic is wrapped in the [destroyLinkedPolygon](https://github.com/uber/h3/blob/master/src/h3lib/lib/linkedGeo.c#L107) function for the end-user.

This API design decision is what has made binding into Java, Python, and Go relatively simple, with the Python bindings living in [basically a single file](https://github.com/uber/h3-py/blob/master/h3/h3.py) with low cyclomatic complexity.

Finally, I'll note that there are [multiple](https://github.com/bytesandbrains/h3-pg) [database](https://github.com/davidfauth/Neo4jH3) bindings already written for H3, so the minimal use of `malloc` and `free` has not been a blocker for database integrations. :)

coconut2015:
The reason why I am requesting the feature is because **any** use of malloc / free, even if for a brief window, can cause memory leak or corruption in MPP environments where threads can be suspended / killed at any given time (unless specific protection is made).  malloc / free also use synchronization primitives that may not be compatible with whatever main MPP program's thread / synchronization mechanisms.

[...]

PR #311, opened 2020-02-07
[WIP] RFC: Overrideable memory allocation

isaacbrodsky:
Draft RFC for customizing memory allocators. I'm stilling collecting reference materials, but these are the main approaches I am seeing so far. Tentatively, I'm suggesting to go with `#define`-based customization.

#241

### DIFF
--- a/dev-docs/RFCs/v4.0.0/overrideable-allocators-rfc.md
+++ b/dev-docs/RFCs/v4.0.0/overrideable-allocators-rfc.md
@@ -13,15 +13,31 @@ malloc implementation.
 
 This will address the following use cases:
 
-* H3 is used inside of another application which has its own heap managememnt scheme. For example, using the allocation
+* H3 is used inside of another application which has its own heap management scheme. For example, using the allocation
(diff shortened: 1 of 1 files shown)


ajfriend:
To understand the impact of this RFC, could we get a rough (if exhaustive is too exhausting) idea of what functions this would impact and/or how significant a change this is to the current approach?

Also, a description of the current approach?

dev-docs/RFCs/v4.0.0/overrideable-allocators-rfc.md
```diff
@@ -0,0 +1,125 @@
...
+## Overview
+
+This is a proposal for adding a mechanism for users of the H3 library to provide heap allocator instead of the default
+malloc implementation.
+
+## Motivation
+
+This will address the following use cases:
+
+* H3 is used inside of another application which has its own heap managememnt scheme. For example, using the allocation
```
nrabinowitz: Nit: `managememnt` typo

PR #313, opened 2020-02-17
Revive the faster polyfill algorithm

dfellis:
The tests all pass!

There is one fudge factor in the polyfill hexagon estimation function that I can mostly explain:

1. It's not used for the actual output set but for the polygon line tracing.
2. Sometimes the line tracing cuts through more hexagons than you expect because of the strange edge cases around odd-numbered hexagons crossing the icosahedron boundary.
3. I went with `12` to be safe (maximum number of points around an odd-numbered hexagon), but I think I could get away with just adding `5` as I think this issue only arises on polygons directly tracing odd-numbered pentagons, but I'm not sure how to prove that.

Finally, I fixed the issue with the "polyfill the world" test case by using a wild derivation of area based on a diagonal: https://math.stackexchange.com/questions/1921929/find-an-area-of-a-rectangle-by-its-diagonal-and-quotient-of-its-sides and knowing that the opposite corners of the bounding box *have* to be very far away unlike the side overlapping the transmeridian.

Also, the formatter inserted a *ton* of whitespace, and I'm not sure why.

### DIFF
--- a/src/h3lib/include/algos.h
+++ b/src/h3lib/include/algos.h
@@ -40,4 +40,12 @@ void h3SetToVertexGraph(const H3Index* h3Set, const int numHexes,
 // Create a LinkedGeoPolygon from a vertex graph
 void _vertexGraphToLinkedGeo(VertexGraph* graph, LinkedGeoPolygon* out);
 
+// Internal function for polyfill that traces a geofence with hexagons of a
+// specific size
+int _getEdgeHexagons(const Geofence* geofence, int numHexagons, int res,
+                     int* numSearchHexes, H3Index* search, H3Index* found);
+
+// The new polyfill algorithm. Separated out because it can theoretically fail
+int _polyfillInternal(const GeoPolygon* geoPolygon, int res, H3Index* out);
+
 #endif
--- a/src/h3lib/include/algos.h
+++ b/src/h3lib/include/algos.h
@@ -45,7 +45,7 @@ void _vertexGraphToLinkedGeo(VertexGraph* graph, LinkedGeoPolygon* out);
 int _getEdgeHexagons(const Geofence* geofence, int numHexagons, int res,
                      int* numSearchHexes, H3Index* search, H3Index* found);
 
-// The new polyfill algorithm. Separated out because it can theoretically fail
+// The polyfill algorithm. Separated out because it can theoretically fail
 int _polyfillInternal(const GeoPolygon* geoPolygon, int res, H3Index* out);
 
 #endif
--- a/src/h3lib/include/bbox.h
+++ b/src/h3lib/include/bbox.h
@@ -38,6 +38,8 @@ bool bboxIsTransmeridian(const BBox* bbox);
 void bboxCenter(const BBox* bbox, GeoCoord* center);
 bool bboxContains(const BBox* bbox, const GeoCoord* point);
 bool bboxEquals(const BBox* b1, const BBox* b2);
-int bboxHexRadius(const BBox* bbox, int res);
+int bboxHexEstimate(const BBox* bbox, int res);
+int lineHexEstimate(const GeoCoord* origin, const GeoCoord* destination,
+                    int res);
 
 #endif
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -46,6 +46,7 @@
 #define HEX_RANGE_K_SUBSEQUENCE 2
 #define MAX_ONE_RING_SIZE 7
 #define HEX_HASH_OVERFLOW -1
+#define POLYFILL_BUFFER 12
 
 /**
  * Directions used for traversing a hexagonal ring counterclockwise around
@@ -661,8 +662,13 @@ int H3_EXPORT(maxPolyfillSize)(const GeoPolygon* geoPolygon, int res) {
         totalVerts += geoPolygon->holes[i].numVerts;
     }
     if (numHexagons < totalVerts) numHexagons = totalVerts;
-    numHexagons += 12;  // TODO: Get a better constant here, but this seems
-                        // related to the rotation issue for odd hexes
+    numHexagons += POLYFILL_BUFFER;  // When the polygon is very small, near an
+                                     // icosahedron edge and is an odd
+                                     // resolution, the line tracing needs an
+                                     // extra buffer than the estimator function
+                                     // provides (but beefing that up to cover
+                                     // causes most situations to overallocate
+                                     // memory)
     return numHexagons;
 }
 
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -660,6 +660,8 @@ int H3_EXPORT(maxPolyfillSize)(const GeoPolygon* geoPolygon, int res) {
         totalVerts += geoPolygon->holes[i].numVerts;
     }
     if (numHexagons < totalVerts) numHexagons = totalVerts;
+    numHexagons += 12;  // TODO: Get a better constant here, but this seems
+                        // related to the rotation issue for odd hexes
     return numHexagons;
 }
 
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -44,6 +44,7 @@
 #define HEX_RANGE_SUCCESS 0
 #define HEX_RANGE_PENTAGON 1
 #define HEX_RANGE_K_SUBSEQUENCE 2
+#define MAX_ONE_RING_SIZE 7
 
 /**
  * Directions used for traversing a hexagonal ring counterclockwise around
@@ -638,9 +639,8 @@ int H3_EXPORT(hexRing)(H3Index origin, int k, H3Index* out) {
  * maxPolyfillSize returns the number of hexagons to allocate space for when
  * performing a polyfill on the given GeoJSON-like data structure.
  *
- * Currently a laughably padded response, being a k-ring that wholly contains
- * a bounding box of the GeoJSON, but still less wasted memory than initializing
- * a Python application? ;)
+ * The size is the maximum of either the number of points in the geofence or the
+ * number of hexagons in the bounding box of the geofence.
  *
  * @param geoPolygon A GeoJSON-like data structure indicating the poly to fill
  * @param res Hexagon resolution (0-15)
@@ -649,12 +649,18 @@ int H3_EXPORT(hexRing)(H3Index origin, int k, H3Index* out) {
 int H3_EXPORT(maxPolyfillSize)(const GeoPolygon* geoPolygon, int res) {
     // Get the bounding box for the GeoJSON-like struct
     BBox bbox;
-    bboxFromGeofence(&geoPolygon->geofence, &bbox);
-    int minK = bboxHexRadius(&bbox, res);
-
-    // The total number of hexagons to allocate can now be determined by
-    // the k-ring hex allocation helper function.
-    return H3_EXPORT(maxKringSize)(minK);
+    const Geofence geofence = geoPolygon->geofence;
+    bboxFromGeofence(&geofence, &bbox);
+    int numHexagons = bboxHexEstimate(&bbox, res);
+    // This algorithm assumes that the number of vertices is usually less than
+    // the number of hexagons, but when it's wrong, this will keep it from
+    // failing
+    int totalVerts = geofence.numVerts;
+    for (int i = 0; i < geoPolygon->numHoles; i++) {
+        totalVerts += geoPolygon->holes[i].numVerts;
+    }
+    if (numHexagons < totalVerts) numHexagons = totalVerts;
+    return numHexagons;
 }
 
 /**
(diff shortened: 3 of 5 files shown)


src/h3lib/lib/algos.c
```diff
@@ -634,28 +636,129 @@ int H3_EXPORT(hexRing)(H3Index origin, int k, H3Index* out) {
...
+                (destination.lon * j / numHexesEstimate);
+            H3Index pointHex = H3_EXPORT(geoToH3)(&interpolate, res);
+            // A simple hash to store the hexagon, or move to another place if
+            // needed
+            int loc = (int)(pointHex % numHexagons);
+            int loopCount = 0;
+            while (found[loc] != 0) {
+                // If this conditional is reached, the `found` memory block is
+                // too small for the given polygon. This should not happen.
+                if (loopCount > numHexagons) return -1;  // LCOV_EXCL_LINE
```
nrabinowitz: Can we use a named error code here?

src/h3lib/lib/algos.c
```diff
@@ -677,36 +780,139 @@ void H3_EXPORT(polyfill)(const GeoPolygon* geoPolygon, int res, H3Index* out) {
...
+
+    // 4. Begin main loop. While the search hash is not empty do the following
+    while (numSearchHexes > 0) {
+        // Iterate through all hexagons in the current search hash, then loop
+        // through all neighbors and test Point-in-Poly, if point-in-poly
+        // succeeds, add to out and found hashes if not already there.
+        int currentSearchNum = 0;
+        int i = 0;
+        while (currentSearchNum < numSearchHexes) {
+            H3Index ring[MAX_ONE_RING_SIZE] = {0};
```
nrabinowitz: For my info: Is there any benefit to declaring this outside the loop?

src/h3lib/lib/algos.c
```diff
@@ -677,36 +780,139 @@ void H3_EXPORT(polyfill)(const GeoPolygon* geoPolygon, int res, H3Index* out) {
...
+                    continue;  // Skip if this was a pentagon and only had 5
+                               // neighbors
+                }
+
+                H3Index hex = ring[j];
+
+                // A simple hash to store the hexagon, or move to another place
+                // if needed. This MUST be done before the point-in-poly check
+                // since that's far more expensive
+                int loc = (int)(hex % numHexagons);
```
nrabinowitz: This is very similar to the code in `_getEdgeHexagons` and elsewhere. I can see why you've repeated it, but is there any way to dry it out with a util like `int findSetIndex(H3Index* set, H3Index h3, int setSize)`? You'd return the index or error codes on dupe or failure, then do the `free` calls depending on error code.

src/h3lib/lib/algos.c
```diff
@@ -677,36 +780,139 @@ void H3_EXPORT(polyfill)(const GeoPolygon* geoPolygon, int res, H3Index* out) {
...
+                int loopCount = 0;
+                while (out[loc] != 0) {
+                    // If this branch is reached, we have exceeded the maximum
+                    // number of hexagons possible and need to clean up the
+                    // allocated memory.
+                    // LCOV_EXCL_START
+                    if (loopCount > numHexagons) {
```
nrabinowitz: What happens if the caller passes in memory smaller than `numHexagons`?

src/h3lib/lib/bbox.c
```diff
@@ -88,38 +90,58 @@ double _hexRadiusKm(H3Index h3Index) {
...
+int bboxHexEstimate(const BBox* bbox, int res) {
+    // Get the area of the pentagon as the maximally-distorted area possible
+    H3Index pentagons[12] = {0};
+    H3_EXPORT(getPentagonIndexes)(res, pentagons);
+    double pentagonRadiusKm = _hexRadiusKm(pentagons[0]);
+    // Area of a regular hexagon is 3/2*sqrt(3) * r * r
+    // The pentagon has the most distortion (smallest edges) and shares its
+    // edges with hexagons, so the most-distorted hexagons have this area
+    double pentagonAreaKm2 =
+        0.8 * (2.59807621135 * pentagonRadiusKm * pentagonRadiusKm);
```
nrabinowitz: Nit: It took me a sec to figure out why `0.8` is here, can you add to the comment?

src/h3lib/lib/bbox.c
```diff
@@ -88,38 +90,58 @@ double _hexRadiusKm(H3Index h3Index) {
...
-    GeoCoord vertex = {lat, bbox->east};
+    // Then get the area of the bounding box of the geofence in question
+    GeoCoord p1, p2;
+    p1.lat = bbox->north;
+    p1.lon = bbox->east;
+    p2.lat = bbox->south;
+    p2.lon = bbox->west;
+    double d = _geoDistKm(&p1, &p2);
+    double a = d * d / 2.5;  // Derived constant based on:
+                             // https://math.stackexchange.com/a/1921940
```
nrabinowitz: Sorry, I'm not following. How can you derive the constant without knowing the ratio? Isn't the whole point here that you have to know the aspect ratio of the bounding box (`fabs((p1.lon - p2.lon) / (p1.lat - p2.lat) )`) to calculate the area?

I also came here looking for what happens with 0-width or 0-height bounding boxes, though it doesn't seem like we'd treat them any differently because we don't take the real ratio?

src/apps/applib/include/args.h
```diff
@@ -23,6 +23,7 @@
 #include <inttypes.h>
 #include <stdbool.h>
 #include <stdio.h>
+
```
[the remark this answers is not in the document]
dfellis: `clang-format` did it.

src/h3lib/include/algos.h
```diff
@@ -40,4 +40,12 @@ void h3SetToVertexGraph(const H3Index* h3Set, const int numHexes,
 // Create a LinkedGeoPolygon from a vertex graph
 void _vertexGraphToLinkedGeo(VertexGraph* graph, LinkedGeoPolygon* out);
 
+// Internal function for polyfill that traces a geofence with hexagons of a
+// specific size
+int _getEdgeHexagons(const Geofence* geofence, int numHexagons, int res,
+                     int* numSearchHexes, H3Index* search, H3Index* found);
+
+// The new polyfill algorithm. Separated out because it can theoretically fail
```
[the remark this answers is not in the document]
dfellis: Will fix. BTW, almost everything was already written and approved the last time, but I agree here that it's silly.

src/h3lib/include/bbox.h
```diff
@@ -37,6 +38,8 @@ bool bboxIsTransmeridian(const BBox* bbox);
 void bboxCenter(const BBox* bbox, GeoCoord* center);
 bool bboxContains(const BBox* bbox, const GeoCoord* point);
 bool bboxEquals(const BBox* b1, const BBox* b2);
-int bboxHexRadius(const BBox* bbox, int res);
+int bboxHexEstimate(const BBox* bbox, int res);
+int lineHexEstimate(const GeoCoord* origin, const GeoCoord* destination,
```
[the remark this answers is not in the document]
dfellis: Probably not. Which file do you think it best fits in?

src/h3lib/lib/algos.c
```diff
@@ -634,28 +636,129 @@ int H3_EXPORT(hexRing)(H3Index origin, int k, H3Index* out) {
...
  * @param res The Hexagon resolution (0-15)
  * @param out The slab of zeroed memory to write to. Assumed to be big enough.
  */
 void H3_EXPORT(polyfill)(const GeoPolygon* geoPolygon, int res, H3Index* out) {
+    // TODO: Eliminate this wrapper with the H3 4.0.0 release
+    int failure = _polyfillInternal(geoPolygon, res, out);
+    // The polyfill algorithm can theoretically fail if the allocated memory is
+    // not large enough for the polygon, but this should be impossible given the
+    // conservative overestimation of the number of hexagons possible.
+    // LCOV_EXCL_START
```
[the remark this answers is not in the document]
dfellis: That's true. We can test this, then. I will write a new test case for it. :)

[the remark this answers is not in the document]
isaacbrodsky: @nrabinowitz The array does not have an associated length so `_polyfillInternal` would have no way to know to return an error (unless maybe `out` were null, in which case we can't zero it anyways)

src/h3lib/lib/algos.c
isaacbrodsky: I think this is from a comment on the previous PR. Declaring it in the loop puts the declaration and initialization in the same place. Declaring it outside the loop requires repeating the initialization code. I don't expect there to be a performance difference.

src/h3lib/lib/algos.c
isaacbrodsky: The code writes past the end of the array? Perhaps more ideally we would explicitly pass the array lengths in?

src/apps/applib/include/args.h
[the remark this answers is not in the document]
isaacbrodsky: @dfellis what version of clang-format do you have?

src/h3lib/lib/algos.c
nrabinowitz: I remember the discussion now - thanks. Stet.

src/h3lib/lib/bbox.c
dfellis: Oops! I accidentally committed the version with the hardwired ratio constant for a polygon for the entire mercator map. I will fix this.

src/apps/applib/include/args.h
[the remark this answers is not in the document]
dfellis: `clang-format version 9.0.0-2 (tags/RELEASE_900/final)`

src/h3lib/lib/algos.c
dfellis: Done

src/apps/applib/include/args.h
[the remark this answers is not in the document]
nrabinowitz: Hm. I have `7.0.0`. We should probably standardize, or we'll end up stepping on each other's toes - @isaacbrodsky, maybe there's a way to do this via CMake?

src/h3lib/include/algos.h
```diff
@@ -40,4 +40,12 @@ void h3SetToVertexGraph(const H3Index* h3Set, const int numHexes,
 // Create a LinkedGeoPolygon from a vertex graph
 void _vertexGraphToLinkedGeo(VertexGraph* graph, LinkedGeoPolygon* out);
 
+// Internal function for polyfill that traces a geofence with hexagons of a
+// specific size
+int _getEdgeHexagons(const Geofence* geofence, int numHexagons, int res,
+                     int* numSearchHexes, H3Index* search, H3Index* found);
+
+// The polyfill algorithm. Separated out because it can theoretically fail
+int _polyfillInternal(const GeoPolygon* geoPolygon, int res, H3Index* out);
```
nrabinowitz, line 49: If these are wholly internal, can we mark them `static` and not include in the header?

We had a conversation about this ages ago, but I don't remember what we settled on.

src/h3lib/include/bbox.h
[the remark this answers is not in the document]
nrabinowitz: If it's only used for the polyfill algo, maybe just an internal function in `algos`? Not a big deal either way.

src/h3lib/lib/algos.c
```diff
@@ -634,28 +637,130 @@ int H3_EXPORT(hexRing)(H3Index origin, int k, H3Index* out) {
...
+    int numHexagons = bboxHexEstimate(&bbox, res);
+    // This algorithm assumes that the number of vertices is usually less than
+    // the number of hexagons, but when it's wrong, this will keep it from
+    // failing
+    int totalVerts = geofence.numVerts;
+    for (int i = 0; i < geoPolygon->numHoles; i++) {
+        totalVerts += geoPolygon->holes[i].numVerts;
+    }
+    if (numHexagons < totalVerts) numHexagons = totalVerts;
+    numHexagons += 12;  // TODO: Get a better constant here, but this seems
```
nrabinowitz, line 651: Nit: Can we name this constant, and possibly add a comment about the cases in which it's needed?

src/apps/applib/include/args.h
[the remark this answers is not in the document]
dfellis: Can I consider this conversation dealt with from the decision in the last meeting?

src/h3lib/include/bbox.h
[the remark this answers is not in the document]
dfellis: Done.

src/h3lib/lib/algos.c
dfellis: Done.

src/h3lib/include/bbox.h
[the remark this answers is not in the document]
dfellis: Actually, now that I moved it. I recompiled and I got a warning about an implicit declaration of `_hexRadiusKm` defined in the `bbox.c` file being used. I'd like to keep this function in `bbox` unless you're okay with exposing `_hexRadiusKm`?

dfellis:
Ugh, why are there merge conflicts now?

dfellis:
Fixed the merge conflicts. It was the recent memory allocator changes.

src/h3lib/include/bbox.h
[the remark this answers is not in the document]
nrabinowitz: Stet

src/h3lib/lib/algos.c
```diff
@@ -649,28 +651,135 @@ int H3_EXPORT(hexRing)(H3Index origin, int k, H3Index* out) {
...
+    int numHexagons = bboxHexEstimate(&bbox, res);
+    // This algorithm assumes that the number of vertices is usually less than
+    // the number of hexagons, but when it's wrong, this will keep it from
+    // failing
+    int totalVerts = geofence.numVerts;
+    for (int i = 0; i < geoPolygon->numHoles; i++) {
+        totalVerts += geoPolygon->holes[i].numVerts;
+    }
+    if (numHexagons < totalVerts) numHexagons = totalVerts;
+    numHexagons += POLYFILL_BUFFER;  // When the polygon is very small, near an
```
nrabinowitz, line 665: Total nit: Move this to an above-line comment for legibility?

AbelVM:
Is that Travis stalled for 13 days? :thinking:

dfellis:
> Is that Travis stalled for 13 days? thinking

I had been waiting for it to return and then it never did and I forgot about it.

It appears [that it worked just fine](https://travis-ci.com/github/uber/h3/builds/169722835) but never updated Github with the results, and I don't see any way to re-run a successful build.

@nrabinowitz is there anything else this needs, or can we merge it?

nrabinowitz:
Merged

seanhandley:
@dfellis I'm updating the ruby bindings and I have this incongruity:

```
  1) H3.max_polyfill_size has the correct number of hexagons
     Failure/Error: expect(max_polyfill_size).to eq expected_count

       expected: 75367
            got: 47018

       (compared using ==)
     # ./spec/regions_spec.rb:50:in `block (3 levels) in <top (required)>'

Finished in 0.31789 seconds (files took 2.02 seconds to load)
121 examples, 1 failure

Failed examples:

rspec ./spec/regions_spec.rb:49 # H3.max_polyfill_size has the correct number of hexagons
```

Is this expected behaviour?

dfellis:
> @dfellis I'm updating the ruby bindings and I have this incongruity:
> 
> ```
>   1) H3.max_polyfill_size has the correct number of hexagons
>      Failure/Error: expect(max_polyfill_size).to eq expected_count
> 
>        expected: 75367
>             got: 47018
> 
>        (compared using ==)
>      # ./spec/regions_spec.rb:50:in `block (3 levels) in <top (required)>'
> 
> Finished in 0.31789 seconds (files took 2.02 seconds to load)
> 121 examples, 1 failure
> 
> Failed examples:
> 
> rspec ./spec/regions_spec.rb:49 # H3.max_polyfill_size has the correct number of hexagons
> ```
> 
> Is this expected behaviour?

Hi @seanhandley yes, that's expected behavior. The `max...` functions are setting an upper bound that is the amount of memory that should be allocated. The new polyfill algorithm has a more accurate estimator along with a faster algorithm (rather, the old algorithm required the max to be larger for it to work). Part of the speedup of the new algorithm is simply having less memory to allocate and fewer indices to test with the point-in-poly algorithm. :)

seanhandley:
Thanks @dfellis ! That clears that up 👍

PR #317, opened 2020-02-27
Define-based alloc

isaacbrodsky:
Work in progress branch for adding defined-based overriding of memory management functions, per #311. Main code change is adding a CMake option and changing source files to use a macro.

One of the advantages of overridable memory managers is the ability to test scenarios, that part has not been hooked up yet (need to run both overridden and non-overridden tests in the same test run to get coverage correct)


src/apps/testapps/testH3Memory.c
```diff
@@ -0,0 +1,75 @@
...
+H3Index sunnyvale = 0x89283470c27ffff;
+
+SUITE(h3Memory) {
+    TEST(t) {
+        int k = 9;
+        int hexCount = H3_EXPORT(maxKringSize)(k);
+
+        // Generate a set of hexagons to compact
+        H3Index* sunnyvaleExpanded = calloc(hexCount, sizeof(H3Index));
+        H3_EXPORT(kRing)(sunnyvale, k, sunnyvaleExpanded);
```
nrabinowitz: Other than inspecting the print logs, how does this test that the custom functions are being called?

One option would be to have global flags for each - reset them at the beginning of each test, then set them to true when the custom function is called.

src/apps/testapps/testH3Memory.c
```diff
@@ -0,0 +1,75 @@
...
+#include "test.h"
+#include "utility.h"
+
+#ifndef H3_ALLOC_PREFIX
+#error "This test can only be run with a prefix for alloc functions"
+#endif
+
+static int failAlloc = 0;
+
+void* H3_MEMORY(malloc)(size_t size) {
```
nrabinowitz: I think the test would be more explicit if you actually called these `custom_malloc` or what have you, then failed the test if `H3_ALLOC_PREFIX != "custom_"`

src/h3lib/include/alloc.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "h3api.h"
+
+#ifdef H3_ALLOC_PREFIX
+#define H3_MEMORY(name) TJOIN(H3_ALLOC_PREFIX, name)
+
+void* H3_MEMORY(malloc)(size_t size);
+void* H3_MEMORY(calloc)(size_t num, size_t size);
+void* H3_MEMORY(realloc)(void* ptr, size_t size);
+void H3_MEMORY(free)(void* ptr);
```
nrabinowitz: Is there a way to fail this at compile-time if any of these are not defined? Also, how does this work if the custom implementation has also declared these functions in a header file?

src/apps/testapps/testH3Memory.c
isaacbrodsky: Only via print logs - I intend to add an "expected number of calls" static variable that will be set in each test, after which the function returns `NULL` instead of forwarding the call. It would also be possible to assert on the number of calls.

src/h3lib/include/alloc.h
isaacbrodsky: I don't believe this can be failed at compile time, because the compiler doesn't know that - only the linker does. The replacement memory allocators might only be provided by a host application dynamically.

The custom implementation also declaring these functions is normal (same as sharing the same header file between multiple source files.)

src/apps/testapps/testH3Memory.c
isaacbrodsky: Updated to have a count of calls

src/apps/testapps/testH3Memory.c
isaacbrodsky: This turns out to be simpler since a separate copy of the library is needed (so that both code paths can be tested)

isaacbrodsky:
I found what was happening to cause coverage to unexpectedly drop for this change. Although lcov will merge coverage from the two different versions of H3 built here (with and without external memory management), this won't cause it to mark branches as hit. They are being considered separate branches. Unfortunately doesn't seem to be fixed in lcov v1.14, but does seem to be fixed in lcov master.

PR #321, opened 2020-03-17
Add Rust binding links

rustyconover:


nrabinowitz:
Thanks for contributing! Can you explain the difference between the two repositories, and possibly include this in either the links or the repo READMEs?

rustyconover:
@nrabinowitz Thank you.  I'm glad to help out.

The difference between the two repos is the Rust norm of having a `-sys` crate that has contains all of the unsafe Rust functions.  Unsafe functions are functions that call the C/C++ library directly.  The `libh3` crate contains safe Rust functions, these functions call into `libh3-sys` but users call them from safe Rust code and don't have to worry about handling all of the vagaries of calling into the C library.

For the difference between safe Rust and unsafe Rust please see:

https://doc.rust-lang.org/nomicon/meet-safe-and-unsafe.html

rustyconover:
I've also updated the module readme files.

nrabinowitz:
Great, looks good! Thanks for the explanation and README updates - possibly this is so idiomatic that Rust developers would find it obvious, but I think it's quite helpful.

PR #324, opened 2020-03-24
Add H3 React Native binding

realPrimoh:
Made H3-js compatible with React Native. This removes browser components as React Native does not use native browser variables like 'document.'

### DIFF
--- a/docs/community/bindings.md
+++ b/docs/community/bindings.md
@@ -34,6 +34,7 @@ As a C library, bindings can be made to call H3 functions from different program
 
 - [uber/h3-js](https://github.com/uber/h3-js)
 - [dfellis/h3-node](https://github.com/dfellis/h3-node)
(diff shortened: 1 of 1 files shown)


nrabinowitz:
Thanks for the contribution! This has been a request in the past, and I'm glad to see an option here. It looks like it might be hard to maintain, however - ideally we'd come up with a solution that can be automatically built from `h3-js` source. I'd love to hear more about what was required here - sounds like the main issues were removing references to `document` and replacing `fs` (which we shouldn't even require) with `react-native-fs`?

realPrimoh:
@nrabinowitz Yup! It was just using `react-native-fs` instead of `fs` and removing references to `document.` I agree that it'd be nice to have it built automatically from h3-js source. I can look into it more when I get a chance and feel free to make any necessary changes to the repo as you see fit.

PR #352, opened 2020-05-13
Build Debian packages for H3

isaacbrodsky:
This adds some CMake code for building Debian packages for H3. I did some quick tests locally and was able to build, install, and use H3 on amd64.

There is probably more work that would need to be done before these packages could be distributed. I'm not sure everything that needs to happen for that, so if someone with experience creating Debian packages is interested in helping, please reach out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,7 +200,7 @@ set(UNCONFIGURED_API_HEADER src/h3lib/include/h3api.h.in)
 set(CONFIGURED_API_HEADER src/h3lib/include/h3api.h)
 configure_file(${UNCONFIGURED_API_HEADER} ${CONFIGURED_API_HEADER})
 
-set(INSTALL_TARGETS h3)
+set(INSTALL_TARGETS)
 
 function(add_h3_library name h3_alloc_prefix_override)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -671,4 +683,27 @@ install(
...
     NAMESPACE "${namespace}"
     DESTINATION "${config_install_dir}"
+    COMPONENT libh3-dev
 )
+
+# Debian package build
+set(CPACK_DEB_COMPONENT_INSTALL 1)
+set(CPACK_GENERATOR "DEB")
+set(CPACK_DEBIAN_FILE_NAME "DEB-DEFAULT")
+# set(CPACK_DEBIAN_PACKAGE_MAINTAINER "TEST PACKAGE") # Required
```
dfellis, line 693: Why is the package maintainer not set?

isaacbrodsky: I wasn't 100% sure how to set this at the moment. I need to go through the Debian packaging documentation to set all the fields. It will probably be set to something like "Isaac Brodsky <debian@...>" or "H3 Open Source Team <h3-team@...>" depending on whether there's a public alias for the H3 maintainers as a whole.

PR #359, opened 2020-06-19
Return code RFC

isaacbrodsky:
Begin the RFC describing how H3 should indicate errors to the callers. Doesn't have an approach recommended yet.


nrabinowitz:
I just merged #364, which included a stub for this - you might want to either build on that stub or delete it, up to you. Sorry, I forgot I was going to remove that stub before merging.

isaacbrodsky:
The RFC was updated per offline review - I am going to merge this in order to make progress on submitting 4.x changes for review. Feedback and comments are of course still welcome.

PR #375, opened 2020-08-02
Documenting num_hexagon recurrence...

ajfriend:
...because I messed it up today.


ajfriend:
> Aren't we renaming this in 4.0?

We are! At that point, the "actually" comment is no longer needed. In the meantime, why not?

Issue #378, opened 2020-08-10
Establish H3 equivalent for `NaN`?


ajfriend:
Maybe somewhat random, but would it make sense to establish a convention for something equivalent to floating point `NaN`, but for H3. Something to indicate missing or invalid data?

This could be useful for when users are processing arrays of H3 indices, and need a way to represent missing data. For example, integer arrays are sometimes cast to float for this purpose, due to the lack of a `NaN` for integers.

This could be as simple as stating that `0` is the "NanH3", which is essentially what we already *implicitly* do in the library. It could be beneficial to establish this convention officially.

I'm not sure this would even require a code change. Maybe just stating this in the docs somewhere as the recommended convention. A code change might be as small as just establishing a constant in the C code.

Thoughts? I don't have a killer use case for this yet, but I thought there was a chance it might be useful in the future.

isaacbrodsky:
We have a constant `H3_INVALID_INDEX` for this purpose:

```
src/h3lib/include/h3Index.h:#define H3_INVALID_INDEX 0
```

We should add to the documentation that unlike other modes, the API specifies the index `0` will not be valid in the future.

ajfriend:
Ah perfect!

However, I'm not sure I understand what you mean here:

> unlike other modes, the API specifies the index `0` will not be valid in the future.

dfellis:
> Ah perfect!
> 
> However, I'm not sure I understand what you mean here:
> 
> > unlike other modes, the API specifies the index `0` will not be valid in the future.

The Index format supports multiple modes, but we currently only use two. We have plans for a third, so that collection of indexes will then become valid. But a full zero set of 64 zeroes will never be a valid H3 Index value (since it's not valid under mode zero, it's impossible for that to change).

ajfriend:
Got it. I do wonder if there's a slight difference between "being invalid" (which, as y'all have sated, may change in the future) and the "NaN of H3".

Uniqueness would be the big one, probably. We wouldn't want folks having the freedom to use just *any* invalid index for NanH3.

The NanH3 might indicate a **little more meaning**: missing data and/or a potential return value for a nonsensically-called function (but error codes might handle that better). Just a regular invalid index would have **a bit less meaning**: just gibberish, and would probably indicate a bug somewhere.

If you buy those arguments, then `H3_INVALID_INDEX` might not be as descriptive of a name as something like `H3_NAN` or `H3_NAI` (not an index?).

dfellis:
[...]

PR #379, opened 2020-08-14
change H3_INVALID_INDEX to H3_NULL

ajfriend:
Following this issue: https://github.com/uber/h3/issues/378

Also, update the build instructions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,8 @@
 /bin/
 /examples/bin/
 /lib/
-
+# Travis CI build directory
+/build/
 # Local build directories
 /Debug/
 /Release/
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,7 @@
 /bin/
 /examples/bin/
 /lib/
-# Travis CI build directory
-/build/
+
 # Local build directories
 /Debug/
 /Release/
--- a/README.md
+++ b/README.md
@@ -76,6 +76,14 @@ You can install system-wide with:
 sudo make install
 ```
 
+To clean the directory of any build artifacts and return it to its initial checked out state, run:
+
+```
+git clean -ffdx
+```
+
+**Warning**: This will remove **all** files not commited to the repo. Make sure you've committed your work!
+
 #### Testing
 
 After making the project, you can test with `make test`, and if `lcov` is installed you can `make coverage` to generate a code coverage report.
--- a/README.md
+++ b/README.md
@@ -72,9 +72,11 @@ cmake ..
 make
 ```
 
+All subsequent `make` commands should be run from within the `build` directory.
+
 **Note**: There are several ways to build H3 with CMake; the method above is just one example that restricts all build artifacts to the `build` directory.
 
-You can install system-wide (from the `build` directory) with:
+You can install system-wide with:
 
 ```
 sudo make install
--- a/README.md
+++ b/README.md
@@ -63,26 +63,28 @@ You will need to install CMake and Visual Studio, including the Visual C++ compi
 
 #### Compilation
 
-From the repository you would then compile like so:
+From the repository root, you can compile H3 with:
 
 ```
-cmake .
+mkdir build
+cd build
+cmake ..
 make
 ```
 
-You can install system-wide with:
+**Note**: There are several ways to build H3 with CMake; the method above is just one example that restricts all build artifacts to the `build` directory.
+
+You can install system-wide (from the `build` directory) with:
 
 ```
 sudo make install
 ```
 
-To clean the directory of any build artifacts and return it to its initial checked out state, run:
+If using the method above, from the repository root, you can clean all build artifacts with:
 
 ```
-git clean -ffdx
+rm -rf build
 ```
-
-**Warning**: This will remove **all** files not commited to the repo. Make sure you've committed your work!
 
 #### Testing
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Documentation is available at [https://uber.github.io/h3/](https://uber.github.i
 
 We recommend using prebuilt bindings if they are available for your programming language. Bindings for [Go](https://github.com/uber/h3-go), [Java](https://github.com/uber/h3-java), [JavaScript](https://github.com/uber/h3-js), [Python](https://github.com/uber/h3-py), and [others](https://uber.github.io/h3/#/documentation/community/bindings) are available.
 
-On macOS, you can install H3 using brew:
+On macOS, you can install H3 using `brew`:
 ```
 brew install h3
 ```
@@ -46,7 +46,7 @@ sudo apt install cmake make gcc libtool
 sudo apt install clang-format cmake-curses-gui lcov doxygen
 ```
 
-* macOS (using brew)
+* macOS (using `brew`)
 
 First make sure you [have the developer tools installed](http://osxdaily.com/2014/02/12/install-command-line-tools-mac-os-x/) and then
 
--- a/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
+++ b/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
@@ -66,6 +66,10 @@ The following technical terms should be used in the documentation, the H3 codeba
     - may include more than 6 points in the case that a cell is not a geometric hexagon, such as when a hexagon crosses an icosahedron boundary
     - may also be used to describe the boundary between two geometric cells, as in the case of an edge
     - represented in the H3 codebase with the `CellBoundary` struct (previously `GeoBoundary` before v4.0)
+- `H3_NULL`;
+    - equivalent to `0` and guaranteed to never be a valid `H3Index` (even after any future H3 **modes** are added)
+    - returned by functions to denote an error, or to denote missing data in arrays of `H3Index`
+    - analogous to `NaN` in floating point
 
 
 ### Use of "hex", "hexagon", "cell", "pentagon", etc.
--- a/docs/core-library/h3indexing.md
+++ b/docs/core-library/h3indexing.md
@@ -24,7 +24,7 @@ The **H3Index** is the integer representation of an **H3** index, which can be p
 Mode 0 contains a special index, `H3_NULL`, which is unique: it is bit-equivalent to `0`.
 This index indicates, *specifically*, an invalid, missing, or uninitialized **H3** index;
 it is analogous to `NaN` in floating point.
-Going forward, it should be used instead of an arbitrary Mode 0 index, due to its uniqueness and easy identifiability.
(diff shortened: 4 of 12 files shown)


ajfriend:
Should we add `#define H3_NULL 0` to `h3api.h.in`? Remove it from `h3index.h`?

src/apps/applib/lib/utility.c
```diff
@@ -242,7 +242,8 @@ void randomGeo(GeoCoord* g) {
 int countActualHexagons(H3Index* hexagons, int numHexagons) {
     int actualNumHexagons = 0;
     for (int i = 0; i < numHexagons; i++) {
-        if (hexagons[i] != H3_INVALID_INDEX) {
+        // should we actually use `h3IsValid` here?
```
dfellis, line 245: In a tight loop that would be expensive. Since this function is not exported, we can know *all* callsites and be sure that it only returns valid indexes or zero.

README.md
```diff
@@ -76,6 +76,14 @@ You can install system-wide with:
 sudo make install
 ```
 
+To clean the directory of any build artifacts and return it to its initial checked out state, run:
```
isaacbrodsky, line 79: I’d suggest advising people to use a separate build directory instead. This is fairly broad advise for git itself

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -66,6 +66,10 @@ The following technical terms should be used in the documentation, the H3 codeba
     - may include more than 6 points in the case that a cell is not a geometric hexagon, such as when a hexagon crosses an icosahedron boundary
     - may also be used to describe the boundary between two geometric cells, as in the case of an edge
     - represented in the H3 codebase with the `CellBoundary` struct (previously `GeoBoundary` before v4.0)
+- `H3_NULL`;
+    - equivalent to `0` and guaranteed to never be a valid `H3Index` (even after any future H3 **modes** are added)
```
isaacbrodsky, line 70: I’d suggest adding this in docs- the RFCs are more internal technical documentation and aren’t on the website

README.md
ajfriend: I'm happy to make that change (especially for my own workflow!), but we should update the recommended build instructions in the `README.md` then, yeah?

We currently have:

```sh
cmake .
make
```

What should we change that to? @isaacbrodsky

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: It wasn't clear to me the best place to add this in the docs. Any ideas?

ajfriend:
> Sure? I'm not 100% sure this is clearer than the previous name, but I don't object to the change.

The summary might be: There are many possible invalid H3 indexes, but `H3_NULL` is specific and unique.

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
isaacbrodsky: Probably around the description of the H3 Index format in `docs/core-library/h3indexing.md`?

README.md
isaacbrodsky: Any of these are roughly the same (slightly different if using cmake to start the build program vs. running `make` directly)

```
cmake -Bbuild
cmake --build build
```

or

```
cmake -Bbuild
cd build
make
```

or

```
mkdir build
cd build
cmake ..
make
```

ajfriend: I like the last one!

.gitignore
```diff
@@ -8,8 +8,7 @@
 /bin/
 /examples/bin/
 /lib/
-# Travis CI build directory
-/build/
+
```
ajfriend, line 11: Removing this in light of the new build instructions, as this "ignore" makes it less obvious to a user that they have a build directory floating around.

Happy to change it back if folks disagree, but I find it helpful when `git status` shows me what new files have been created by the build.

src/apps/applib/lib/utility.c
ajfriend: got it. removed the comment!

README.md
ajfriend: updated!

ajfriend:
> Should we add `#define H3_NULL 0` to `h3api.h.in`? Remove it from `h3index.h`?

I guess we can stay conservative in this PR and not make these changes. Can always do it in the future if we think it would be valuable.

ajfriend:
Oh, just saw [this](https://github.com/uber/h3/blame/master/docs/core-library/h3indexing.md#L16):

```
Mode 0 is reserved and indicates an invalid **H3** index.
```

Do we want to reserve a *whole mode* for "intentionally invalid" indices? That would make them non-unique, where uniqueness seems like the main benefit of defining `H3_NULL`.

If we go that route, should we revisit this constant, which is used in a few places:

https://github.com/uber/h3/blob/fd0f8af156df2f98649304ebbc080e974f708644/src/h3lib/include/h3Index.h#L85-L86

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
ajfriend: I can add that in, after we figure out what to do with the `Mode 0` issue below.

dfellis:
> Oh, just saw [this](https://github.com/uber/h3/blame/master/docs/core-library/h3indexing.md#L16):
> 
> ```
> Mode 0 is reserved and indicates an invalid **H3** index.
> ```
> 
> Do we want to reserve a _whole mode_ for "intentionally invalid" indices? That would make them non-unique, where uniqueness seems like the main benefit of defining `H3_NULL`.
> 
> If we go that route, should we revisit this constant, which is used in a few places:
> 
> https://github.com/uber/h3/blob/fd0f8af156df2f98649304ebbc080e974f708644/src/h3lib/include/h3Index.h#L85-L86

That was for distinguishing the H3 indexes that this OSS project produces versus the original version of H3 when we were first developing it and needed to make a breaking change to the bit format. There were systems that had mixed sets of these indices for a time.

ajfriend:
> That was for distinguishing the H3 indexes that this OSS project produces versus the original version of H3 when we were first developing it and needed to make a breaking change to the bit format. There were systems that had mixed sets of these indices for a time.

At this point, would it make sense to remove it?

dfellis:
> > That was for distinguishing the H3 indexes that this OSS project produces versus the original version of H3 when we were first developing it and needed to make a breaking change to the bit format. There were systems that had mixed sets of these indices for a time.
> 
> At this point, would it make sense to remove it?

I would argue no. Changes to the bit format of H3 are inherently backwards incompatible, so while it does make 1/16th of the address space inaccessible, adjusting that could cause other problems, particularly with a mixed set of systems that may be operating with different versions of the H3 library.

ajfriend:
> > At this point, would it make sense to remove it?
> 
> I would argue no. Changes to the bit format of H3 are inherently backwards incompatible, so while it does make 1/16th of the address space inaccessible, adjusting that could cause other problems, particularly with a mixed set of systems that may be operating with different versions of the H3 library.

That makes sense for Mode 0 in general. I've updated the description in the diff above. Let me know if that jives with everyone.

Specifically, regarding `H3_INIT`, would it make sense to replace it with `H3_NULL`, and remove `H3_INIT`?

isaacbrodsky:
> Specifically, regarding `H3_INIT`, would it make sense to replace it with `H3_NULL`, and remove `H3_INIT`?

`H3_INIT` seems to be fulfilling a different need, which is for an index initialized with all indexing digits set to 7.

.gitignore
isaacbrodsky: I disagree with this change, this seems to be opposite the usual Git workflow of considering build files to not be tracked by version control. Most of the paths listed here are specifically for the build process.

docs/core-library/h3indexing.md
```diff
@@ -13,7 +13,18 @@ Child hexagons are linearly smaller than their parent hexagons.
...
+* Mode 1 is an **H3** Cell (Hexagon/Pentagon) index.
+* Mode 2 is an **H3** Unidirectional Edge (Cell A -> Cell B) index.
+* Mode 3 is planned to be a bidirectional edge (Cell A <-> Cell B).
+* Mode 0 is reserved and indicates an invalid **H3** index.
+  * This mode remains, partially, for backwards compatibility with an older version of H3.
+
+Mode 0 contains a special index, `H3_NULL`, which is unique: it is bit-equivalent to `0`.
+This index indicates, *specifically*, an invalid, missing, or uninitialized **H3** index;
+it is analogous to `NaN` in floating point.
+Going forward, it should be used instead of an arbitrary Mode 0 index, due to its uniqueness and easy identifiability.
```
isaacbrodsky, line 27: Remove "Going forward," - suggests that other values were used to indicate invalid indexes which I don't believe was the case.

dfellis:
https://github.com/uber/h3/pull/379#discussion_r472251275

True, but since Travis CI is going away, it won't be generated in the first place?

isaacbrodsky:
> [#379 (comment)](https://github.com/uber/h3/pull/379#discussion_r472251275)
> 
> True, but since Travis CI is going away, it won't be generated in the first place?

This will still be the name of the build directory in Github Actions. It would also be created as part of the README instructions.

dfellis:
> > [#379 (comment)](https://github.com/uber/h3/pull/379#discussion_r472251275)
> > True, but since Travis CI is going away, it won't be generated in the first place?
> 
> This will still be the name of the build directory in Github Actions. It would also be created as part of the README instructions.

Ahh, then, yes @ajfriend can we undo the changes to `.gitignore`?

.gitignore
ajfriend: That makes sense. I'll change it back.

docs/core-library/h3indexing.md
ajfriend: OK, I can remove that. My understanding of Mode 0 was that anything in that mode could be used to indicate an invalid index, which I presume did happen, and that's why we've kept this "legacy" mode.

src/h3lib/include/h3Index.h
```diff
@@ -82,7 +82,11 @@
 /** 0's in the 7 base cell bits, 1's everywhere else. */
 #define H3_DIGIT_MASK_NEGATIVE (~H3_DIGIT_MASK)
 
-/** H3 index with mode 0, res 0, base cell 0, and 7 for all index digits. */
+/**
+ * H3 index with mode 0, res 0, base cell 0, and 7 for all index digits.
+ * Typically used to initialize the creation of an H3 cell index, which
+ * expects all direction digits to be 7 beyond the cell's resolution.
+ */
```
ajfriend, line 89: Make sense?

PR #380, opened 2020-08-14
Add exhaustive tests and benchmark for getH3UnidirectionalEdgeBoundary

nrabinowitz:
This is a precursor to refactoring `getH3UnidirectionalEdgeBoundary` to an approach not based on comparing geo coordinates, which should be faster and will fix correctness at fine resolutions. But first, tests!

- Adds exhaustive tests for `getH3UnidirectionalEdgeBoundary` at res 0, 1, 2, and 3.
  - For every cell, get all edges
  - For every edge, assert that the boundary equals the reversed boundary of the same edge in the other direction

- Adds a benchmark for `getH3UnidirectionalEdgeBoundary`, current output on my machine:
```
-- getH3UnidirectionalEdgeBoundary: 53.851000 microseconds per iteration (10000 iterations)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,7 +167,7 @@ set(OTHER_SOURCE_FILES
     src/apps/testapps/testVec2d.c
     src/apps/testapps/testVec3d.c
     src/apps/testapps/testH3UniEdge.c
-    src/apps/testapps/testH3uniEdgeExhaustive.c
+    src/apps/testapps/testH3UniEdgeExhaustive.c
     src/apps/testapps/testLinkedGeo.c
     src/apps/testapps/mkRandGeo.c
     src/apps/testapps/testH3Api.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,6 +192,7 @@ set(OTHER_SOURCE_FILES
     src/apps/benchmarks/benchmarkH3SetToLinkedGeo.c
     src/apps/benchmarks/benchmarkKRing.c
     src/apps/benchmarks/benchmarkH3Line.c
+    src/apps/benchmarks/benchmarkH3UniEdge.c
     src/apps/benchmarks/benchmarkH3Api.c)
 
 set(ALL_SOURCE_FILES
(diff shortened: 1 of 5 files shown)


src/apps/testapps/testH3UniEdgeExhaustive.c
```diff
@@ -0,0 +1,98 @@
...
+
+    for (int i = 0; i < 6; i++) {
+        if (isPentagon && i == 0) {
+            t_assert(edges[i] == H3_INVALID_INDEX,
+                     "last pentagon edge is empty");
+            continue;
+        }
+        t_assert(H3_EXPORT(h3UnidirectionalEdgeIsValid)(edges[i]) == 1,
+                 "edge is an edge");
+        destination =
```
ajfriend, line 47: Check that the origin is `h3`?

CMakeLists.txt
```diff
@@ -167,6 +167,7 @@ set(OTHER_SOURCE_FILES
     src/apps/testapps/testVec2d.c
     src/apps/testapps/testVec3d.c
     src/apps/testapps/testH3UniEdge.c
+    src/apps/testapps/testH3uniEdgeExhaustive.c
```
isaacbrodsky: Inconsistent case might cause an issue here

src/apps/testapps/testH3UniEdgeExhaustive.c
nrabinowitz: Added assertion.

src/apps/benchmarks/benchmarkH3UniEdge.c
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "benchmark.h"
+#include "geoCoord.h"
+#include "h3api.h"
+
+// Fixtures
+H3Index edges[6] = {0};
+H3Index hex = 0x89283080ddbffff;
```
dfellis, line 22: Is there a particular significance to this exactly hex ID, or was it chosen arbitrarily? It would be useful to have a comment on this in the code either way.

CMakeLists.txt
nrabinowitz: Fixed, thanks

src/apps/benchmarks/benchmarkH3UniEdge.c
nrabinowitz: It's arbitrary, copy/pasta from another benchmark. I can add a comment to both.

nrabinowitz: Appears to be Pier 39: https://observablehq.com/@nrabinowitz/h3-index-inspector#89283080ddbffff

dfellis: Interesting. But good to know there was no explicit significance to it.

PR #360, opened 2020-06-19
Enable some optimization flags in Release

isaacbrodsky:
I wasn't able to enable a few of the flags from that PR due to compiler differences and the flags not being available in the compiler I tested with. I turned on interprocedural optimization by default, and added `-O2 -finline-functions`.

From my quick test, interprocedural optimization provides a small benefit in most benchmarks, and a more noticeable benefit in benchmarkKRing. The other compiler flags (`-O2`...) seemed to have a fairly small effect.

Before:
```
$ ./bin/benchmarkH3Api
	-- geoToH3: 0.697000 microseconds per iteration (10000 iterations)
	-- h3ToGeo: 0.406500 microseconds per iteration (10000 iterations)
	-- h3ToGeoBoundary: 2.181400 microseconds per iteration (10000 iterations)
$ ./bin/benchmarkH3Line
	-- h3LineNear: 14.893200 microseconds per iteration (10000 iterations)
	-- h3LineFar: 660.081000 microseconds per iteration (1000 iterations)
$ ./bin/benchmarkH3SetToLinkedGeo
	-- h3SetToLinkedGeoRing2: 62.637000 microseconds per iteration (10000 iterations)
	-- h3SetToLinkedGeoDonut: 23.926700 microseconds per iteration (10000 iterations)
	-- h3SetToLinkedGeoNestedDonuts: 97.380100 microseconds per iteration (10000 iterations)
$ ./bin/benchmarkKRing
	-- kRing10: 8.973500 microseconds per iteration (10000 iterations)
	-- kRing20: 31.355800 microseconds per iteration (10000 iterations)
	-- kRing30: 63.647000 microseconds per iteration (10000 iterations)
	-- kRing40: 108.356200 microseconds per iteration (10000 iterations)
	-- kRingPentagon10: 203.794000 microseconds per iteration (500 iterations)
	-- kRingPentagon20: 1782.148000 microseconds per iteration (500 iterations)
	-- kRingPentagon30: 5789.460000 microseconds per iteration (50 iterations)
	-- kRingPentagon40: 13681.900000 microseconds per iteration (10 iterations)
$ ./bin/benchmarkPolyfill
	-- polyfillSF: 1237.274000 microseconds per iteration (500 iterations)
	-- polyfillAlameda: 1704.760000 microseconds per iteration (500 iterations)
	-- polyfillSouthernExpansion: 55909.100000 microseconds per iteration (10 iterations)
$ ./bin/benchmarkPolygon
[...]

PR #326, opened 2020-03-26
Enable interprocedural optimization

Komzpa:
without inlining:
```
kom@nucat:~/proj/h3/build$ bin/benchmarkH3Api 
        -- geoToH3: 1.187343 microseconds per iteration (10000 iterations)
        -- h3ToGeo: 0.796665 microseconds per iteration (10000 iterations)
        -- h3ToGeoBoundary: 2.219408 microseconds per iteration (10000 iterations)
```
with inlining:
```
kom@nucat:~/proj/h3/build$ bin/benchmarkH3Api
        -- geoToH3: 0.447104 microseconds per iteration (10000 iterations)
        -- h3ToGeo: 0.265080 microseconds per iteration (10000 iterations)
        -- h3ToGeoBoundary: 1.341663 microseconds per iteration (10000 iterations)
```

Addresses #325.

PR #381, opened 2020-08-17
Begin converting to GitHub Actions

isaacbrodsky:
This begins moving us over to GitHub Actions for CI. We're able to remove a few of the Travis CI tests that are now run in GitHub Actions.

GitHub Actions should make it easier to test pull requests that work with compiler flags (#372, #360, #326) since the compiler installs seem to be more "standard" compared to Travis CI, which does more custom management of the compiler installs, which can result in standard "extras" missing. Also, Github Actions seems to be faster than Travis CI or Appveyor.

Not all tests are ported, this will be addressed in a later PR. The following doesn't work:
* FOSSA/Coveralls reports are not switched over since the API keys haven't been installed yet
* Didn't determine how to test on ARM Linux.
* Badges aren't updated yet


CMakeLists.txt
```diff
@@ -35,7 +35,8 @@ option(BUILD_FILTERS "Build filter applications." ON)
 option(BUILD_GENERATORS "Build code generation applications." ON)
 
 if(WIN32)
-    set(SHELL PowerShell -Command)
+    # Use bash (usually from Git for Windows) for piping results
+    set(SHELL bash -c)
```
dfellis, line 39: Do we know of any Windows users that can comment on this?

I know that I have bash on my Windows VM, but is that true for those who treat it as a daily driver and use git?

PR #384, opened 2020-08-20
Fix bug in pentagon-to-neighbor getH3UnidirectionalEdge

nrabinowitz:
We had a bug in `getH3UnidirectionalEdge` where certain pentagons would produce invalid or incorrect edges for some neighbors. This is because we were testing all directions, including `K_AXES_DIGIT`, to see if they matched the destination using `h3NeighborRotations` - but apparently `h3NeighborRotations` will wrap around to other valid neighbors in the K direction for some pentagons.

The fix here is one line, but I also added a more exhaustive test for all pentagons at all resolutions.

### DIFF
--- a/src/apps/testapps/testH3UniEdge.c
+++ b/src/apps/testapps/testH3UniEdge.c
@@ -124,13 +124,22 @@ SUITE(h3UniEdge) {
     }
 
     TEST(getH3UnidirectionalEdgeFromPentagon) {
-        H3Index pentagon;
-        setH3Index(&pentagon, 0, 4, 0);
-        H3Index adjacent;
-        setH3Index(&adjacent, 0, 8, 0);
+        H3Index pentagon = 0x801dfffffffffff;
+        H3Index ring[7] = {0};
+        H3Index edge;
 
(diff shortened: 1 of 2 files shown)


src/apps/testapps/testH3UniEdge.c
```diff
@@ -124,13 +124,31 @@ SUITE(h3UniEdge) {
     }
 
     TEST(getH3UnidirectionalEdgeFromPentagon) {
+        H3Index pentagons[NUM_PENTAGONS] = {0};
```
isaacbrodsky, line 127: nit: `= {H3_NULL}` now? I prefer using the constant instead of 0 when comparing H3Index to be clear what the expected type is.

nrabinowitz: Agree for comparison, and updated below, but this is actually initializing the array to `0`, right? According to [this SO answer](https://stackoverflow.com/a/201116/380487), this wouldn't work for any non-`0` value, so a) it's idiomatic in a way that `{CONST_RESOLVING_TO_ZERO}` isn't, and b) putting in `{H3_NULL}` requires the reader to know that `H3_NULL` resolves to zero in order to understand that we're initializing the whole array, not just the first element.

Issue #396, opened 2020-09-30
Changes to support h3 builds on FreeBSD


dcj:
I modified all the shebang lines in the bash scripts to `#!/usr/bin/env bash`

`sudo pkg install bash cmake gmake doxygen lcov # can’t find the others`

And followed the instructions, `make test` returned: 

`100% tests passed, 0 tests failed out of 120
Total Test time (real) = 101.22 sec`

Pull request coming...

ajfriend:
Any chance these changes would address the issue here: https://github.com/uber/h3-py/issues/170

isaacbrodsky:
Closing as these changes have been released in v3.7.1, thanks!

PR #397, opened 2020-09-30
Changed bash script shebang lines to find bash via /usr/bin/env

dcj:
Minor changes in support of FreeBSD builds: #396


dfellis (approved):
I am incredibly surprised that this is what was breaking FreeBSD. I thought putting `sh` and `bash` in `/bin/` was part of the POSIX spec itself?

dfellis:
Oh, in #396 you mentioned the command you ran to install the dependencies. Would you mind also adding that to the README @dcj ?

isaacbrodsky (approved):

PR #399, opened 2020-10-02
Add invalid input test cases for exactEdgeLength

isaacbrodsky:


isaacbrodsky:
This should be about ready for review. I believe the issue was an invalid startVertex was passed in to `_faceIjkToGeoBoundary`, which leads to a segfault in `_adjustOverageClassII` around line 901 (`fijk->face` has an invalid value, leading to out of bounds read)

ajfriend:
LGTM! And just to verify, we're still not validating the input here, we're just avoiding the currently identified crash.

It would still be possible, for example, to input an an `H3Index` which is not valid, but does return a nonzero length, right? Should we add a test to be specific about that behavior?

The example I'm thinking of would be to take a valid H3 edge, and but change its `mode` bits to be for a cell. The result won't be a valid cell or edge, but if input to `exactEdgeLengthRads` it would give the same answer as passing in the unmodified edge.

ajfriend (approved):

ajfriend:
Put another way, are we trying to give the following guarantee: "If you provide invalid input, the output is undefined, but the function at least shouldn't crash"?

dfellis (approved):

isaacbrodsky:
> Put another way, are we trying to give the following guarantee: "If you provide invalid input, the output is undefined, but the function at least shouldn't crash"?

I think the library's direction for this has been:

* Priority one is to not crash. Crashes in C code tend to be very bad (difficult to reproduce, may have security implications, break the error handling in other languages the library is bound to)
* Priority two is to try to do something useful. Like you mentioned, if the caller didn't set the mode bits correctly but did set the direction, we can try to resolve the edge without checking the rest of the index. (Usually this is a side effect of only validating what we need)

PR #404, opened 2020-10-15
Improve GitHub Actions workflows

smorimoto:
As the title says.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -18,7 +18,7 @@ jobs:
         compiler: [clang, gcc]
 
     steps:
-      - uses: actions/checkout@v2.1.1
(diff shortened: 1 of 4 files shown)


smorimoto:
`FOSSA_API_KEY` does not seem to be found. I think we can remove the Action from the workflow or just add it, what do you think?

isaacbrodsky:
Thanks for the PR!

> `FOSSA_API_KEY` does not seem to be found. I think we can remove the Action from the workflow or just add it, what do you think?

Yes, unfortunately we haven't migrated the secret API key over from Travis CI yet. :( We can remove the action until that's done (and then use the `fossa-contrib/fossa-action@v1` action you use here)

smorimoto:
I see. Then let's disable it once. I will commit that change later 🙂

PR #420, opened 2021-01-09
Add isValidVertex

nrabinowitz:
Adds the `isValidVertex` function, with tests.

### DIFF
--- a/src/apps/testapps/testVertex.c
+++ b/src/apps/testapps/testVertex.c
@@ -88,12 +88,34 @@ SUITE(Vertex) {
 
     TEST(isValidVertex_hex) {
         H3Index origin = 0x823d6ffffffffff;
-        H3Index vert;
+        H3Index vert = 0x2222597fffffffff;
+
+        t_assert(H3_EXPORT(isValidVertex)(vert), "known vertex is valid");
 
         for (int i = 0; i < NUM_HEX_VERTS; i++) {
             vert = H3_EXPORT(cellToVertex)(origin, i);
             t_assert(H3_EXPORT(isValidVertex)(vert), "vertex is valid");
         }
+    }
+
+    TEST(isValidVertex_badOwner) {
+        H3Index origin = 0x823d6ffffffffff;
+        int vertexNum = 0;
+        H3Index vert = H3_EXPORT(cellToVertex)(origin, vertexNum);
+
+        // Assert that origin does not own the vertex
+        H3Index owner = vert;
+        H3_SET_MODE(owner, H3_HEXAGON_MODE);
+        H3_SET_RESERVED_BITS(owner, 0);
+
+        t_assert(origin != owner, "origin does not own the canonical vertex");
+
+        H3Index nonCanonicalVertex = origin;
+        H3_SET_MODE(nonCanonicalVertex, H3_VERTEX_MODE);
+        H3_SET_RESERVED_BITS(nonCanonicalVertex, vertexNum);
+
+        t_assert(H3_EXPORT(isValidVertex)(nonCanonicalVertex) == 0,
+                 "vertex with incorrect owner is not valid");
     }
 
     TEST(isValidVertex_badVerts) {
--- a/src/apps/testapps/testVertex.c
+++ b/src/apps/testapps/testVertex.c
@@ -85,4 +85,35 @@ SUITE(Vertex) {
         t_assert(directionForVertexNum(pentagon, 5) == INVALID_DIGIT,
                  "invalid pent vertex should return invalid direction");
     }
+
+    TEST(isValidVertex_hex) {
+        H3Index origin = 0x823d6ffffffffff;
+        H3Index vert;
+
+        for (int i = 0; i < NUM_HEX_VERTS; i++) {
+            vert = H3_EXPORT(cellToVertex)(origin, i);
+            t_assert(H3_EXPORT(isValidVertex)(vert), "vertex is valid");
+        }
+    }
+
+    TEST(isValidVertex_badVerts) {
+        H3Index origin = 0x823d6ffffffffff;
+        t_assert(H3_EXPORT(isValidVertex)(origin) == 0, "cell is not valid");
+
+        H3Index fakeEdge = origin;
+        H3_SET_MODE(fakeEdge, H3_UNIEDGE_MODE);
+        t_assert(H3_EXPORT(isValidVertex)(fakeEdge) == 0,
+                 "edge mode is not valid");
+
+        H3Index vert = H3_EXPORT(cellToVertex)(origin, 0);
+        H3_SET_RESERVED_BITS(vert, 6);
+        t_assert(H3_EXPORT(isValidVertex)(vert) == 0,
+                 "invalid vertexNum is not valid");
+
+        H3Index pentagon = 0x823007fffffffff;
+        H3Index vert2 = H3_EXPORT(cellToVertex)(pentagon, 0);
+        H3_SET_RESERVED_BITS(vert2, 5);
+        t_assert(H3_EXPORT(isValidVertex)(vert2) == 0,
+                 "invalid pentagon vertexNum is not valid");
+    }
 }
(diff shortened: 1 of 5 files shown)


src/h3lib/lib/vertex.c
```diff
@@ -269,3 +269,26 @@ void H3_EXPORT(vertexToPoint)(H3Index vertex, GeoCoord* coord) {
...
+ * @param  vertex H3 index possibly describing a vertex
+ * @return        Whether the input is valid
+ */
+int H3_EXPORT(isValidVertex)(H3Index vertex) {
+    if (H3_GET_MODE(vertex) != H3_VERTEX_MODE) {
+        return 0;
+    }
+
+    int vertexNum = H3_GET_RESERVED_BITS(vertex);
+    H3Index owner = vertex;
```
ajfriend, line 284: Do we need the extra variable `owner` here? Could you just work with `vertex` and avoid the copy? If just for naming purposes, maybe change the input parameter name to `index` (which is maybe actually more correct, since it is perfectly valid to input a non-vertex index into this function)?

src/h3lib/lib/vertex.c
```diff
@@ -269,3 +269,26 @@ void H3_EXPORT(vertexToPoint)(H3Index vertex, GeoCoord* coord) {
     // Copy from boundary to output coord
     *coord = gb.verts[0];
 }
+
+/**
+ * Whether the input is a valid H3 vertex
+ * @param  vertex H3 index possibly describing a vertex
+ * @return        Whether the input is valid
+ */
+int H3_EXPORT(isValidVertex)(H3Index vertex) {
```
ajfriend, line 278: Do vertices have a unique representation? That is, for a given vertex, is it only allowed a single cell "owner"?

It doesn't look like this code checks if the vertex has the correct owner.

ajfriend:
I probably should have figured this out with the previous PR, but how exactly do vertices work? Is it the case that some cells are allowed to "own" vertices, and some are not?

If so, would it be possible to provide a function `is_allowed_to_own_vertices(H3Index cell)`?

If that isn't the case, however, does that mean vertices are not uniquely represented? For example, for a given vertex, could be `vertexNum=2` for one cell, but `vertexNum=5` for a different incident cell?

ajfriend:
Adding this for consideration: https://github.com/uber/h3/issues/421

src/apps/testapps/testVertex.c
```diff
@@ -85,4 +85,35 @@ SUITE(Vertex) {
         t_assert(directionForVertexNum(pentagon, 5) == INVALID_DIGIT,
                  "invalid pent vertex should return invalid direction");
     }
+
+    TEST(isValidVertex_hex) {
```
isaacbrodsky, line 89: Please add a test for a specific valid vertex, to validate the right mode number is being used.

src/h3lib/lib/vertex.c
isaacbrodsky: My assumption is this function will assure the vertex is in the canonical form, if we want to support non-canonical vertexes, perhaps we need `canonicalizeVertex`/`isCanonicalVertex`?

src/h3lib/lib/vertex.c
nrabinowitz: That's a good question - I guess it should be fine to modify `vertex`, though I generally think it's confusing to modify arguments in-place. Will update.

src/h3lib/lib/vertex.c
nrabinowitz: Hm. That's a good point, though there's no way at present that you can create a non-canonical vertex, so I'm leery about adding public functions addressing a practically unreachable issue. If we wanted this check, I think I'd need to implement `vertexToCells` first, and then check that the owner had the lowest numerical index. 

I don't have a clear picture for the validation functions on whether we should prioritize speed (because you might need to validate many indexes at once) or exhaustiveness, which in this case would mean an expensive check for an unlikely case.

So I guess this brings up two questions:
- Is there a use for a non-canonical vertex (i.e. an index indicating a cell and vertex number)? This would be easy to produce, but I'm not sure it has any value, and it would muddy the waters about whether all vertexes are canonical.
- If not, is it worth an expensive check in this function for a case that at present would need to be hand-crafted?

nrabinowitz:
> I probably should have figured this out with the previous PR, but how exactly do vertices work? Is it the case that some cells are allowed to "own" vertices, and some are not?

See https://github.com/uber/h3/blob/master/dev-docs/RFCs/v4.0.0/vertex-mode-rfc.md - the basic idea is, you take the three cells sharing the vertex, and assign the lowest numeric index as the canonical owner. Ownership is an implementation detail, and shouldn't be exposed in the public API I think.

> If so, would it be possible to provide a function `is_allowed_to_own_vertices(H3Index cell)`?

Nope, because a) not possible (a cell could own some vertices and not others), and b) not public info.

> If that isn't the case, however, does that mean vertices are not uniquely represented? For example, for a given vertex, could be `vertexNum=2` for one cell, but `vertexNum=5` for a different incident cell?

See above - we just pick one cell and say that's the owner. What you're describing here is a non-canonical vertex, which I haven't implemented and don't see much use for, though let me know if you think there's some value.

src/apps/testapps/testVertex.c
nrabinowitz: Will do.

src/h3lib/lib/vertex.c
ajfriend: I think we definitely want uniqueness for the indices (and thus not allow "non-canonical" vertices), and if we have a function called `isValidVertex` it has to be exhaustive and strictly correct. I think speed is secondary to correctness, and is something we could potentially improve in the future.

src/h3lib/lib/vertex.c
dfellis: I'm actually going to recommend the opposite tack. Getting the `owner` H3Index is *vital* because I think this function should confirm that it can reproduce the original `vertex` value in here.

That will make this function slower, but since you'd only be validating the vertex if you're not sure about it, and since there are non-canonical representations that could be reached through some bit manipulation, it feels like the only time you'd be calling `isVertexValid` would be when you *aren't* getting them generated by the H3 library itself.

nrabinowitz: Oh huh. I hadn't even thought of it, but yeah, the best way to check whether we have the canonical representation is to run this through `cellToVertex` again and compare output. Easier than what I was thinking, where I'd have to get neighbors again. Good idea.

src/h3lib/lib/vertex.c
nrabinowitz: Updated to check validity by recreating the valid vertex and comparing, plus added a test for the incorrect-owner case. Recreating the vertex also covers the invalid vertex number case, so this simplified things nicely.

dfellis (approved):

src/h3lib/lib/vertex.c
```diff
@@ -269,3 +269,28 @@ void H3_EXPORT(vertexToPoint)(H3Index vertex, GeoCoord* coord) {
...
+    H3_SET_RESERVED_BITS(owner, 0);
+
+    // The easiest way to ensure that the owner + vertex number is valid,
+    // and that the vertex is canonical, is to recreate and compare.
+    H3Index canonical = H3_EXPORT(cellToVertex)(owner, vertexNum);
+    if (vertex != canonical) {
+        return 0;
+    }
+
+    return H3_EXPORT(h3IsValid)(owner);
```
isaacbrodsky, line 295: May want to check that the owner is valid first?

isaacbrodsky (approved):

src/h3lib/lib/vertex.c
nrabinowitz: Hm. I guess it depends whether `h3IsValid` is faster - but it also would avoid handing invalid indexes to `cellToVertex`, so I think that makes sense.

PR #422, opened 2021-01-21
Vertex optimizations, first pass

nrabinowitz:
Small optimizations for `cellToVertex`. The main benefit of this work was to identify the key perf bottleneck, which is the call to `_h3ToFaceIjk` in `vertexRotations`. The optimizations here are mainly to avoid expensive calls:

- Skip calculating left and right neighbors if the origin is a center child (center children will always have the lowest index)
- Skip calculating right neighbor if the left neighbor is a center child
- For hexagon cells, use a simple direction lookup table to determine the reverse direction, instead of using the slightly slower `directionForNeighbor`.

This improves things, but doesn't add up as much as I'd like - approx 20% speedup:

**Before** 
```
	-- cellToVertexes: 257.191600 microseconds per iteration (10000 iterations)
	-- cellToVertexesPent: 267.356200 microseconds per iteration (10000 iterations)
```

**After**
```
	-- cellToVertexes: 13.119300 microseconds per iteration (10000 iterations)
	-- cellToVertexesPent: 0.357300 microseconds per iteration (10000 iterations)
	-- cellToVertexesRing: 215.354700 microseconds per iteration (10000 iterations)
	-- cellToVertexesRingPent: 227.581100 microseconds per iteration (10000 iterations)
```

I think I can improve on this significantly by refactoring `cellToVertexes` to avoid repeated calculations for the origin and neighbors, but that will come in a later PR.

### DIFF
--- a/src/apps/benchmarks/benchmarkVertex.c
+++ b/src/apps/benchmarks/benchmarkVertex.c
@@ -19,6 +19,10 @@
 
 // Fixtures. Cells are arbitrary, except that ring2 is all hexagons and
 // ring2Pent is centered on a pentagon.
+
+H3Index hex = 0x89283080ddbffff;
+H3Index pentagon = 0x89080000003ffff;
+
 H3Index ring2[] = {
     0x89283081083ffff, 0x8928308109bffff, 0x8928308108bffff, 0x8928308108fffff,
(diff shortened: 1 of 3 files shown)


dfellis (approved):
Looks good. The overall complexity isn't much higher and the perf bump is good.

isaacbrodsky (approved):

ajfriend (approved):

Issue #392, opened 2020-09-10
Guard against potential out-of-bounds lookups in base cell tables


nrabinowitz:
The functions in `baseCells.c` should guard against invalid base cell numbers that could result in out-of-bounds reads on the lookup tables.

alexey-milovidov:
For zero-cost implementation (safe but implementation specific behaviour on out-of-bounds lookup) we can simply extend the arrays to 127 elements (filling the rest with zeros).

PR #423, opened 2021-01-21
Add range checks

alexey-milovidov:
Implement https://github.com/uber/h3/issues/392

Not sure if everything is correct, I did not dig into the algorithms.
But at least it fixes https://github.com/ClickHouse/ClickHouse/issues/19219

There should not be significant performance degradation for two reasons:
- the branch is easy predictable and it can be executed in parallel due to instruction level parallelism;
- the algorithms are already heavy enough, the difference in a few clock ticks should be neglible.

### DIFF
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -290,6 +290,9 @@ H3Index h3NeighborRotations(H3Index origin, Direction dir, int* rotations) {
 
     int newRotations = 0;
     int oldBaseCell = H3_GET_BASE_CELL(out);
+    if (oldBaseCell < 0 || oldBaseCell >= NUM_BASE_CELLS)
+        return 0;
+
     Direction oldLeadingDigit = _h3LeadingNonZeroDigit(out);
 
     // Adjust the indexing digits and, if needed, the base cell.
--- a/src/h3lib/lib/baseCells.c
+++ b/src/h3lib/lib/baseCells.c
@@ -822,6 +822,8 @@ const BaseCellData baseCellData[NUM_BASE_CELLS] = {
 
 /** @brief Return whether or not the indicated base cell is a pentagon. */
 int _isBaseCellPentagon(int baseCell) {
+    if (baseCell < 0 || baseCell >= NUM_BASE_CELLS)
+        return 0;
     return baseCellData[baseCell].isPentagon;
 }
 
@@ -840,6 +842,8 @@ bool _isBaseCellPolarPentagon(int baseCell) {
  * Valid ijk+ lookup coordinates are from (0, 0, 0) to (2, 2, 2).
  */
 int _faceIjkToBaseCell(const FaceIJK* h) {
+    if (!h || h->face < 0 || h->face >= NUM_ICOSA_FACES)
+        return 0;
     return faceIjkBaseCells[h->face][h->coord.i][h->coord.j][h->coord.k]
         .baseCell;
 }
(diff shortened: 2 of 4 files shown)


alexey-milovidov:
I see there are some issues with formatting. I don't think I'm ready to finish this PR, feel free to apply any edits or close and resubmit another PR...

src/h3lib/lib/baseCells.c
```diff
@@ -840,6 +842,8 @@ bool _isBaseCellPolarPentagon(int baseCell) {
  * Valid ijk+ lookup coordinates are from (0, 0, 0) to (2, 2, 2).
  */
 int _faceIjkToBaseCell(const FaceIJK* h) {
+    if (!h || h->face < 0 || h->face >= NUM_ICOSA_FACES)
```
nrabinowitz, line 845: @isaacbrodsky - Do you think we need guards on `face` here? Or is this value only calculated within the library? If we can't trust `face`, we should probably add guards for `coord.i`, `coord.j`, and `coord.k` as well.

nrabinowitz:
Thanks for submitting this! You should be able to fix the test failures with `make format`. The bigger issue here is probably test coverage - we'll need to add tests for the new branches introduced here.

@isaacbrodsky - thoughts on bringing this into 3.x? This is potentially a security issue.

sahrk:
I haven't been following the issues daily so sorry for being late to the conversation again...

The lack of range checks on arrays throughout H3 core (including the recent `baseCells` array merge) was intentional (as it is in the C language itself). Personally I think that performance degradation in H3 core should be avoided at all costs. The hit might be small for each change, but such things add up. Rather than changing the core functions, I would strongly encourage you to add "safe" wrapper functions that can be used in situations where the user doesn't want the onus of checking their own array indexes when necessary.

src/h3lib/lib/algos.c
```diff
@@ -290,6 +290,9 @@ H3Index h3NeighborRotations(H3Index origin, Direction dir, int* rotations) {
 
     int newRotations = 0;
     int oldBaseCell = H3_GET_BASE_CELL(out);
+    if (oldBaseCell < 0 || oldBaseCell >= NUM_BASE_CELLS)
+        return 0;
```
isaacbrodsky, line 294: nit: return `H3_NULL`

src/h3lib/lib/baseCells.c
isaacbrodsky: I am a little curious what case this is guarding against. While I agree with preventing undefined behavior, I am not sure offhand whether this is a needed check.

isaacbrodsky (commented):
To @sahrk' point, another suggestion posted in the issues is to expand the base cells array and ensure we never calculate an index into that array outside of 0-127 inclusive.

I would like to ensure that H3 is resilient to dereferencing out of bounds. Beyond the security concerns, this type of crash is often very difficult to debug, especially when it happens in a library loaded into a manageed framework. I would be willing to create a new 3.x release to incorporate this kind of change.

If possible, could you post test cases demonstrating some of these? Even if not coded in olur test framework it would be very helpful for coding the tests.

sahrk:
> To @sahrk' point, another suggestion posted in the issues is to expand the base cells array and ensure we never calculate an index into that array outside of 0-127 inclusive.

I should point-out that the baseCells and the faces in faceIjkToBaseCell are not the only arrays that are accessed without bounds checking in the core. There are a whole bunch of arrays in the core, many with multiple dimensions. If I ever did bounds checking on any of them it was by mistake :).

The core was designed to be as tight as possible, which is why so much of it is table lookups, which are as cheap as it gets in C. In the case of the baseCells we're talking about a function that's just a wrapper for an array access. However you slice it, if you add bounds checking you're at least doubling the cost of the function. 

> I would like to ensure that H3 is resilient to dereferencing out of bounds. Beyond the security concerns, this type of crash is often very difficult to debug, especially when it happens in a library loaded into a manageed framework. I would be willing to create a new 3.x release to incorporate this kind of change.

That is all reasonable of course. Earlier I suggested making safe versions of table lookup functions that are exposed to external users (like baseCells), but what if you made the API functions safe and created internal functions without bounds checking for use just in the core?

PR #427, opened 2021-02-01
Just adding my translation

RichardVasquez:
While mine is a translation rather than a binding, it seemed like the appropriate location to put the link.

### DIFF
--- a/docs/community/bindings.md
+++ b/docs/community/bindings.md
@@ -9,7 +9,7 @@ As a C library, bindings can be made to call H3 functions from different program
 ## C&#35;
 
 - [entrepreneur-interet-general/h3.standard](https://github.com/entrepreneur-interet-general/H3.Standard)
-- [richardvasquez/h3net](https://github.com/RichardVasquez/h3net)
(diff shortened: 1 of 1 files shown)


docs/community/bindings.md
```diff
@@ -9,6 +9,7 @@ As a C library, bindings can be made to call H3 functions from different program
 ## C&#35;
 
 - [entrepreneur-interet-general/h3.standard](https://github.com/entrepreneur-interet-general/H3.Standard)
+- [richardvasquez/h3net](https://github.com/RichardVasquez/h3net)
```
isaacbrodsky, line 12: Might be useful to note this is a translation rather than a binding for those interested. Not totally necessary, we only have two links. :)

isaacbrodsky (approved):
Thanks for contributing this!

docs/community/bindings.md
RichardVasquez: Noted and updated

dfellis (approved):

PR #433, opened 2021-02-16
Add fuzzer harnesses

isaacbrodsky:
This pull request adds harness programs for use with the AFL fuzzer in order to detect crashes. At first, tests for `geoToH3`, `h3ToGeo`, `h3ToGeoBoundary`, and `kRing` are added.

### DIFF
--- a/src/apps/fuzzers/README.md
+++ b/src/apps/fuzzers/README.md
@@ -1,0 +1,34 @@
+# Fuzzer harnesses for H3
+
+This directory contains helper programs for testing the H3 library using the
+''[American fuzzy lop](https://lcamtuf.coredump.cx/afl/)'' fuzzer.
+
+# Installation
+
+```
+apt install afl-clang
+```
+
+(There is also an afl-cov which looks interesting but isn't necessary.)
+
+# Usage
+
+You must compile with the instrumented compiler:
+
+```
+CXX=afl-clang++ CC=afl-clang cmake .
+make fuzzers
+```
+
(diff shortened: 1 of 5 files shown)


dfellis (approved):

src/apps/fuzzers/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```
+CXX=afl-clang++ CC=afl-clang cmake .
+make fuzzers
+```
+
+An individual fuzzer run is invoked as follows. The argument is a file containing the number of bytes needed.
+
+```
+fuzzerGeoToH3 bytes24
```
nrabinowitz, line 26: I'm a bit confused by the file arguments here. Are these files we could check into the repo?

nrabinowitz (approved):
Looks good, though I'm not sure I'd be able to run the fuzzers based on the docs w/o a fair amount of trial and error.

src/apps/fuzzers/README.md
dfellis: Aren't they invoked by the new fuzzer test commands with random data over and over again in the attempt to cause breakage?

isaacbrodsky: Yes, this is an example of how to test the fuzzer program (check that it does what you think it does) before putting it through AFL. I didn't check in the initial files (like `bytes24`) because they had no useful content - they were just ASCII `'0'` repeated a given number of times and then I had the fuzzer permute from that. I could add either a script that sets up the file or check in a binary file of the right size to make it easier to start the fuzzer.

PR #441, opened 2021-04-03
Naming RFC: drop `Cells` from `gridPath`, `compact`, `uncompact`

ajfriend:
The name `gridPathCells` was originally intended to distinguish from future possible names like `gridPathEdges` or `gridPathVertexes`. However, `gridPathCells` is a bit of an awkward name, and we may never actually implement the edge and vertex versions.

Instead, we're proposing just `gridPath`. 

This could extend to a general naming rule: **When an H3 mode is not specified (cell, edge, vertex...), the user should assume the function pertains to Cells.**

### DIFF
--- a/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
+++ b/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
@@ -13,6 +13,7 @@
 - **Discussions**:
     - <https://github.com/uber/h3/pull/308>
     - <https://github.com/uber/h3/pull/403>
+    - <https://github.com/uber/h3/pull/441>
 
 ## Motivation
 
--- a/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
+++ b/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
@@ -140,8 +140,8 @@ There is some ambiguity between property, transform, and computation, so use you
 | `h3GetFaces`                  | `getIcosahedronFaces` |
 | `geoToH3`                     | `pointToCell`         |
 | `h3ToGeo`                     | `cellToPoint`         |
-| `compact`                     | `compactCells`        |
-| `uncompact`                   | `uncompactCells`      |
+| `compact`                     | (unchanged)           |
+| `uncompact`                   | (unchanged)           |
 | `polyfill`                    | `polygonToCells`      |
 
 **Note**: `getResolution` and `getBaseCellNumber` should work for both cells and edges.
--- a/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
+++ b/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
@@ -8,9 +8,11 @@
 - **Dates**:
     - Started: 2020-02-02
     - Accepted: 2020-03-26
+    - Updated: 2021-04-03
 - **Status**: Accepted
 - **Discussions**:
     - <https://github.com/uber/h3/pull/308>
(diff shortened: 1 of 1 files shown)


ajfriend:
We might consider this new "cell is assumed if omitted" rule for a few other functions:

| Current library name | Current RFC proposal |
|----------------------|----------------------|
| `compact`            | `compactCells`       |
| `uncompact`          | `uncompactCells`     |
| `polyfill`           | `polygonToCells`     |


Of those, I think `polygonToCells` is an improvement, but the  compact/uncompact functions probably don't need the full specificity of "Cells", which can just be implicit. What would folks think about the following change to the RFC?

| Current library name | New RFC proposal |
|----------------------|------------------|
| `compact`            | (unchanged)      |
| `uncompact`          | (unchanged)      |
| `polyfill`           | `polygonToCells` |

isaacbrodsky (approved):
I think this matches most user's experiences, which tend to center around the cell mode. 

It also seems it would evolve well in the future if non-cell based modes become more a draw for H3, since we can append "Cells" to the name

dfellis:
Oh, you merged this before I could leave a comment. I don't really agree with this proposal. Or at least I don't agree with the partial application if we are going to make "Cells" a default choice. `polyfill` should be reverted back, as well, in my opinion.

ajfriend:
By revert, do you mean the function could be called `polyfill` or it should be called `polygonToCells`?

dfellis:
> By revert, do you mean the function could be called `polyfill` or `polygonToCells`?

In the context of this RFC, I'd put it back to `polyfill`. But if that isn't desired, I'd revert the RFC change itself and put `Cells` back on all of the functions? Being consistent one way or the other is what I mean.

ajfriend:
I think the RFC as is is still fairly consistent, even with `polygonToCells`.

dfellis:
> I think the RFC as is is still fairly consistent, even with `polygonToCells`.

I don't agree.

ajfriend:
How I'm thinking about it (and why I don't think it causes any consistency issues) is that we have two types of function names here.

1. transformation names like `<X>to<Y>`
2. operation names like `<Operation><H3_Mode>`

Then `polygonToCells` has `<X>to<Y>` type.

Similarly, `gridPath`, `compact`, and `uncompact` are of `<Operation><H3_Mode>` type, but since

1. `gridPathCells` is a bit cumbersome to type (same with `compactCells` and `uncompactCells`),
2. `Cells` are the "default" mode that users will assume,
3. we may never actually implement these operations for other types, and
4. there's little chance of confusion,

we can drop the `<H3_mode>` and leave `Cells` as implicitly assumed.

That leaves us the option for things like `gridPathDirectedEdge` in the future, or even adding a `gridPathCells` alias.

ajfriend:
For `polyfill`, I think a previous discussion (with @nrabinowitz?) came around to the idea that `polygonToCells` was a slightly more descriptive name.

I'm actually OK with either, but with a slight preference for `polygonToCells`.

And this is actually one of the operations that I think is most likely to have versions for different H3 Modes. Following the convention we have above, we would have the options:

|              |        `polyfill`       |      `polygonTo<Y>`      |
|--------------|-------------------------|--------------------------|
| Cell         | `polyfill`              | `polygonToCells`         |
| DirectedEdge | `polyfillDirectedEdges` | `polygonToDirectedEdges` |
| Vertex       | `polyfillVertexes`      | `polygonToVertexes`      |

With an optional alias or alternative name of `polyfillCells`.

ajfriend:
Moving this conversation to https://github.com/uber/h3/issues/451

PR #448, opened 2021-04-14
Add libfuzzer fuzzer for continuous fuzzing

AdamKorcz:
Hi all, this is [Adam](https://twitter.com/AdamKorcz4) of [AdaLogics](https://adalogics.com/). I work on securing open source software and have worked on setting up continuous fuzzing of h3 through OSS-fuzz.

Since we cannot use the existing AFL fuzzers for the OSS-fuzz integration, I have added a libfuzzer fuzzer in this PR. Once this one is merged, the build on the OSS-fuzz side (https://github.com/google/oss-fuzz/pull/5616) will pass. 

Since each target of h3 in general is pretty small I have added them all in a single fuzzer. For now I recommend following this until it becomes stricly necessary to have more fuzzers, since there is little point in having dedicated fuzzers for 100-200 lines of code in terms of maintenance.

I will be happy to add more targets to the fuzzer. This will become easier once h3 is fuzzed by OSS-fuzz, since OSS-fuzz offers some useful diagnostic tools and coverage reports.

On the OSS-fuzz side a maintainers email address is needed for bug reports.

### DIFF
--- a/src/apps/fuzzers/libfuzzer.c
+++ b/src/apps/fuzzers/libfuzzer.c
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2021 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+#include "h3api.h"
+#include "utility.h"
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    char *new_str = (char *)malloc(size+1);
+    if (new_str == NULL){
+        return 0;
+    }
+    memcpy(new_str, data, size);
+    new_str[size] = '\0';
+
+    
+    H3Index h3;
+    h3 = H3_EXPORT(stringToH3)(new_str);
+    
+    H3Index input[] = {h3, h3};
+    int inputSize = sizeof(input) / sizeof(H3Index);
+
+    // fuzz compactCells
+    H3Index* compacted = calloc(inputSize, sizeof(H3Index));
+    int err = compactCells(input, compacted, inputSize);
+    
+    // fuzz uncompactCells
+    int compactedCount = 0;
+    for (int i = 0; i < inputSize; i++) {
+        if (compacted[i] != 0) {
+            compactedCount++;
+        }
+    }
+    if(uncompactedSize<10) {
+        int uncompactRes = 10;
+        int uncompactedSize =
+            maxUncompactCellsSize(compacted, inputSize, uncompactRes);
+        
+        H3Index* uncompacted = calloc(uncompactedSize, sizeof(H3Index));
+        int err2 = uncompactCells(compacted, compactedCount, uncompacted,
+                              uncompactedSize, uncompactRes);
+        free(uncompacted);
+    }
+    
+    // fuzz h3NeighborRotations 
+    int rotations = 0;
+    h3NeighborRotations(h3, CENTER_DIGIT, &rotations);
+
+    free(compacted);    
+    free(new_str);
+    return 0;
+}
--- a/src/apps/fuzzers/libfuzzer.c
+++ b/src/apps/fuzzers/libfuzzer.c
@@ -22,6 +22,17 @@
 #include "h3api.h"
 #include "utility.h"
 
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+#include "h3api.h"
+#include "utility.h"
+
+static const Direction DIGITS[7] = {CENTER_DIGIT, K_AXES_DIGIT, J_AXES_DIGIT, JK_AXES_DIGIT, I_AXES_DIGIT,
+     IK_AXES_DIGIT, IJ_AXES_DIGIT};
+
 int
 LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
@@ -32,17 +43,17 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     memcpy(new_str, data, size);
     new_str[size] = '\0';
 
-    
+
     H3Index h3;
     h3 = H3_EXPORT(stringToH3)(new_str);
-    
+
     H3Index input[] = {h3, h3};
     int inputSize = sizeof(input) / sizeof(H3Index);
 
     // fuzz compactCells
     H3Index* compacted = calloc(inputSize, sizeof(H3Index));
     int err = compactCells(input, compacted, inputSize);
-    
+
     // fuzz uncompactCells
     int compactedCount = 0;
     for (int i = 0; i < inputSize; i++) {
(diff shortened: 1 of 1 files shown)


dfellis:
So I don't think this will work as-is, because in here you call the file `libfuzzer.c` and in the other PR you call it [`h3_fuzzer.c`](https://github.com/google/oss-fuzz/pull/5616/files#diff-dbfede3af30407e0d2ee688c33f1c9ff06b72b9a667989aa53f74e5e9e89ec1aR28) which doesn't exist.

I would also note that there is no explanatory comments in this PR about what the purpose of this file is, so it could get accidentally deleted. I would definitely include comments with a link to the https://github.com/google/oss-fuzz project to make this clear to future developers.

Finally, I don't personally care for the compile rules for this file all living in the other repo. It would be relatively easy to break things if the code here is refactored. Better to have it in cmake as a new make rule that is simply invoked in the other repo. Ideally also with a clear way to run this fuzzing locally for debugging with gdb attached? I'm not sure what kind of report, if any, oss-fuzz provides on failures beyond pass/fail.

nrabinowitz:
This looks great! We have some fuzzers in the lib, but being able to run continuous fuzzing would be great. Some thoughts here:

- It looks like you added a [custom build target](https://github.com/google/oss-fuzz/pull/5616/files#diff-dbfede3af30407e0d2ee688c33f1c9ff06b72b9a667989aa53f74e5e9e89ec1a) for the new fuzzer file in your `oss-fuzz` PR. We should probably add a Cmake target for this instead; that would reduce what the `oss-fuzz` code has to know about the internals of the H3 repo, and it would make it easier to update the build if we add additional files, etc.
- We should probably run the fuzzer you have here with known-good input as part of the standard CI test suite. That would ensure that it compiles and runs (for example, we just landed a significant renaming PR, which will be the basis for a `4.0.0-alpha-1` release, and I'd be surprised if this compiles at present).

src/apps/fuzzers/libfuzzer.c
```diff
@@ -0,0 +1,71 @@
...
+        
+        H3Index* uncompacted = calloc(uncompactedSize, sizeof(H3Index));
+        int err2 = uncompactCells(compacted, compactedCount, uncompacted,
+                              uncompactedSize, uncompactRes);
+        free(uncompacted);
+    }
+    
+    // fuzz h3NeighborRotations 
+    int rotations = 0;
+    h3NeighborRotations(h3, CENTER_DIGIT, &rotations);
```
isaacbrodsky, line 66: Note that `h3NeighborRotations` won't be exported on Windows, so I don't expect this to work on that platform.

Could the digit be included as one of the uffzer inputs (in the range 0-6, inclusive)? I don't know off hand but I expect CENTER_DIGIT to be short circuited in this function.

src/apps/fuzzers/libfuzzer.c
```diff
@@ -0,0 +1,71 @@
...
+    char *new_str = (char *)malloc(size+1);
+    if (new_str == NULL){
+        return 0;
+    }
+    memcpy(new_str, data, size);
+    new_str[size] = '\0';
+
+    
+    H3Index h3;
+    h3 = H3_EXPORT(stringToH3)(new_str);
```
isaacbrodsky, line 37: I'm a little unclear what's happening here. It looks like the fuzzer program is providing a block of input data, and this is being converted into H3 indexes? Would it make sense to just read 8 bytes (64 bits) out of that input data and treat it as an H3 Index, rather than going through stringToH3? Or is the intent to include stringToH3 in the fuzzing?

Nit that stringToH3 seems to be the only one using H3_EXPORT - I'd just remove usage of that from this file.

src/apps/fuzzers/libfuzzer.c
```diff
@@ -0,0 +1,71 @@
...
+    
+    H3Index input[] = {h3, h3};
+    int inputSize = sizeof(input) / sizeof(H3Index);
+
+    // fuzz compactCells
+    H3Index* compacted = calloc(inputSize, sizeof(H3Index));
+    int err = compactCells(input, compacted, inputSize);
```
isaacbrodsky, line 44: Not sure how rigorously this will be able to test `compactCells` since it always uses two inputs that are the same cell, but this seems like a good idea we can add other functions to.

isaacbrodsky (commented):
Thanks for this PR! This is a great contribution and it's exciting to get more tooling and analysis looking at the library.

I agree with David and Nick that having a CMake target for this file would help us compile & run it as part of our CI, and thus prevent accidentally breaking this. I'd suggest adding something like the following, above line 640 in CMakeLists.txt:

```
    if(ENABLE_REQUIRES_ALL_SYMBOLS)
        add_library(libfuzzer src/apps/fuzzers/libfuzzer.c)
        target_link_libraries(libfuzzer PUBLIC h3)
        target_include_directories(libfuzzer  PUBLIC
          $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/apps/applib/include>)
        target_compile_options(libfuzzer PRIVATE ${H3_COMPILE_FLAGS})
        target_link_libraries(libfuzzer PRIVATE ${H3_LINK_FLAGS})
    endif()
```

isaacbrodsky:
> On the OSS-fuzz side a maintainers email address is needed for bug reports.

Please let me know where a good place to provide this would be - here or on the OSS-fuzz PR or over Slack?

src/apps/fuzzers/libfuzzer.c
AdamKorcz: The intention was to both fuzz `stringToH3` as well as create a new `H3Index`.

src/apps/fuzzers/libfuzzer.c
AdamKorcz: > Note that h3NeighborRotations won't be exported on Windows, so I don't expect this to work on that platform.

Thanks! OSS-fuzz only fuzz on linux. 

> Could the digit be included as one of the uffzer inputs (in the range 0-6, inclusive)? I don't know off hand but I expect CENTER_DIGIT to be short circuited in this function.

Absolutely!

AdamKorcz:
> Please let me know where a good place to provide this would be - here or on the OSS-fuzz PR or over Slack?

The OSS-fuzz PR would be great.

AdamKorcz:
@dfellis @nrabinowitz 
> So I don't think this will work as-is, because in here you call the file libfuzzer.c and in the other PR you call it h3_fuzzer.c which doesn't exist.

You are right. The build script will require a few modifications to use the fuzzer from this PR. I will take care of that.

> I would also note that there is no explanatory comments in this PR about what the purpose of this file is, so it could get accidentally deleted. I would definitely include comments with a link to the https://github.com/google/oss-fuzz project to make this clear to future developers.

Absolutely! I will add that, however I would suggest we wait with that until H3 is integrated into OSS-fuzz. There may be other things that will need commenting in the process.

Regarding adding a rule in the build system, thank you very much for the great feedback, and thank you for the suggestion @isaacbrodsky. I will be glad to arrange a PR for that.
I suggest that this gets done after H3 is integrated into OSS-fuzz. Once the fuzzer is merged here, and the build script is merged upstream, we will have a complete picture of what needs to be added to the build script. At the moment there may still be a few adjustments, and I would prefer to have and end-to-end prototype of the integration before making the prototype easier manageable by setting up a make rule.
Once H3 is integrated into OSS-fuzz, I can look at it immediately afterwards. It should maximum take me a week for a PR with the added logic in the makefile.

What are your thoughts on the above?

AdamKorcz:
vim

src/apps/fuzzers/libfuzzer.c
AdamKorcz: We can add several random `H3Index` in `input[]`.

For now, though, I think it should be good for an initial setup.

src/apps/fuzzers/libfuzzer.c
AdamKorcz: I added so that all digits are being tested during each fuzz iteration. In case any false positives come up, we can adjust. Otherwise, as long as the fuzzer can run continuously, this is a good solution.

src/apps/fuzzers/libfuzzer.c
AdamKorcz: Closing. Feel free to reopen.

AdamKorcz:
@isaacbrodsky Kind ping. Do you have an email address for bug reports?

isaacbrodsky:
> @isaacbrodsky Kind ping. Do you have an email address for bug reports?

Oops, seems like this got dropped. I need to confirm with @ajfriend if we have a distribution group we can use for this. In the mean time you can use my email, isaac@isaacbrodsky.com for reports.

ajfriend:
@AdamKorcz, you can send bug reports to h3-dev@googlegroups.com.

The (eventual) group history can be seen here: https://groups.google.com/g/h3-dev/

isaacbrodsky:
I think this has been superseded by new fuzzers in uber/h3 - if we still want to use custom fuzzers in OSS-Fuzz please let me know

Issue #121, opened 2018-08-15
Projected coordinate system


greatathrun:
For those functions related to latitude and longitude, is the projected coordinate system used as WGS84 or something?

sahrk:
It is a sphere with the WGS84 authalic radius (i.e., a sphere with the same surface area as the WGS84  spheroid).

nrabinowitz:
Closing as answered - thanks @sahrk!

nrabinowitz:
On second thought - reopening this until we include the answer in the core documentation.

zachasme:
So should we assume coordinate system WGS84 or something like EPSG:3857 (when assigning an SRID to coordinates in PostGIS for example)?

nagytech:
I think it would be valuable to at least specify WGS84 as a coordinate system.  

In terms of SRIDs, there are always trade-offs between the different spatial references.  Specifying EPSG:3857 in the documentation would be a good starting point and may help solve [these types of questions](https://gis.stackexchange.com/questions/342845/distorted-h3-hexagons-even-though-projection-is-matching/353042).

Komzpa:
WGS84 2D lon-lat is EPSG:4326, not EPSG:3857.
You can try stealing def from GeoJSON spec:
https://tools.ietf.org/html/rfc7946
```
4.  Coordinate Reference System

   The coordinate reference system for all GeoJSON coordinates is a
   geographic coordinate reference system, using the World Geodetic
   System 1984 (WGS 84) [WGS84] datum, with longitude and latitude units
   of decimal degrees.  This is equivalent to the coordinate reference
   system identified by the Open Geospatial Consortium (OGC) URN
   urn:ogc:def:crs:OGC::CRS84.  An OPTIONAL third-position element SHALL
   be the height in meters above or below the WGS 84 reference
   ellipsoid.  In the absence of elevation values, applications
   sensitive to height or depth SHOULD interpret positions as being at
   local ground or sea level.
```

ajfriend:
Closed with #455

PR #455, opened 2021-04-26
Add note about WGS84 to docs

ajfriend:
Following up from #121, these seems like a reasonable place to add a comment about the projection we use.

I'm not as familiar with the technical details here. Could someone who is suggest a sentence or two to add? @sahrk @isaacbrodsky @nrabinowitz @dfellis ?

### DIFF
--- a/website/docs/core-library/overview.md
+++ b/website/docs/core-library/overview.md
@@ -5,7 +5,9 @@ sidebar_label: Overview
 slug: /core-library/overview
 ---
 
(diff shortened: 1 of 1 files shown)


ajfriend:
Also, @isaacbrodsky, do I need to do something else to fix the website build?

dfellis:
> Also, @isaacbrodsky, do I need to do something else to fix the website build?

`<>` are used for HTML tags, and the markdown parser we're using allows inserting arbitrary HTML (since that's what we're generating and we fully control the input) so it's parsing it as a tag and then failing hard on the second `<` you added.

isaacbrodsky:
Perhaps something like:

> The coordinate reference system (CRS) is sphere coordinates with the [WGS84](en.wikipedia.org/wiki/WGS84)/[EPSG:4326](https://epsg.io/4326) authalic radius. It is common to use WGS84 CRS data with the H3 library.

I'm attempting to adapt this from @sahrk's comments without changing any technical details. My understanding is that what is being described is not quite the same as WGS84 because H3 assumes a sphere rather than an ellipsoid, but in applications we usually treat them as interchangeable. (I.e. this was the practice at Uber)

dfellis (approved):

isaacbrodsky (approved):

PR #459, opened 2021-05-03
Docs: add branch instructions and bump

isaacbrodsky:
Adds text to the docs for users to use the most recent version, rather than development.

The latest release version is read from `VERSION` using some sh specific magic - I don't think this works on Windows CMD or PS, unlike the other lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,14 @@ sudo pkg install bash cmake gmake doxygen lcov
 
 #### Compilation
 
+When checking out the H3 Git repository, by default you will check out the latest
+development version of H3. When using H3 in an application, you will want to check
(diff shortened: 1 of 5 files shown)


nrabinowitz (approved):
Looks good. Do we want to add a note somewhere about 3.x vs 4.x changing all the names?

ajfriend (approved):

dfellis (approved):

isaacbrodsky:
> Looks good. Do we want to add a note somewhere about 3.x vs 4.x changing all the names?

I briefly discussed with @ajfriend on Slack, but I'd like to make a separate PR for that introducing versioning to the docs.

PR #461, opened 2021-05-07
Docs: Add versioning

isaacbrodsky:
Adds versioning to the docs site, adding a new 4.x section (default is still 3.x - those docs are "frozen" in the versioned_docs directory), along with trimmed versions of the naming and error code RFCs for 4.x. Edits to docs will go to 4.x first, and we can backport to 3.x if needed. Note that changes in website/docs will apply to 4.x only, which you will need to explicitly select from the drop down.


dfellis (approved):

ajfriend (approved):
Very nice!

Jonhi1393 (approved):

PR #463, opened 2021-05-12
Docs: Update for 4.x in progress for hierarchy, indexing

isaacbrodsky:
Begin updating function docs for v4. Starts with hierarchy and indexing. Note that only indexing has the error codes since that is what's in h3api.h right now.

Note that I'm also changing the argument names in the docs to try to make it a little clearer what an `H3Index` represents. We should be able to do this in h3api.h as well if we want - the argument names are not part of the API for versioning purposes.

### DIFF
--- a/website/docs/api/hierarchy.mdx
+++ b/website/docs/api/hierarchy.mdx
@@ -323,7 +323,7 @@ function example() {
 
 Uncompacts the set `compactedSet` of indexes to the resolution `res`. `cellSet` must be at least of size `maxUncompactSize(compactedSet, numCells, res)`.
 
-## maxuncompactCellsSize
+## maxUncompactCellsSize
 
 <Tabs
   groupId="language"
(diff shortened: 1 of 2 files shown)


website/docs/api/indexing.mdx
```diff
@@ -66,7 +66,7 @@ Indexes the location at the specified resolution, returning the index of the cel
 
 Returns 0 on error.
```
dfellis, line 67: This is also incorrect now, right?

isaacbrodsky: Yes, I will update this to mention 0/E_SUCCESS

dfellis (approved):

website/docs/api/hierarchy.mdx
```diff
@@ -286,48 +284,46 @@ Returns 0 on success.
...
 </TabItem>
 </Tabs>
 
-Uncompacts the set `compactedSet` of indexes to the resolution `res`. `h3Set` must be at least of size `maxUncompactSize(compactedSet, numHexes, res)`.
-
-Returns 0 on success.
+Uncompacts the set `compactedSet` of indexes to the resolution `res`. `cellSet` must be at least of size `maxUncompactSize(compactedSet, numCells, res)`.
 
-## maxUncompactSize
+## maxuncompactCellsSize
```
nrabinowitz, line 326: Nit: Capitalization here is incorrect, but this is probably moot due to #413

nrabinowitz (approved):

Issue #431, opened 2021-02-10
Move some user-facing definitions from `h3Index.h` to `h3api.h`?


ajfriend:
Would it make sense to move the `H3_NULL` definition and the `compact` return code definitions from `h3index.h` to `h3api.h`? This would make them more visible/accessible to the user.

https://github.com/uber/h3/blob/b04c78fa08adbd6857aa6f4b2a0f8e08a4ce2ead/src/h3lib/include/h3Index.h#L167-L180

Since https://github.com/uber/h3/blob/master/dev-docs/RFCs/v4.0.0/error-handling-rfc.md is in progress, maybe we'd really only need to consider moving `H3_NULL`?

isaacbrodsky:
Since the define is part of our public API, I think it makes sense for it to be in the public API file.

ajfriend:
`H3_NULL` got done in #413. Still need to move over the compact return codes.

ajfriend:
Done with #465

PR #465, opened 2021-05-18
move return codes for compactCells to h3api.h.in

ajfriend:
To finish off #431


dfellis (approved):

isaacbrodsky (approved):
Looks good, the constants will of course be removed with moving to H3Error.

PR #468, opened 2021-05-20
Add error return codes to compactCells

isaacbrodsky:


website/docs/api/hierarchy.mdx
```diff
@@ -325,7 +325,7 @@ function example() {
 
 Uncompacts the set `compactedSet` of indexes to the resolution `res`. `h3Set` must be at least of size `uncompactCellsSize(compactedSet, numHexes, res)`.
 
-Returns 0 on success.
+Returns 0 (`E_SUCCESS`) on success.
```
nrabinowitz, line 328: `... or non-zero error code on failure`? Ideally we'd list out known errors here, linking to the error docs.

nrabinowitz (approved):

src/apps/testapps/testCompactCells.c
```diff
@@ -297,24 +295,23 @@ SUITE(compactCells) {
...
     }
 
     TEST(someHexagon) {
         H3Index origin;
         setH3Index(&origin, 1, 5, 0);
 
-        int64_t childrenSz = H3_EXPORT(uncompactCellsSize)(&origin, 1, 2);
+        int64_t childrenSz;
+        t_assertSuccess(
+            H3_EXPORT(uncompactCellsSize)(&origin, 1, 2, &childrenSz));
```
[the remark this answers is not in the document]
isaacbrodsky: This is rouhgly my usage. There may be some cases where I use `t_assertSuccess` for logic under test, but usually not so as to preserve the message. I think it can be treated as just a convenient way to write the `t_assert` invocation, as the messages are fairly opaque anyways (unless the message is used to clarify what is going on in the test)

website/docs/api/hierarchy.mdx
isaacbrodsky: Still thinking about how to add the known failure codes in here, but I agree it would be a useful addition.

src/h3lib/lib/h3Index.c
```diff
@@ -471,21 +471,21 @@ int H3_EXPORT(compactCells)(const H3Index* h3Set, H3Index* compactedSet,
...
-                              const int64_t numCompacted, H3Index* outSet,
-                              const int64_t numOut, const int res) {
+H3Error H3_EXPORT(uncompactCells)(const H3Index* compactedSet,
+                                  const int64_t numCompacted, H3Index* outSet,
+                                  const int64_t numOut, const int res) {
     int64_t i = 0;
 
     for (int64_t j = 0; j < numCompacted; j++) {
-        if (!_hasChildAtRes(compactedSet[j], res)) return -2;
+        if (!_hasChildAtRes(compactedSet[j], res)) return E_FAILED;
```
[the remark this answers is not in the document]
isaacbrodsky: I used the E_RES_MISMATCH for these cases, since it covers the H3Index res being invalid. There is a bit of overlap between E_RES_MISMATCH (the index resolution is out of range) and E_RES_DOMAIN (the resolution argument itself is out of range), which is tricky.

website/docs/core-library/usage.md
```diff
@@ -7,7 +7,7 @@ slug: /core-library/usage
 
 ## API Versioning
 
-The public API of the H3 Core Library is defined in the file `h3api.h`. The functions defined in h3api.h adhere to [Semantic Versioning](http://semver.org/).
+The public API of the H3 Core Library is defined in the file [`h3api.h`](https://github.com/uber/h3/blob/master/src/h3lib/include/h3api.h.in). The functions defined in `h3api.h` adhere to [Semantic Versioning](http://semver.org/).
```
dfellis, line 10: Since we're linking to the `.in` file that we preprocess (not using the C preprocessor) maybe we need to explain that a bit?

dfellis (approved):

PR #469, opened 2021-05-20
Remove Travis CI

isaacbrodsky:
Completes the migration to Github Actions. Note that these last two uses of Travis CI were not ported over: FOSSA, and tests on non-x64 architectures. We would welcome adding back tests for non-x64 architectures, as we'd still like to support them (I've seen things like this done with QEMU on Github Actions)


isaacbrodsky:
https://github.com/sqlalchemy/sqlalchemy/pull/5480/files might be an option for getting multiarch testing back

nrabinowitz (approved):

dfellis (approved):

PR #471, opened 2021-05-21
Swallow some errors from the live code editor, format as JSON

isaacbrodsky:
Workaround certain crashes in react-live that happen if bad code blocks are passed in.

I did notice that we may have issues supporting live code editor for both 3.x and 4.x on the website at once - perhaps once we switch to 4.x as "live" we will need to disable live code on the old site. I think that would be a fair workaround.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -97,7 +97,7 @@ module.exports = {
           },
           versions: {
             current: {
-              label: '4.x (in development)'
+              label: '4.x (alpha)'
             },
           },
         },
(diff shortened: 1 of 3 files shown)


dfellis (approved):
It should be possible to depend on multiple versions of `h3-js` at the same time and not need to disable the live code editor for 3.x, but I also think it's not a priority once we have 4.x ready.

website/docusaurus.config.js
```diff
@@ -97,7 +97,7 @@ module.exports = {
           },
           versions: {
             current: {
-              label: '4.x'
+              label: '4.x (in development)'
```
nrabinowitz, line 100: Maybe "alpha"?

nrabinowitz (approved):
This seems good as a stopgap; we could probably do a little better in terms of error reporting, etc, but seems fine for now.

Issue #484, opened 2021-06-10
Possible vulnarability in outdated css-what, dns-packet


michaelbrichko:
**Description:**
The css-what package before 5.0.1 for Node.js does not ensure that attribute parsing has Linear Time Complexity relative to the size of the input.

This affects the package dns-packet before versions 1.3.2 and 5.2.2. It creates buffers with allocUnsafe and does not always fill them before forming network packets. This can expose internal application memory over unencrypted network when querying crafted invalid domain names.

**Needed fix:**
Upgrade css-what from 3.4.2 to 5.0.1 to fix the vulnerability.
Upgrade css-what from 2.1.3 to 5.0.1 to fix the vulnerability.
Upgrade dns-packet from 1.3.1 to 1.3.2 to fix the vulnerability.

isaacbrodsky:
Thanks for the note. I created #485 to bump the dependencies on our static website.

michaelbrichko:
Thank you !

PR #485, opened 2021-06-11
Website: upgrade Docusaurus to 2.0.0-beta.0

isaacbrodsky:
Should address #484 for dns-packet, does not appear to address for css-what as that version seems to be brought in by Docusaurus.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -14,9 +14,9 @@
     "write-heading-ids": "docusaurus write-heading-ids"
   },
   "dependencies": {
-    "@docusaurus/core": "2.0.0-alpha.73",
-    "@docusaurus/preset-classic": "2.0.0-alpha.73",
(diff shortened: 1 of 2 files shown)


dfellis:
So, we can try using [`"resolutions"` in the `package.json`](https://stackoverflow.com/a/41082766) to update `css-what`. I don't know if it'll work, but since it's a static site, once we build locally if it does work, then we're fine?

isaacbrodsky:
> So, we can try using [`"resolutions"` in the `package.json`](https://stackoverflow.com/a/41082766) to update `css-what`. I don't know if it'll work, but since it's a static site, once we build locally if it does work, then we're fine?

Since the site is static and the vulnerability seems to be denial of service on the build side, I don't see a significant risk to us.

ajfriend (approved):

PR #488, opened 2021-06-24
Website: Add top level nav

isaacbrodsky:
Add top level navigation to the documentation site. One of the most requested pages (don't have analytics to tell us which other ones are highly visited) is the resolution reference table, so that one needs to be top level and easy to access. I guessed about the other ones that would be important to a reader.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -22,9 +22,26 @@ module.exports = {
       },
       items: [
         {
-          to: 'docs/',
-          activeBasePath: 'docs',
-          label: 'Docs',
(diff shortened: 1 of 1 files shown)


dfellis (approved):

nrabinowitz (approved):
Resolutions seems like an odd one for the top level, but seems OK for now.

ajfriend (approved):

isaacbrodsky:
> Resolutions seems like an odd one for the top level, but seems OK for now.

I was trying to come up with a short descriptor of it. The table is really commonly used as a reference.

PR #494, opened 2021-07-05
CMake: fix installation of filters on iOS

SpaceIm:
bundle targets produce a configuration error is they don't have a BUNDLE DESTINATION in install target since https://cmake.org/cmake/help/latest/policy/CMP0006.html.
By removing RUNTIME, we set DESTINATION for all types.


prince-chrismc (approved):

isaacbrodsky (approved):

dfellis (approved):

isaacbrodsky:
Thanks for submitting this fix!

PR #495, opened 2021-07-05
change H3_HEXAGON_MODE to H3_CELL_MODE

ajfriend:
`H3_CELL_MODE` is the more correct name due to pentagons.

### DIFF
--- a/src/h3lib/include/constants.h
+++ b/src/h3lib/include/constants.h
@@ -78,7 +78,7 @@
 #define NUM_PENTAGONS 12
 
 /** H3 index modes */
-#define H3_HEXAGON_MODE 1
+#define H3_CELL_MODE 1
 #define H3_DIRECTEDEDGE_MODE 2
 #define H3_EDGE_MODE 3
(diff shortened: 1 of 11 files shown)


src/h3lib/include/constants.h
```diff
@@ -78,7 +78,7 @@
 #define NUM_PENTAGONS 12
 
 /** H3 index modes */
-#define H3_HEXAGON_MODE 1
+#define H3_CELL_MODE 1
 #define H3_DIRECTEDEDGE_MODE 2
 #define H3_EDGE_MODE 3
 #define H3_VERTEX_MODE 4
```
ajfriend, line 84: Should these maybe be exposed in `h3api.h`?

nrabinowitz (approved):

src/h3lib/include/constants.h
isaacbrodsky: Are these returned or accepted by the API anywhere? Definitely if so.

isaacbrodsky (approved):

PR #497, opened 2021-07-08
First part of unified `h3` command

dfellis:
Only two functions implemented so far, but this demonstrates the potential pattern possible. This makes the sub-command names case-insensitive and forces sub-commands to be in explicit locations in the `argv` list. (Not just hardwired to `argv[1]` in case we ever need sub-commands of sub-commands).

Work to tackle:

- All of the rest of the public H3 functions as sub-commands
- KML output (and input?) where appropriate
- GeoJSON output (and input?) where appropriate

These can be in follow-up PRs, I hope?

### DIFF
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2021 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/** @file
+ * @brief cli app that exposes most of the H3 C library for scripting
+ *
+ *  See `h3 --help` for usage.
+ */
+
+#include <string.h>
+
+#include "args.h"
+#include "h3Index.h"
+#include "utility.h"
+
+bool has(char *subcommand, int argc, char *argv[]) {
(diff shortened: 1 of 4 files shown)


ajfriend:
This looks great! One question though. Do we have a way to add tests for these commands?

src/apps/filters/h3.c
```diff
@@ -0,0 +1,144 @@
...
+    Arg *args[] = {&helpArg, &cellToLatLngArg, &latLngToCellArg};
+    const char *helpText =
+        "Please use one of the subcommands listed to perform an H3 "
+        "calculation. Use h3 <SUBCOMMAND> --help for details on the usage of "
+        "any subcommand.";
+    return parseArgs(argc, argv, 3, args, &helpArg, helpText);
+}
+
+int main(int argc, char *argv[]) {
+    if (has("cellToLatLng", 1, argv) && cellToLatLngCmd(argc, argv)) {
```
isaacbrodsky, line 133: Possible out of range read if argc = 1. (Calls `strcasecmp` with `argv[1] = 0`)?

dfellis:
> This looks great! One question though. Do we have a way to add tests for these commands?

It would be integration testing (from the perspective of CMake) that we need to write. I have used [shellspec](https://github.com/shellspec/shellspec) to do this relatively cleanly in other projects. Do we want to potentially use that here but have it actively needed only if you run something like `make bdd` so most users don't have to download that?

Thoughts @isaacbrodsky @ajfriend @nrabinowitz ? (In my own usage, [I have the Makefile acquire the version of shellspec I want to use as a dependency for running the BDD tests](https://github.com/alantech/alan/blob/main/Makefile#L44-L45) so it could also be done without requiring changes to setup for github actions or in local development.)

src/apps/filters/h3.c
dfellis: This is true. I could have a simple `if (argc <= 1)` block that prints help and exits immediately to avoid this.

isaacbrodsky (approved):

isaacbrodsky:
I don't have any concerns with using an additional tool for testing the CLI. Using shellspec seems reasonable.

ajfriend (approved):

PR #498, opened 2021-07-15
Fix potential segfault in gridDisk due to invalid digit

isaacbrodsky:
h3NeighborRotations would index off the end of the NEW_DIGIT_II table (and others) if the input index had invalid digits in it. This can result in a crash.

Detected via AFL fuzzer and then narrowed down what was happening with UBSan. The segfault itself only reproduces with certain build configurations.

### DIFF
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -346,7 +346,10 @@ H3Index h3NeighborRotations(H3Index origin, Direction dir, int *rotations) {
         } else {
             Direction oldDigit = H3_GET_INDEX_DIGIT(out, r + 1);
             Direction nextDir;
-            if (isResolutionClassIII(r + 1)) {
+            if (oldDigit == INVALID_DIGIT) {
+                // Only possible on invalid input
+                return H3_NULL;
+            } else if (isResolutionClassIII(r + 1)) {
                 H3_SET_INDEX_DIGIT(out, r + 1, NEW_DIGIT_II[oldDigit][dir]);
                 nextDir = NEW_ADJUSTMENT_II[oldDigit][dir];
             } else {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The public API of this library consists of the functions declared in file
 
 ## [Unreleased]
 
+### Fixed
+- `gridDisk` of invalid indexes should not crash. (#498)
+
 ### Added
 - Vertex mode and associated functions:
     - `cellToVertex(cell, vertexNum)`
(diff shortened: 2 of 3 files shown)


src/h3lib/lib/algos.c
```diff
@@ -346,7 +346,10 @@ H3Index h3NeighborRotations(H3Index origin, Direction dir, int *rotations) {
         } else {
             Direction oldDigit = H3_GET_INDEX_DIGIT(out, r + 1);
             Direction nextDir;
-            if (isResolutionClassIII(r + 1)) {
+            if (oldDigit == INVALID_DIGIT) {
```
dfellis, line 349: I am wondering if expanding the size of the constant matrices like `NEW_DIGIT_II` and having an invalid detection at the end would be a bit more performant due to less branch code. But since this is already a branch it probably has only a small impact?

(So not a blocker, but it might be worthwhile to make a branch less version of this function since it is often in the deepest parts of looping code.)

dfellis (approved):

nrabinowitz (approved):

src/h3lib/lib/algos.c
isaacbrodsky: I'm not exactly sure how invalid detection would work in this case without doing a check for the digit being `7` here, or calling `h3IsValid`.

Further optimizing this function is a good idea since it is performance critical to `gridDisk`.

dfellis: So it's a hunch, but because of CPU pipelining, I think something very much like that would be faster.

Get rid of the `isResolutionClassIII` check by just having pointers set at the beginning of the loop that is toggled between `NEW_DIGIT_II/NEW_ADJUSTMENT_II` and `NEW_DIGIT_III/NEW_ADJUSTMENT_III` each time `r--` is also run, eliminating that conditional check.

Then the `INVALID_DIGIT` check can be turned into something like:

```c
// before the while loop
byte foundInvalidDigit = 0;

...

// immediately after `oldDigit` is declared
foundInvalidDigit |= oldDigit == INVALID_DIGIT;

... // the rest of the loop continues on, with the tables having extra (invalid) rows for the invalid digit

// while loop is done
if (foundInvalidDigit) return H3_NULL; // Could have returned this earlier, but no need to optimize for this case
```

I might also change up `while (true)` to `while (r > -1)` and put the final part of the top-level conditional outside of the loop, but that might not have any measurable impact on performance as currently there's an unconditional jump at the end of the loop and then a conditional jump, while this would reduce it to just a conditional jump, but unconditional jumps are cheap to schedule for the CPU.

PR #500, opened 2021-07-16
Backport fix potential segfault in kRing due to invalid digit (#498)

isaacbrodsky:
Backport #498 to `stable-3.x`, making release 3.7.2.

Note we'll need to update the changelog on `master` to also reflect this version.


isaacbrodsky:
Unfortunately we do not get CI on this, I think because we did not previously have a stable-3.x branch and so did not enable CI for it.

PR #501, opened 2021-07-16
Add changelog entry for 3.7.2, enable CI on stable-*

isaacbrodsky:
Changelog entry for #500.


dfellis (approved):

.github/workflows/test-linux.yml
```diff
@@ -2,9 +2,9 @@ name: test-linux
 
 on:
   push:
-    branches: [master]
+    branches: [master, stable-*]
```
nrabinowitz, line 5: For my info: We decided on `stable-` vs `release-` or `maintenance-`?

nrabinowitz (approved):

.github/workflows/test-linux.yml
isaacbrodsky: That wasn't discussed, but that was the name I chose when I made the branch. If you have a preference I can change that before merging today.

nrabinowitz: No preference, just clarifying. Not sure what other projects use for this.

PR #502, opened 2021-07-27
Add test driver for h3-pg#56

isaacbrodsky:
Adds a test driver for bytesandbrains/h3-pg#56 to the core library for debugging this type of polygon that currently has a very high estimation.


nrabinowitz:
Not sure I'm following the goal of this PR - the test behavior seems correct. Is the idea that we'll have some strategy to identify cases like this in the future and allocate less memory?

isaacbrodsky:
> Not sure I'm following the goal of this PR - the test behavior seems correct. Is the idea that we'll have some strategy to identify cases like this in the future and allocate less memory?

Yes, this is to help test those cases in the C library, and also ensure we don't regress to requesting more memory in cases like this.

Aside: We may want to consider partial polyfill approaches that would allow us to provide a stream (or iterator) of results to the caller rather than requiring pre-allocating all the memory.

isaacbrodsky:
Note also that the Win32 test for this actually fails with a segfault, so this does currently fail in some configurations.

PR #503, opened 2021-08-09
Add error return codes to h3Index.c (part 1)

isaacbrodsky:
Applies the error return codes RFC to h3Index.c with the exception of some commonly used functions that return true/false.

### DIFF
--- a/src/apps/testapps/testCellToChildren.c
+++ b/src/apps/testapps/testCellToChildren.c
@@ -63,11 +63,17 @@ static void assertSetsEqual(H3Index *set1, int len1, H3Index *set2, int len2) {
     assertSubset(set2, len2, set1, len1);
 }
 
-static void checkChildren(H3Index h, int res, H3Index *expected,
-                          int numExpected) {
-    int64_t numChildren = H3_EXPORT(cellToChildrenSize)(h, res);
+static void checkChildren(H3Index h, int res, H3Error expectedError,
+                          H3Index *expected, int numExpected) {
+    int64_t numChildren;
+    H3Error numChildrenError =
+        H3_EXPORT(cellToChildrenSize)(h, res, &numChildren);
+    t_assert(numChildrenError == expectedError, "Expected error code");
+    if (expectedError != E_SUCCESS) {
+        return;
+    }
     H3Index *children = calloc(numChildren, sizeof(H3Index));
-    H3_EXPORT(cellToChildren)(h, res, children);
+    t_assertSuccess(H3_EXPORT(cellToChildren)(h, res, children));
 
     assertSetsEqual(children, numChildren, expected, numExpected);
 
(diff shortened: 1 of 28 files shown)


dfellis (approved):

src/apps/testapps/testCellToChildren.c
```diff
@@ -63,11 +63,17 @@ static void assertSetsEqual(H3Index *set1, int len1, H3Index *set2, int len2) {
...
-    int64_t numChildren = H3_EXPORT(cellToChildrenSize)(h, res);
+static void checkChildren(H3Index h, int res, H3Error expectedError,
+                          H3Index *expected, int numExpected) {
+    int64_t numChildren;
+    H3Error numChildrenError =
+        H3_EXPORT(cellToChildrenSize)(h, res, &numChildren);
+    t_assert(numChildrenError == expectedError, "Expected error code");
+    if (expectedError != E_SUCCESS) {
+        return;
+    }
```
nrabinowitz, line 74: Does this imply that there are no errors we can test that won't be returned by `cellToChildrenSize`? I.e. there are no errors in `cellToChildren` that aren't in `cellToChildrenSize`?

src/apps/testapps/testH3Index.c
```diff
@@ -168,24 +168,26 @@ SUITE(h3Index) {
...
 
     TEST(stringToH3) {
-        t_assert(H3_EXPORT(stringToH3)("") == 0, "got an index from nothing");
-        t_assert(H3_EXPORT(stringToH3)("**") == 0, "got an index from junk");
-        t_assert(
-            H3_EXPORT(stringToH3)("ffffffffffffffff") == 0xffffffffffffffff,
-            "failed on large input");
+        H3Index h3;
+        t_assert(H3_EXPORT(stringToH3)("", &h3) == E_FAILED,
+                 "got an index from nothing");
```
nrabinowitz, line 186: Nit:
```suggestion
                 "no index from nothing");
```

src/apps/testapps/testH3Index.c
```diff
@@ -168,24 +168,26 @@ SUITE(h3Index) {
...
+        t_assert(H3_EXPORT(stringToH3)("**", &h3) == E_FAILED,
+                 "got an index from junk");
```
nrabinowitz, line 188:
```suggestion
                 "no index from junk");
```

src/apps/testapps/testH3Index.c
```diff
@@ -168,24 +168,26 @@ SUITE(h3Index) {
...
+        t_assertSuccess(H3_EXPORT(stringToH3)("ffffffffffffffff", &h3));
+        t_assert(h3 == 0xffffffffffffffff, "failed on large input");
```
nrabinowitz, line 190:
```suggestion
        t_assert(h3 == 0xffffffffffffffff, "got expected on large input");
```
Clearly we have some inconsistencies on how we write our assertion messages (I usually prefer the positive "got expected" or the prescriptive "must get expected"; I don't really like the negative "here's what went wrong", as it doesn't explain expected behavior well)

nrabinowitz (approved):

src/apps/testapps/testCellToChildren.c
isaacbrodsky: `cellToChildren` always returns `E_SUCCESS` right now, so that is correct

PR #505, opened 2021-08-13
Add error return codes to gridDisk functions

isaacbrodsky:
Adds the ability for the gridDisk series of functions (but not `maxGridDiskSize`, which may need to be updated to return int64_t) to return errors.

### DIFF
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -41,9 +41,6 @@
  * Return codes from gridDiskUnsafe and related functions.
  */
 
-#define HEX_RANGE_SUCCESS 0
-#define HEX_RANGE_PENTAGON 1
-#define HEX_RANGE_K_SUBSEQUENCE 2
 #define MAX_ONE_RING_SIZE 7
 #define HEX_HASH_OVERFLOW -1
 #define POLYGON_TO_CELLS_BUFFER 12
@@ -176,8 +173,8 @@ int H3_EXPORT(maxGridDiskSize)(int k) { return 3 * k * (k + 1) + 1; }
  * @param  k        k >= 0
  * @param  out      zero-filled array which must be of size maxGridDiskSize(k)
  */
-void H3_EXPORT(gridDisk)(H3Index origin, int k, H3Index *out) {
-    H3_EXPORT(gridDiskDistances)(origin, k, out, NULL);
+H3Error H3_EXPORT(gridDisk)(H3Index origin, int k, H3Index *out) {
+    return H3_EXPORT(gridDiskDistances)(origin, k, out, NULL);
 }
 
 /**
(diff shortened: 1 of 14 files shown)


isaacbrodsky:
There may be some unreachable branches in this PR, I will see if I can get coverage on them or otherwise deduplicate some of the error handling.

src/h3lib/lib/algos.c
```diff
@@ -212,14 +209,20 @@ void H3_EXPORT(gridDiskDistances)(H3Index origin, int k, H3Index *out,
             distances = H3_MEMORY(calloc)(maxIdx, sizeof(int));
             if (!distances) {
                 // TODO: Return an error code when this is not void
```
dfellis, line 211: Should delete the TODO here :)

src/h3lib/lib/algos.c
```diff
@@ -1079,13 +1087,14 @@ void _vertexGraphToLinkedGeo(VertexGraph *graph, LinkedGeoPolygon *out) {
...
-void H3_EXPORT(h3SetToLinkedGeo)(const H3Index *h3Set, const int numHexes,
-                                 LinkedGeoPolygon *out) {
+H3Error H3_EXPORT(h3SetToLinkedGeo)(const H3Index *h3Set, const int numHexes,
+                                    LinkedGeoPolygon *out) {
     VertexGraph graph;
     h3SetToVertexGraph(h3Set, numHexes, &graph);
     _vertexGraphToLinkedGeo(&graph, out);
     // TODO: The return value, possibly indicating an error, is discarded here -
     // we should use this when we update the API to return a value
     normalizeMultiPolygon(out);
```
dfellis, line 1097: Should this function also be updated so it can return the error here?

isaacbrodsky:
We also need to update the docs for gridDisk.

isaacbrodsky:
This should be ready for review. I began updating some additional functions in `traversal.mdx` while doing this.

src/h3lib/lib/algos.c
```diff
@@ -395,27 +408,28 @@ H3Index h3NeighborRotations(H3Index origin, Direction dir, int *rotations) {
...
                     // Rotate out of the deleted k subsequence
                     // We also need an additional change to the direction we're
                     // moving in
-                    out = _h3Rotate60cw(out);
+                    working = _h3Rotate60cw(working);
                     *rotations = *rotations + 5;
                 } else {
                     // Should never occur
-                    return H3_NULL;  // LCOV_EXCL_LINE
+                    return E_FAILED;  // LCOV_EXCL_LINE
```
nrabinowitz, line 426: I suspect that the coverage drops are because this is the only failure case for this function, and we've added a few `if (neighborResult != E_SUCCESS)` branches after calls to this function.

It looks like we might be able to force failure here with an invalid index, with an `INVALID_DIGIT` setting as the leading non-zero digit?

src/h3lib/lib/algos.c
```diff
@@ -640,19 +662,23 @@ int H3_EXPORT(gridRingUnsafe)(H3Index origin, int k, H3Index *out) {
...
     }
 
     for (int ring = 0; ring < k; ring++) {
-        origin = h3NeighborRotations(origin, NEXT_RING_DIRECTION, &rotations);
+        H3Error neighborResult = h3NeighborRotations(
+            origin, NEXT_RING_DIRECTION, &rotations, &origin);
+        if (neighborResult) {
+            return neighborResult;
+        }
         if (origin == 0) {  // LCOV_EXCL_BR_LINE
```
nrabinowitz, line 674: Nit: `H3_NULL`? Also, would `h3NeighborRotations` ever set `out` to `H3_NULL` now, or is this check meaningless?

nrabinowitz (approved):

src/h3lib/lib/algos.c
```diff
@@ -395,27 +406,28 @@ H3Index h3NeighborRotations(H3Index origin, Direction dir, int *rotations) {
                 // base cell.
                 if (oldLeadingDigit == CENTER_DIGIT) {
                     // Undefined: the k direction is deleted from here
-                    return H3_NULL;
+                    return E_SUCCESS;
```
isaacbrodsky, line 409: This doesn't look right - `out` is not set in this case

dfellis (approved):

PR #518, opened 2021-09-27
Add benchmarkIsValidCell.c

ajfriend:
Adding in benchmarks before making changes to isValidCell (#496).

### DIFF
--- a/src/apps/benchmarks/benchmarkIsValidCell.c
+++ b/src/apps/benchmarks/benchmarkIsValidCell.c
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2021 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "benchmark.h"
+#include "h3api.h"
+
+// return a pentagon at the given resolution
+H3Index pentagonAtRes(int res) {
+    H3Index p = 0x80c3fffffffffff;  // res 0 pentagon
+    p = H3_EXPORT(cellToCenterChild)(p, res);
+
+    return p;
+}
+
+int parentRes;
+int childRes;
+
+H3Index p;
+int64_t N;
+H3Index *cells;
+
+BEGIN_BENCHMARKS();
+
+// pentagon 8->14
+//
+// Starting with a parent *pentagon* cell of resolution 8,
+// loop through all its children at resolution 14, applying `isValidCell`
+parentRes = 8;
+childRes = 14;
+
+p = pentagonAtRes(parentRes);
+H3_EXPORT(cellToChildrenSize)(p, childRes, &N);
+
+cells = calloc(N, sizeof(H3Index));
+H3_EXPORT(cellToChildren)(p, childRes, cells);
+
+BENCHMARK(pentagonChildren_8_14, 1000, {
+    for (int64_t i = 0; i < N; i++) {
+        H3_EXPORT(isValidCell)(cells[i]);
+    }
+});
+free(cells);
+
+// pentagon 2->8
+//
+// Starting with a parent *pentagon* cell of resolution 2,
+// loop through all its children at resolution 8, applying `isValidCell`
+parentRes = 2;
+childRes = 8;
+p = pentagonAtRes(parentRes);
+H3_EXPORT(cellToChildrenSize)(p, childRes, &N);
+
+cells = calloc(N, sizeof(H3Index));
+H3_EXPORT(cellToChildren)(p, childRes, cells);
+
+BENCHMARK(pentagonChildren_2_8, 1000, {
+    for (int64_t i = 0; i < N; i++) {
+        H3_EXPORT(isValidCell)(cells[i]);
+    }
+});
+free(cells);
+
+END_BENCHMARKS();
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,6 +226,7 @@ set(OTHER_SOURCE_FILES
     src/apps/benchmarks/benchmarkGridPathCells.c
     src/apps/benchmarks/benchmarkDirectedEdge.c
     src/apps/benchmarks/benchmarkVertex.c
+    src/apps/benchmarks/benchmarkIsValidCell.c
     src/apps/benchmarks/benchmarkH3Api.c)
 
 set(ALL_SOURCE_FILES
@@ -657,6 +658,7 @@ if(BUILD_BENCHMARKS)
     add_h3_benchmark(benchmarkGridPathCells src/apps/benchmarks/benchmarkGridPathCells.c)
     add_h3_benchmark(benchmarkDirectedEdge src/apps/benchmarks/benchmarkDirectedEdge.c)
     add_h3_benchmark(benchmarkVertex src/apps/benchmarks/benchmarkVertex.c)
+    add_h3_benchmark(benchmarkIsValidCell src/apps/benchmarks/benchmarkIsValidCell.c)
     add_h3_benchmark(benchmarkH3SetToLinkedGeo src/apps/benchmarks/benchmarkH3SetToLinkedGeo.c)
     add_h3_benchmark(benchmarkPolygonToCells src/apps/benchmarks/benchmarkPolygonToCells.c)
     if(ENABLE_REQUIRES_ALL_SYMBOLS)


dfellis (approved):
Can we also get a branch that checks invalid data, possibly of different types of incorrectness?

I think we should optimize the valid path, but it can still be useful to see how big the performance difference is, especially for those wanting to put this check in a hot loop.

ajfriend:
> Can we also get a branch that checks invalid data, possibly of different types of incorrectness?
> 
> I think we should optimize the valid path, but it can still be useful to see how big the performance difference is, especially for those wanting to put this check in a hot loop.

Makes sense. Any thoughts on what kinds of incorrectness it would make sense to check for? There are lots of ways to be incorrect, so I'm not sure off the top of my head what a benchmark suite with "good" coverage might look like.

dfellis:
> > Can we also get a branch that checks invalid data, possibly of different types of incorrectness?
> > I think we should optimize the valid path, but it can still be useful to see how big the performance difference is, especially for those wanting to put this check in a hot loop.
> 
> Makes sense. Any thoughts on what kinds of incorrectness it would make sense to check for? There are lots of ways to be incorrect, so I'm not sure off the top of my head what a benchmark suite with "good" coverage might look like.

The kinds of invalid cells:

- sign bit set
- invalid mode selected
- invalid base cell
- invalid child selection
  - setting a '7' before the chosen resolution
  - setting the deleted child of a pentagon

I don't which would have what effect, though I suspect the last two would be the slowest to recognize so they might be the ones to prioritize?

ajfriend:
> I don't which would have what effect, though I suspect the last two would be the slowest to recognize so they might be the ones to prioritize?

They'd be recognized in the order you stated. Thinking about it now, a benchmark like this would basically be on "doing only the first n out of N steps" of the validation. The "valid data" case then sort of includes all of the "invalid data" cases implicitly (assuming `isValidCell` works correctly, but that's covered by the test suite).

I'm now leaning towards just keeping it like we have it now, with just the "valid data" tests, since the speed of identifying invalid data (of any of the types listed above) should basically correspond how quickly the function runs through the full stage of validity tests. I guess we would miss the exact n/N breakdown, but I'm not sure that's providing much more information--unless we have a specific hypothesis we'd like to look into.

dfellis:
> > I don't which would have what effect, though I suspect the last two would be the slowest to recognize so they might be the ones to prioritize?
> 
> They'd be recognized in the order you stated. Thinking about it now, a benchmark like this would basically be on "doing only the first n out of N steps" of the validation. The "valid data" case then sort of includes all of the "invalid data" cases implicitly (assuming `isValidCell` works correctly, but that's covered by the test suite).
> 
> I'm now leaning towards just keeping it like we have it now, with just the "valid data" tests, since the speed of identifying invalid data (of any of the types listed above) should basically correspond how quickly the function runs through the full stage of validity tests. I guess we would miss the exact n/N breakdown, but I'm not sure that's providing much more information--unless we have a specific hypothesis we'd like to look into.

Mixer valid and invalid cells may be slower than either 100% valid or 100% invalid because of branch prediction failures?

src/apps/benchmarks/benchmarkIsValidCell.c
```diff
@@ -0,0 +1,75 @@
...
+// loop through all its children at resolution 8, applying `isValidCell`
+parentRes = 2;
+childRes = 8;
+p = pentagonAtRes(parentRes);
+H3_EXPORT(cellToChildrenSize)(p, childRes, &N);
+
+cells = calloc(N, sizeof(H3Index));
+H3_EXPORT(cellToChildren)(p, childRes, cells);
+
+BENCHMARK(pentagonChildren_2_8, 1000, {
```
nrabinowitz, line 68: Why are we only benchmarking pentagon parent cells?

nrabinowitz:
FWIW - in my experience the most common case here is a mix of valid H3 indexes and nulls. So it might be worth benchmarking that, e.g. with alternating indexes. But I'm not sure the branch prediction failure question makes sense to benchmark - won't that vary widely over platforms, not to mention across bindings?

src/apps/benchmarks/benchmarkIsValidCell.c
ajfriend: Just an easy way to generate data. At least with pentagon parents, you get one pentagon child cell.

src/apps/benchmarks/benchmarkIsValidCell.c
```diff
@@ -0,0 +1,75 @@
...
+free(cells);
+
+// pentagon 2->8
+//
+// Starting with a parent *pentagon* cell of resolution 2,
+// loop through all its children at resolution 8, applying `isValidCell`
+parentRes = 2;
+childRes = 8;
+p = pentagonAtRes(parentRes);
+H3_EXPORT(cellToChildrenSize)(p, childRes, &N);
```
isaacbrodsky, line 63: Total nit but could consider doing all setup before all `BENCHMARK` macros. I don't expect it to effect results.

isaacbrodsky (approved):

isaacbrodsky:
> 
> 
> FWIW - in my experience the most common case here is a mix of valid H3 indexes and nulls. So it might be worth benchmarking that, e.g. with alternating indexes. But I'm not sure the branch prediction failure question makes sense to benchmark - won't that vary widely over platforms, not to mention across bindings?

I am also curious about performance across valid cells and `H3_NULL`.

I think the branch prediction question is applicable to our entire benchmarking approach. Both that we test the same data repeatedly, and that we don't have a performance reference.

src/apps/benchmarks/benchmarkIsValidCell.c
ajfriend: I'd also like to organize it that way, but I could't figure out how to do a problem setup for each benchmark without it being timed.

I'd like to setup and free any memory right before and after each timing, but I don't see an easy way to do that with the current benchmarking code. Is there a way I'm not seeing?

ajfriend:
Here are the numbers I got from running on my macbook air (so take them with a grain of salt). The last test is the same as the first, just to give an idea of how much the numbers can vary based on how hot my laptop is running.

```
build/bin/benchmarkIsValidCell
    -- pentagonChildren_8_14_null_100: 1844.567000 microseconds per iteration (1000 iterations)
    -- pentagonChildren_8_14: 1938.043000 microseconds per iteration (1000 iterations)
    -- pentagonChildren_2_8: 1663.274000 microseconds per iteration (1000 iterations)
    -- pentagonChildren_8_14_null_2: 1104.476000 microseconds per iteration (1000 iterations)
    -- pentagonChildren_8_14_null_10: 2010.307000 microseconds per iteration (1000 iterations)
    -- pentagonChildren_8_14_null_100: 1988.599000 microseconds per iteration (1000 iterations)
```

ajfriend:
I added the H3_NULL tests above. I kicked the other branch prediction benchmark questions to #520, just because I'd like to close this out and get back to #496.

ajfriend:
Run with all cells at resolution 4:

```
build/bin/benchmarkIsValidCell
	-- allCellsAtRes4: 12272.413000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14: 5825.150000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_2_8: 4668.005000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_2: 2515.056000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_10: 4762.763000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_100: 5794.260000 microseconds per iteration (1000 iterations)
```

nrabinowitz (approved):

ajfriend:
Run with the latest code:

```
build/bin/benchmarkIsValidCell
	-- pentagonChildren_2_8: 4460.855000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14: 4377.846000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_2: 2168.223000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_10: 4188.899000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_100: 4368.239000 microseconds per iteration (1000 iterations)
```

ajfriend:
And here are the latest benchmarks (still on my macbook air). Also, the previous benchmark numbers are off because I'd forgotten to `cmake -DCMAKE_BUILD_TYPE=Release ..`. That's why these numbers are lower.

```
build/bin/benchmarkIsValidCell
	-- pentagonChildren_2_8_func: 1636.138000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14: 2254.166000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_2: 1229.321000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_10: 2064.721000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_100: 2220.962000 microseconds per iteration (1000 iterations)
```

ajfriend:
```
build/bin/benchmarkIsValidCell
	-- pentagonChildren_2_8: 1602.169000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14: 1939.921000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_2: 1071.529000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_10: 1828.599000 microseconds per iteration (1000 iterations)
	-- pentagonChildren_8_14_null_100: 1878.061000 microseconds per iteration (1000 iterations)
```

PR #519, opened 2021-10-02
Polyfill modes RFC

isaacbrodsky:


dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,79 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+    }
+    // The polygon definitely does not contain the point, the expensive check
+    // is not needed.
+    return false;
+}
+```
+
+* Analytics
+
+There is no convenient function to index a polygon into the H3 grid for analytics purposes. Polygons that are small enough, or shaped in specific ways, will not be converted into points, requiring workarouns in order to use them.
```
nrabinowitz, line 39:
```suggestion
There is no convenient function to index a polygon into the H3 grid for analytics purposes. Polygons that are small enough or shaped in specific ways will not be converted into cells, requiring workarounds in order to use them.
```

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,79 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+* Spherical geometry consistency
+
+Most of the H3 library uses spherical geoemtry. For example, cell boundaries are spherical hexagons
+and pentagons. The `polyfill` function is different that it assumes Cartesian geometry. For
+consistency with the rest of the library, the `polyfill` functions should be able to use the same
+cell boundaries.
+
+* Very large polygons
+
+Polyfills of very large polygons requires allocating large blocks of memory, and spending large
```
nrabinowitz, line 50:
```suggestion
Polyfills of very large polygons require allocating large blocks of memory, and spending large
```

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,79 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+amounts of time in a library function without progress information being available to the caller.
+(To be determiend if this is in scope for this RFC or for another.)
```
nrabinowitz, line 52:
```suggestion
(To be determined if this is in scope for this RFC or for another.)
```

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,79 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+There is no convenient function to index a polygon into the H3 grid for analytics purposes. Polygons that are small enough, or shaped in specific ways, will not be converted into points, requiring workarouns in order to use them.
+
+* Spherical geometry consistency
+
+Most of the H3 library uses spherical geoemtry. For example, cell boundaries are spherical hexagons
+and pentagons. The `polyfill` function is different that it assumes Cartesian geometry. For
+consistency with the rest of the library, the `polyfill` functions should be able to use the same
+cell boundaries.
+
+* Very large polygons
```
nrabinowitz, line 48: Do you have suggestions here? E.g. `maxPolyfillSizes` could return an array of sizes, and `polyfillIncremental(batchNum, polygon, out)` could polyfill a batch? (How this would work in the lib is still a bit fuzzy, but I think it's possible)

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,79 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+```
+
+* Analytics
+
+There is no convenient function to index a polygon into the H3 grid for analytics purposes. Polygons that are small enough, or shaped in specific ways, will not be converted into points, requiring workarouns in order to use them.
+
+* Spherical geometry consistency
+
+Most of the H3 library uses spherical geoemtry. For example, cell boundaries are spherical hexagons
+and pentagons. The `polyfill` function is different that it assumes Cartesian geometry. For
```
nrabinowitz, line 44: Should this RFC refer to `polyfill` or `polygonToCells`?

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,79 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+
+Polyfills of very large polygons requires allocating large blocks of memory, and spending large
+amounts of time in a library function without progress information being available to the caller.
+(To be determiend if this is in scope for this RFC or for another.)
+
 ## Approaches
 
 *What are the various options to address this issue?*
 
+On an API level, we will need to expose the containment mode and spherical/Cartesian choice as
```
nrabinowitz, line 58: I think we might decide in v4 that polyfill is spherical, full stop, unless it's much slower this way. But I'm not sure there's a benefit to exposing both modes to the user, and it would mean a bunch of redundant code to check point inclusion and line intersection in both ways.

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
isaacbrodsky: I'm inconsistent in the term used, but we should do a pass making that consistent. `polyfill` might be useful as a term as the abstract concept of filling a polygon (regardless of inclusion mode, etc).

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
isaacbrodsky: I was considering if it would be possible to do this in an iterator but haven't looked into this more. We would need some allocation done, but perhaps that could be made relative to the number of points in the polygon rather than the number of cells. A simple polygon producing a large number of cells is related to the problem seen in bytesandbrains/h3-pg#56

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -77,14 +75,52 @@ Contains separate functions for [intersection](http://turfjs.org/docs/#booleanIn
...
+The same value used for `maxPolygonToCellsSize` must be used for the subsequent call to `polygonToCells`, just as the `GeoPolygon` and `res` must be the same.
+
+In bindings, this could be expressed using enums, for example:
+
+```python
+polygon_to_cells(polygon, res=res, cartesian=True, containment=h3.Containment.CENTROID)
+```
+
+```js
+polygonToCells(polygon, {res: res, cartesian: true, containment: h3.Containment.CENTROID})
```
dfellis, line 121: Nit: In JS, if a new object's property exactly matches a variable that will fill in that property, you can just list the property/variable name and it'll do the right thing. Eg:

```suggestion
polygonToCells(polygon, {res, cartesian: true, containment: h3.Containment.CENTROID})
```

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,115 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+/** @brief hexagons within the given geopolygon */
+DECLSPEC H3Error H3_EXPORT(polygonToCells)(const GeoPolygon *geoPolygon,
+                                           int res, uint32_t flags, H3Index *out);
+```
+
+`flags` would have the following possible bit layout:
+
+|       Bits | Meaning
+| ---------- | -------
+| 1-2 (LSB)  | If 0, containment mode centroid.<br>If 1, containment mode cover.<br>If 2, containment mode intersects.<br>3 is a reserved value.
```
nrabinowitz, line 108: I'd reserve more bits here, there are other possible approaches (e.g. > 50% containment). Got plenty of bits.

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
isaacbrodsky: I think it was raised offline that cartesian is useful when maps are drawn on top of web mercator maps.

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
isaacbrodsky: As in make the flags 64 bits?

nrabinowitz: No, as in make this bit section wider (say 3 bits instead of two) so that we don't constrain ourselves to 4 containment modes.

PR #537, opened 2021-11-11
[Docs] Add Conda Forge links to bindings

### LINKED ISSUE #355: Add conda links to h3geo.org
- https://github.com/conda-forge/h3-py-feedstock
- https://anaconda.org/conda-forge/h3-py

isaacbrodsky:
Fixes #355

I hope I understood correctly that the h3-py-feedstock repo is what builds the Conda package -- the anaconda.org page doesn't link back to Github.

### DIFF
--- a/website/docs/community/bindings.md
+++ b/website/docs/community/bindings.md
@@ -80,6 +80,7 @@ As a C library, bindings can be made to call H3 functions from different program
 
 - [uber/h3-py](https://github.com/uber/h3-py)
 - [DahnJ/H3-Pandas](https://github.com/DahnJ/H3-Pandas)
+- [h3-py on Conda Forge](https://anaconda.org/conda-forge/h3-py) ([conda-forge/h3-py-feedstock](https://github.com/conda-forge/h3-py-feedstock))
 
 ## R
 


website/docs/community/bindings.md
```diff
@@ -80,6 +80,7 @@ As a C library, bindings can be made to call H3 functions from different program
 
 - [uber/h3-py](https://github.com/uber/h3-py)
 - [DahnJ/H3-Pandas](https://github.com/DahnJ/H3-Pandas)
+- [h3-py on Conda Forge](https://anaconda.org/conda-forge/h3-py) ([conda-forge/h3-py-feedstock](https://github.com/conda-forge/h3-py-feedstock))
```
dfellis, line 83: We probably want to also sync this with the 3.x docs? There's some confusion about bindings not showing up in different API versions https://github.com/uber/h3/commit/14865690c8f68f39c4d16d1019d6b4817c77e90f (perhaps we just symlink the `bindings.md` between the two directories, or maybe even the entire `community` directory?

nrabinowitz: I wonder if it's possible to take this section out of the "versioned" docs? Only the API really needs to be versioned.

ajfriend:
I'm not sure we need a separate link for conda. It isn't a separate binding, just a separate packaging index of the same python binding.

We should certainly link to it, but we should do that on the h3-py GitHub repo readme and on the h3-py docs site. And I think we already do this.

PR #539, opened 2021-11-12
Update bindings.md

michaelbrichko:


dfellis:
@michaelbrichko after your comment, we discussed the situation and this PR: https://github.com/uber/h3/pull/538 should prevent that sort of mixup again in the future, so this one *should* be unnecessary once that one's deployed.

michaelbrichko:
ok

michaelbrichko:
By unnecessary you mean that 4.x will become default very soon ? At the moment it's default on main page (https://h3geo.org/) but if you go to bindings tab, it's 3.x

dfellis:
> By unnecessary you mean that 4.x will become default very soon ? At the moment it's default on main page (https://h3geo.org/) but if you go to bindings tab, it's 3.x

No, sorry, the other PR I linked should make it clear; we're just making the 3.x "community" pages a symlink to the 4.x ones, since they're identical.

4.0 has more work to go (and most of us no longer work for Uber, so this is much more of a volunteer project now in between our regular work duties).

isaacbrodsky:
I think this is no longer needed with #538 - please let me know if not the case.

PR #549, opened 2021-12-09
polygonToCells flags argument

isaacbrodsky:


src/h3lib/lib/algos.c
```diff
@@ -813,6 +817,58 @@ H3Error _getEdgeHexagons(const GeoLoop *geoloop, int64_t numHexagons, int res,
     return E_SUCCESS;
 }
 
+/**
+ * Returns true if the given cell is in the polygon using the containment
+ * mode flags.
+ */
+bool cellInPolygon(H3Index hex, const GeoPolygon *geoPolygon,
```
isaacbrodsky, line 824: Per discussion, this adds a few modes we know we can implement (ANY_VERTEX, ALL_VERTEXES) which shouldn't require any more complicated changes to the library right now.

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,118 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+There is no convenient function to index a polygon into the H3 grid for analytics purposes. Polygons that are small enough or shaped in specific ways will not be converted into cells, requiring workarounds in order to use them.
+
+* Spherical geometry consistency
+
+Most of the H3 library uses spherical geoemtry. For example, cell boundaries are spherical hexagons
+and pentagons. The `polyfill` function is different that it assumes Cartesian geometry. For
+consistency with the rest of the library, the `polyfill` functions should be able to use the same
+cell boundaries.
+
+Maintaing a Cartesian option is useful for cases where polygons have been drawn on a projected map
```
nrabinowitz, line 48:
```suggestion
Maintaining a Cartesian option is useful for cases where polygons have been drawn on a projected map
```

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
```diff
@@ -12,10 +12,118 @@ Our current polyfill algorithm allocates cells to polygons based on whether the
...
+/** @brief hexagons within the given geopolygon */
+DECLSPEC H3Error H3_EXPORT(polygonToCells)(const GeoPolygon *geoPolygon,
+                                           int res, uint32_t flags, H3Index *out);
+```
+
+`flags` would have the following possible bit layout:
+
+|       Bits | Meaning
+| ---------- | -------
+| 1-2 (LSB)  | If 0, containment mode centroid.<br>If 1, containment mode cover.<br>If 2, containment mode intersects.<br>3 is a reserved value.
```
nrabinowitz, line 111: As discussed previously, were we going to swap this around to put the cartesian/spherical at the LSB and allow this section more room to easily expand?

Also, are we planning to have boolean flags for each mode (i.e. each mode has one bit and we check LTR using simple AND checks) or are we setting aside a multi-bit section that we convert to an integer to check an enum value?

src/apps/testapps/testH3Memory.c
```diff
@@ -187,31 +187,31 @@ SUITE(h3Memory) {
         sfGeoPolygon.numHoles = 0;
 
         int64_t numHexagons;
-        t_assertSuccess(
-            H3_EXPORT(maxPolygonToCellsSize)(&sfGeoPolygon, 9, &numHexagons));
+        t_assertSuccess(H3_EXPORT(maxPolygonToCellsSize)(&sfGeoPolygon, 9, 0,
```
nrabinowitz, line 190: Nit: For clarity, I'd rather have a const like `FLAGS` used here

src/h3lib/lib/algos.c
```diff
@@ -37,9 +37,12 @@
...
 #include "vertexGraph.h"
 
-/*
- * Return codes from gridDiskUnsafe and related functions.
+/**
+ * TODO: Move to h3api.h.in
+ * TODO: Use correct codes from RFC
  */
+#define FLAG_ANY_VERTEX 1
+#define FLAG_ALL_VERTEXES 2
```
nrabinowitz, line 45: Note that the RFC doesn't actually include these flags, which are less rigorous than "contains" or "intersects". So that gives us minimum 5 modes at the outset.

Also, should this be an enum?

src/h3lib/lib/algos.c
```diff
@@ -813,6 +817,58 @@ H3Error _getEdgeHexagons(const GeoLoop *geoloop, int64_t numHexagons, int res,
     return E_SUCCESS;
 }
 
+/**
+ * Returns true if the given cell is in the polygon using the containment
+ * mode flags.
+ */
+bool cellInPolygon(H3Index hex, const GeoPolygon *geoPolygon,
+                   const BBox *bboxes, uint32_t flags) {
+    if ((flags & FLAG_ANY_VERTEX) == FLAG_ANY_VERTEX) {
```
nrabinowitz, line 826: Does this style of check make sense if we aren't using 1 bit per flag? If we are using 1 bit per flag, do we need the equality check? (Maybe this is a standard flag pattern I'm just not following?)

src/h3lib/lib/algos.c
```diff
@@ -813,6 +817,58 @@ H3Error _getEdgeHexagons(const GeoLoop *geoloop, int64_t numHexagons, int res,
...
+            if (!pointInsidePolygon(geoPolygon, bboxes,
+                                    &hexBoundary.verts[i])) {
+                return false;
+            }
+        }
+
+        LatLng hexCenter;
+        H3_EXPORT(cellToLatLng)(hex, &hexCenter);
+
+        if (pointInsidePolygon(geoPolygon, bboxes, &hexCenter)) {
```
nrabinowitz, line 856: Nit: This is faster than the boundary check, should it move above the boundary check and return false if the condition fails?

src/h3lib/lib/algos.c
```diff
@@ -813,6 +817,58 @@ H3Error _getEdgeHexagons(const GeoLoop *geoloop, int64_t numHexagons, int res,
...
+        for (int i = 0; i < hexBoundary.numVerts; i++) {
+            if (pointInsidePolygon(geoPolygon, bboxes, &hexBoundary.verts[i])) {
+                return true;
+            }
+        }
+
+        LatLng hexCenter;
+        H3_EXPORT(cellToLatLng)(hex, &hexCenter);
+
+        if (pointInsidePolygon(geoPolygon, bboxes, &hexCenter)) {
```
nrabinowitz, line 839: As below - this is faster than the boundary check, move to the beginning of the block?

src/h3lib/lib/polygon.c
```diff
@@ -83,3 +83,94 @@ bool pointInsidePolygon(const GeoPolygon *geoPolygon, const BBox *bboxes,
...
+                                           &bboxes[i + 1], boundary)) {
+                return false;
+            }
+        }
+    }
+
+    return contains;
+}
+
+bool boundaryIntersectsGeoLoop(const GeoLoop *loop, const BBox *bbox,
```
nrabinowitz, line 133: Nit: Doc comment?

dev-docs/RFCs/v4.0.0/polyfill-modes-rfc.md
isaacbrodsky: I think I updated thison the RFC branch but this PR wasn't rebased on it (or I didn't push it?)

src/h3lib/lib/algos.c
isaacbrodsky: Indeed this check is incorrect and needs to be fixed.

PR #551, opened 2021-12-14
Add error returns to maxGridDiskSize

isaacbrodsky:
This PR also changes the signature for this function to support `k` values where the number of cells is more than an `int32_t` can represent (i.e. `k=26755, maxGridDiskSize(26755)=2147570341`)

### DIFF
--- a/src/apps/benchmarks/benchmarkGridDiskCells.c
+++ b/src/apps/benchmarks/benchmarkGridDiskCells.c
@@ -23,7 +23,9 @@ H3Index pentagon = 0x89080000003ffff;
 
 BEGIN_BENCHMARKS();
 
-H3Index *out = malloc(H3_EXPORT(maxGridDiskSize)(40) * sizeof(H3Index));
+int64_t outSz;
+H3_EXPORT(maxGridDiskSize)(40, &outSz);
+H3Index *out = calloc(outSz, sizeof(H3Index));
 
 BENCHMARK(gridDisk10, 10000, { H3_EXPORT(gridDisk)(hex, 10, out); });
 BENCHMARK(gridDisk20, 10000, { H3_EXPORT(gridDisk)(hex, 20, out); });
--- a/src/apps/benchmarks/benchmarkGridDiskCells.c
+++ b/src/apps/benchmarks/benchmarkGridDiskCells.c
@@ -24,7 +24,10 @@ H3Index pentagon = 0x89080000003ffff;
 BEGIN_BENCHMARKS();
 
 int64_t outSz;
-H3_EXPORT(maxGridDiskSize)(40, &outSz);
+if (H3_EXPORT(maxGridDiskSize)(40, &outSz)) {
+    printf("Failed\n");
+    return 1;
+}
 H3Index *out = calloc(outSz, sizeof(H3Index));
 
 BENCHMARK(gridDisk10, 10000, { H3_EXPORT(gridDisk)(hex, 10, out); });
--- a/src/apps/testapps/testGridDisk.c
+++ b/src/apps/testapps/testGridDisk.c
@@ -388,10 +388,10 @@ SUITE(gridDisk) {
             "gridDiskDistances invalid k");
         t_assert(H3_EXPORT(gridDiskDistancesUnsafe)(index, -1, NULL, NULL) ==
                      E_DOMAIN,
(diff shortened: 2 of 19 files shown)


src/apps/benchmarks/benchmarkGridDiskCells.c
```diff
@@ -23,7 +23,9 @@ H3Index pentagon = 0x89080000003ffff;
 
 BEGIN_BENCHMARKS();
 
-H3Index *out = malloc(H3_EXPORT(maxGridDiskSize)(40) * sizeof(H3Index));
+int64_t outSz;
+H3_EXPORT(maxGridDiskSize)(40, &outSz);
```
dfellis, line 27: Should the benchmarks check for the error condition?

dfellis (approved):
Only a minor comment on the validity of benchmarks that all fallible functions but assume correctness.

src/apps/testapps/testGridDisk.c
```diff
@@ -370,11 +372,37 @@ SUITE(gridDisk) {
...
     }
+
+    TEST(gridDiskDistances_invalidK) {
+        H3Index index = 0x811d7ffffffffff;
+        t_assert(
+            H3_EXPORT(gridDiskDistances)(index, -1, NULL, NULL) == E_DOMAIN,
+            "gridDiskDistances invalid k");
+        t_assert(H3_EXPORT(gridDiskDistancesUnsafe)(index, -1, NULL, NULL) ==
+                     E_DOMAIN,
+                 "gridDiskDistancesUnsafe invaldi k");
```
nrabinowitz, line 391:
```suggestion
                 "gridDiskDistancesUnsafe invalid k");
```

src/apps/testapps/testGridDisk.c
```diff
@@ -370,11 +372,37 @@ SUITE(gridDisk) {
...
+        t_assert(
+            H3_EXPORT(gridDiskDistancesSafe)(index, -1, NULL, NULL) == E_DOMAIN,
+            "gridDiskDistancesSafe invaldi k");
```
nrabinowitz, line 394:
```suggestion
            "gridDiskDistancesSafe invalid k");
```

src/h3lib/lib/algos.c
```diff
@@ -155,9 +155,16 @@ static const Direction NEW_ADJUSTMENT_III[7][7] = {
...
- * @param  k   k value, k >= 0.
+ * @param   k   k value, k >= 0.
+ * @param out   size in indexes
  */
-int H3_EXPORT(maxGridDiskSize)(int k) { return 3 * k * (k + 1) + 1; }
+H3Error H3_EXPORT(maxGridDiskSize)(int k, int64_t *out) {
+    if (k < 0) {
+        return E_DOMAIN;
+    }
+    *out = 3 * (int64_t)k * ((int64_t)k + 1) + 1;
```
nrabinowitz, line 165: Would it make sense to check somewhere that this value is less than the max number of cells at the current res? E.g. should `kRing(res0cell, 100)` throw `E_DOMAIN`?

nrabinowitz (approved):

src/apps/benchmarks/benchmarkGridDiskCells.c
isaacbrodsky: Updated for the setup part. I didn't feel the benchmark should check this as part of what it times since this isn't part of the code we're intending to measure.

ajfriend (approved):

src/h3lib/lib/algos.c
isaacbrodsky: I have a concern that in some cases this will overflow, in that case we should consider returning the maximum number of cells?

ajfriend:
One minor comment: Should we standardize the type we use for distance? Here, it looks like we're using `int`, but we use `int64_t` in `gridDistance`.

https://github.com/uber/h3/blob/cd470158dc499e9bd5e6c6ec1a7e9a2373bfb1f3/src/h3lib/include/h3api.h.in#L701-L702

src/h3lib/lib/algos.c
nrabinowitz: Yeah, that's probably better than returning an error (and should be correct, in that I think k-rings covering more than the total number of cells should return all cells at that res, right?).

We should be able to check a constant to avoid integer overflow, right?

PR #553, opened 2021-12-22
Add more fuzzers and libFuzzer harness

isaacbrodsky:
This adds additional fuzzing harnesses (consult `README.md` in the PR for which functions should be covered and uncovered - a few are left uncovered since they are trivial) and also adds an adapter from AFL++ to libFuzzer. The idea being we can move the fuzzer harness from https://github.com/google/oss-fuzz/blob/master/projects/h3/h3_fuzzer.c to in repo (replaces #448) and cover all functions.

Possible improvements include testing the build using afl++ itself and hooking this up to OSS-Fuzz.

cc @AdamKorcz


.github/workflows/test-fuzzer.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Build
+        run: |
+          cd build
+          make fuzzers
+
+      - name: Run fuzzers once
+        run: |
+          cd build
+          for fuzzer in bin/fuzzers*; do
```
isaacbrodsky, line 37: Github isn't expanding this - need to find it's preferred way to do that.

isaacbrodsky:
The windows failure appears to be related to a problem fixed in #550 about those functions not having `DECLSPEC`.

src/apps/fuzzers/fuzzerPolygonToCells.c
```diff
@@ -0,0 +1,96 @@
...
+    const inputArgs *args = (const inputArgs *)data;
+    int res = args->res % (MAX_RES + 1);
+
+    GeoPolygon geoPolygon;
+    geoPolygon.numHoles = args->numHoles % MAX_HOLES;
+    if (geoPolygon.numHoles < 0) {
+        return 0;
+    }
+    geoPolygon.holes = calloc(geoPolygon.numHoles, sizeof(GeoLoop));
+    size_t offset = sizeof(inputArgs);
```
nrabinowitz, line 76: I'm quite confused here - isn't the fuzzer only going to provide `sizeof(inputArgs)` worth of data? Where does the rest of the data after `offset` come from in that case?

src/apps/fuzzers/fuzzerResolutions.c
```diff
@@ -0,0 +1,47 @@
...
+    int res;
+} inputArgs;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    if (size < sizeof(inputArgs)) {
+        return 0;
+    }
+    const inputArgs *args = (const inputArgs *)data;
+
+    // TODO: Do not perform bound checks
```
nrabinowitz, line 34: Does this mean these functions are currently unsafe?

nrabinowitz (approved):
This looks great overall - I think my confusion here is mostly my own lack of understanding of fuzzers, not problems with the code, but more comments might be helpful overall.

CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ option(BUILD_BENCHMARKS "Build benchmarking applications." ON)
 option(BUILD_FUZZERS "Build fuzzer applications (for use with afl)." ON)
 option(BUILD_FILTERS "Build filter applications." ON)
 option(BUILD_GENERATORS "Build code generation applications." ON)
+# If ON, libfuzzer settings are used to build the fuzzer harnesses. If OFF, a frontend
+# for afl++ is provided instead.
```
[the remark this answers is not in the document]
isaacbrodsky: `frontend` in this context means an implementation of `main` that accepts arguments in a way that AFL can integrate with. Should I expand on the comments for this option?

src/apps/fuzzers/fuzzerPolygonToCells.c
```diff
@@ -0,0 +1,96 @@
...
+    *offset = *offset + sizeof(LatLng) * numVerts;
+    return 0;
+}
+
+void run(GeoPolygon *geoPolygon, int res) {
+    int64_t sz;
+    H3Error err = H3_EXPORT(maxPolygonToCellsSize)(geoPolygon, res, &sz);
+    if (!err && sz < MAX_SZ) {
+        if (sz < 0) {
+            printf("Oh no - sz is negative\n");
```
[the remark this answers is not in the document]
isaacbrodsky: As it is it will crash on line 57. Not sure if this will be relevant after #551.

src/apps/fuzzers/fuzzerPolygonToCells.c
isaacbrodsky: I will need to check if this will work with AFL (might need to adjust the seed file so it has data after inputArgs). For libFuzzer it does not know about `inputArgs` so it will generate larger test cases.

src/apps/fuzzers/fuzzerResolutions.c
isaacbrodsky: Yes, this is addressed in #550. Once that is merged these can be fuzzed. As it is they crash almost instantly (consider resolution = 17)

src/apps/fuzzers/fuzzerH3SetToLinkedGeo.c
```diff
@@ -0,0 +1,46 @@
...
+    int sz;
+} inputArgs;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    if (size < sizeof(inputArgs)) {
+        return 0;
+    }
+    const inputArgs *args = (const inputArgs *)data;
+    if (args->sz >= 1024) {
+        return 0;
```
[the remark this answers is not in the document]
isaacbrodsky: This is a serious issue with memory management in C. I considered this case to be out of scope for the fuzzers since I don't know of any way to make H3 resilient to this kind of error.

We should be careful to document the expected array sizes when users pass memory into H3 so this can be correctly implemented on the user side.

src/apps/fuzzers/fuzzerIndexIO.c
```diff
@@ -0,0 +1,47 @@
...
+    H3Index index;
+    char str[STRING_LENGTH];
+} inputArgs;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    if (size < sizeof(inputArgs)) {
+        return 0;
+    }
+    inputArgs args;
+    memcpy(&args, data, sizeof(inputArgs));
```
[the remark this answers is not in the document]
nrabinowitz: Worth a comment in that case I think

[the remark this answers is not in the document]
isaacbrodsky: Will add. Come to think of it we may wish to address this in the API of `stringToH3` by passing the buffer length.

src/apps/fuzzers/fuzzerCellArea.c
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2021 Uber Technologies, Inc.
```
ajfriend, line 2: 2022?

src/apps/fuzzers/fuzzerCompact.c
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+        if (compactedCount < 2) {
+            int64_t uncompactedSize;
+            H3Error err = H3_EXPORT(uncompactCellsSize)(
+                compacted, inputSize, uncompactRes, &uncompactedSize);
+            if (!err && uncompactedSize < MAX_UNCOMPACT_SIZE) {
+                H3Index *uncompacted = calloc(uncompactedSize, sizeof(H3Index));
+                H3_EXPORT(uncompactCells)
+                (compacted, compactedCount, uncompacted, uncompactedSize,
```
ajfriend, line 56: I don't think you can use `compactedCount` here, as nothing guarantees the non `H3_NULL` cells that you counted are at the front of the array.

isaacbrodsky: In that case I would suggest just removing this entire block from lines 42-61 since we test uncompact later on the original fuzzer input anyways.

src/apps/fuzzers/fuzzerCellArea.c
isaacbrodsky: I think I started the file in 2021 thus the header. Can be updated.

src/apps/fuzzers/fuzzerCompact.c
ajfriend: At least, I don't think so. But maybe we should have @dfellis confirm.

@isaacbrodsky, what was the reasoning for restricting to `< 2` compacted cells?

isaacbrodsky: As I recall I copied this from oss-fuzz, I think the idea there was the input was expected to behave in a certain way.

src/apps/fuzzers/fuzzerCompact.c
```diff
@@ -0,0 +1,79 @@
...
+const int MAX_UNCOMPACT_RES = 9;
+const int64_t MAX_UNCOMPACT_SIZE = 4000000;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    if (size < sizeof(H3Index) * 2) {
+        return 0;
+    }
+
+    uint8_t res = *data;
+    uint8_t uncompactRes = (*(data + 1)) % (MAX_UNCOMPACT_RES + 1);
```
ajfriend, line 33: It would be nice if we could try uncompacting to res 15 (or beyond, and catch the error), as I'd expect that would be the place where out of bounds issues would come up. Maybe we put a bound on the uncompact res delta instead of res = 9?

src/apps/fuzzers/fuzzerCompact.c
```diff
@@ -0,0 +1,79 @@
...
+    H3Error err =
+        H3_EXPORT(uncompactCellsSize)(input, inputSize, res, &uncompactedSize);
+
+    if (!err && uncompactedSize < MAX_UNCOMPACT_SIZE) {
+        H3Index *uncompacted = calloc(uncompactedSize, sizeof(H3Index));
+        H3_EXPORT(uncompactCells)
+        (input, inputSize, uncompacted, uncompactedSize, res);
+        free(uncompacted);
+    }
+    free(compacted);
```
ajfriend, line 74: If we do end up uncompacting more than once, maybe an helper function like:

```c
void tryUncompact(H3Index *cells, int64_t numCells, int res){
    int64_t uncompactedSize;
    H3Error err =
        H3_EXPORT(uncompactCellsSize)(cells, numCells, res, &uncompactedSize);

    if (!err && uncompactedSize < MAX_UNCOMPACT_SIZE) {
        H3Index *uncompacted = calloc(uncompactedSize, sizeof(H3Index));
        H3_EXPORT(uncompactCells)
        (cells, numCells, uncompacted, uncompactedSize, res);
        free(uncompacted);
    }
    free(compacted);
}
```

That would help to make sure the `numCells` matches between the `uncompactCells` and `uncompactCellsSize` calls.

src/apps/fuzzers/fuzzerCompact.c
isaacbrodsky: I'd suggest deferring this to a follow up PR. There are some other enhancements and fixes to these functions and this fuzzer but I would prefer to make them a new, more digestible PR.

src/apps/fuzzers/fuzzerCompact.c
isaacbrodsky: Agreed if we go that approach, although right now I am planning to reduce this to a single call to uncompactCells in a follow up PR. Are you OK with that approach or would prefer it being rolled into this PR?

ajfriend: Next PR works!

src/apps/fuzzers/fuzzerResolutions.c
```diff
@@ -0,0 +1,49 @@
...
+    H3_EXPORT(getHexagonAreaAvgKm2)(args->res, &out);
+    H3_EXPORT(getHexagonAreaAvgM2)(args->res, &out);
+    H3_EXPORT(getHexagonEdgeLengthAvgKm)(args->res, &out);
+    H3_EXPORT(getHexagonEdgeLengthAvgM)(args->res, &out);
+
+    int64_t outInt;
+    H3_EXPORT(getNumCells)(args->res, &outInt);
+
+    H3Index pentagons[12];
+    H3_EXPORT(getPentagons)(args->res, pentagons);
```
ajfriend, line 44: This is probably just a reflection of me not knowing how fuzzers work, but when you're doing multiple sequential tests like this, does it become harder to test the last function because you need to "get past" all the previous ones? Is there any value is splitting out each function separately? But that's also a lot of annoying boilerplate...

isaacbrodsky: It's correct that if any of the previous functions called were to crash, this function wouldn't be exercised. But in that case we'd still have found a crash warranting investigation anyways. If we explicitly terminate the fuzzing run before reaching this function (i.e. if getNumCells were to return an error then we return 0), then that would not be a good test because this function's invocation would be dependent on the previous function. That doesn't happen here so I believe this is a valid exercise of all the functions in this file.

src/apps/fuzzers/fuzzerLocalIj.c
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    if (size < sizeof(inputArgs)) {
+        return 0;
+    }
+    const inputArgs *args = (const inputArgs *)data;
+
+    // Note that index and index2 need to be in the approximate area for these
```
ajfriend, line 50: "same approximate area"?

ajfriend (approved):

PR #554, opened 2021-12-28
[website] Fix LiveEditor so code is displayed in the documentation (e.g. quickstart)

kinow:
Hi,

I am experimenting with H3 for a project at $work, and noticed that all the live code editors were showing the same error message in the H3 website (screenshot from <https://h3geo.org/docs/quickstart#point--cell>):

image

I think it was due this this PR in docusaurus: https://github.com/facebook/docusaurus/pull/5349

The `isClient` was removed from the context, so in the theme configuration, `isClient` is always `undefined`. That causes the `code` to be always an empty string too.

This PR uses the `useIsBrowser` used in the linked PR in docusaurus. It works for me locally after this change :+1: 

image

Thanks for H3 :rocket: 
-Bruno


isaacbrodsky (approved):
Thanks for noticing & fixing this issue - confirmed the fix works locally for me

PR #556, opened 2022-01-03
Create two unit tests for the 'h3' cli app

dfellis:
Not adding any extra functionality to the app until we have tests for it. Wanted to put this approach in front of you guys to see what you think.

I can't figure out how to do coverage tests on the CLI app, but I don't think that's too big of a problem?


CMakeLists.txt
```diff
@@ -533,6 +533,18 @@ if(H3_IS_ROOT_PROJECT AND BUILD_TESTING)
...
     endmacro()
 
+    macro(add_h3_cli_test name h3_args expect_string)
+        add_test(NAME ${name}_test${test_number}
+                 # /usr/bin/sh -c "test \"`./bin/h3 cellToLatLng -c 8928342e20fffff`\" = \"37.5012466151, -122.5003039349\""
+                 COMMAND ${SHELL} "test \"`./bin/h3 ${h3_args}`\" = '${expect_string}'")
+
+        if(PRINT_TEST_FILES)
+            message("${name}_test${test_number} - ${argfile}")
+        endif()
```
isaacbrodsky, line 543: `argfile` doesn't look like it'd be defined here, so may as well drop this (or update it)

CMakeLists.txt
```diff
@@ -533,6 +533,18 @@ if(H3_IS_ROOT_PROJECT AND BUILD_TESTING)
         endif()
     endmacro()
 
+    macro(add_h3_cli_test name h3_args expect_string)
+        add_test(NAME ${name}_test${test_number}
+                 # /usr/bin/sh -c "test \"`./bin/h3 cellToLatLng -c 8928342e20fffff`\" = \"37.5012466151, -122.5003039349\""
+                 COMMAND ${SHELL} "test \"`./bin/h3 ${h3_args}`\" = '${expect_string}'")
```
isaacbrodsky, line 539: We may need to check for `test` being available and error the CMake configuration if not.

CMakeLists.txt
```diff
@@ -533,6 +533,18 @@ if(H3_IS_ROOT_PROJECT AND BUILD_TESTING)
...
+
+        if(PRINT_TEST_FILES)
+            message("${name}_test${test_number} - ${argfile}")
+        endif()
+
+        # No coverage possible for H3 CLI testing
```
isaacbrodsky, line 545: I assume coverage should be somehow possible (would need to configure the binaries to produce coverage, and include them in the considered source paths)?

nrabinowitz:
Not a huge fan of this approach - seems like it will bloat the CMakeLists.txt file, and I don't like the idea that we're expanding the purpose of that file beyond build tasks.

Other options:
- Put all tests in a shell script
- Put all input/output in some text file, and make a test via CMake that reads the file, runs commands from input, and asserts output
- Use a unit testing framework in other language, e.g. Node or Python

isaacbrodsky:
@nrabinowitz Another option would be to define the tests in a separate CMake file but still in that framework.

dfellis:
> @nrabinowitz Another option would be to define the tests in a separate CMake file but still in that framework.

I personally like this compromise. The CMakeLists file does already handle a lot of stuff, and this would help us organize not just the test logic but other things like the fuzzers.

nrabinowitz:
> > @nrabinowitz Another option would be to define the tests in a separate CMake file but still in that framework.
> 
> I personally like this compromise. The CMakeLists file does already handle a lot of stuff, and this would help us organize not just the test logic but other things like the fuzzers.

That seems okay by me. The main issue for me is just that the rest of the CMake file doesn't have actual data or need the kind of review that test cases do, whereas this does. Separating it, ideally with a header comment, should be fine (and I agree that other parts could be separated out as well).

CMakeLists.txt
dfellis: Discussed offline. `test` is defined because we depend on git for windows, which brings in bash, but `./bin/h3` doesn't work because it's `./bin/h3.exe` there.

CMakeLists.txt
dfellis: I'll study it, but okay if that's a separate PR?

CMakeLists.txt
```diff
@@ -533,6 +533,24 @@ if(H3_IS_ROOT_PROJECT AND BUILD_TESTING)
         endif()
     endmacro()
 
+    macro(add_h3_cli_test name h3_args expect_string)
+        # Determine the h3 binary name
+        if(WIN32)
+          set(H3CLI ./bin/$<IF:$<CONFIG:Debug>,Debug,Release>/h3.exe)
+        else()
+            set(H3CLI ./bin/h3)
+        endif()
```
isaacbrodsky, line 542: Can `TARGET_FILE` be used? https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

dfellis: I will test. I am not super confident, though. (I wish we were still on bare Make, tbh, CMake has too much hidden behavior and irregular grammar)

dfellis: It worked perfectly. Replacing the entirety of the conditional with it, now.

CMakeTests.txt
```diff
@@ -0,0 +1,228 @@
...
+            add_dependencies(${name}_coverage${test_number} clean-coverage)
+        endif()
+    endmacro()
+
+    macro(add_h3_cli_test name h3_args expect_string)
+        add_test(NAME ${name}_test${test_number}
+          COMMAND ${SHELL} "test \"`$<TARGET_FILE:h3_bin> ${h3_args}`\" = '${expect_string}'")
+
+        if(PRINT_TEST_FILES)
+            message("${name}_test${test_number} - ${expect_string}")
```
isaacbrodsky, line 124: I'd suggest adding the `h3_args` here, as the purpose of PRINT_TEST_FILES is to connect which test number (which can be opaque) corresponds with which test in this file.

isaacbrodsky (approved):

CMakeTests.txt
dfellis:
```suggestion
            message("${name}_test${test_number} - ${h3_args} - ${expect_string}")
```

CMakeTests.txt
```diff
@@ -0,0 +1,228 @@
+# Copyright 2017-2022 Uber Technologies, Inc.
```
isaacbrodsky, line 1: You might want to use the `cmake` extension for this file instead of `txt`. I think that will trigger Github's code highlighting to treat this as CMake.

dfellis: I avoided it, though, because the docs implied that it will switch to a CMake module mode instead of simple C-style include insertion of the contents of the file. I will give it a try, but this may not be possible.

dfellis: Looks like it did work.

CMakeTests.cmake
```diff
@@ -0,0 +1,228 @@
...
+    # test-fast exists so that Travis CI can run Valgrind on tests without taking a very long time.
+    add_h3_test(testDirectedEdgeExhaustive src/apps/testapps/testDirectedEdgeExhaustive.c)
+    add_h3_test(testVertexExhaustive src/apps/testapps/testVertexExhaustive.c)
+    add_h3_test(testH3ToLocalIjExhaustive src/apps/testapps/testH3ToLocalIjExhaustive.c)
+    add_h3_test(testGridPathCellsExhaustive src/apps/testapps/testGridPathCellsExhaustive.c)
+    add_h3_test(testGridDistanceExhaustive src/apps/testapps/testGridDistanceExhaustive.c)
+    add_h3_test(testH3CellAreaExhaustive src/apps/testapps/testH3CellAreaExhaustive.c)
+
+    add_h3_cli_test(testCliCellToLatLng "cellToLatLng -c 8928342e20fffff" "37.5012466151, -122.5003039349")
+    add_h3_cli_test(testCliLatLngToCell "latLngToCell --lat 20 --lng 123 -r 2" "824b9ffffffffff")
```
nrabinowitz, line 219: I still feel like there's a substantive difference between tasks that *run* a test and tasks that *are* tests. This is fine for now, but I don't see how it scales up to a large number of tests - I'd much rather have inputs and outputs defined outside of CMake (in a text file, in yaml, or even as separate lines in a script file).

nrabinowitz (approved):

PR #557, opened 2022-01-11
Add polygonToCellsNoHoles fuzzer

isaacbrodsky:
Based on #553. This adds a second fuzzer for polygonToCells which does not require the fuzzer to guess the correct number of cells.


dfellis (approved):

src/apps/fuzzers/fuzzerPolygonToCellsNoHoles.c
```diff
@@ -0,0 +1,61 @@
...
+
+    uint8_t res = *data;
+    size_t vertsSize = size - 1;
+    int numVerts = vertsSize / sizeof(LatLng);
+
+    GeoPolygon geoPolygon;
+    geoPolygon.numHoles = 0;
+    geoPolygon.holes = NULL;
+    geoPolygon.geoloop.numVerts = numVerts;
+    geoPolygon.geoloop.verts = (LatLng *)(data + 1);
```
nrabinowitz, line 54: For my info - why `+ 1`? Oh, because the first `uint_8` is the res? Might be worth a comment.

I imagine that most fuzzer runs would error quickly on an invalid `res` here, and never exercise the polyfill...

nrabinowitz (approved):

src/apps/fuzzers/fuzzerPolygonToCellsNoHoles.c
isaacbrodsky: This is still better than fuzzerPoylgonToCells.c which is very difficult for the fuzzer to make progress on; here it only needs to guess one byte correctly (polygonToCells actually takes an int but I don't think that's quite as relevant) which I believe is very doable.

PR #559, opened 2022-01-13
Bug fixes for directed edge

isaacbrodsky:
Based upon #553. This updates the tests with some cases found via fuzzing to the test suite.


src/apps/testapps/testGridDisk.c
```diff
@@ -329,6 +329,19 @@ SUITE(gridDisk) {
         t_assert(out == origin, "Moving to self goes to self");
     }
 
+    TEST(h3NeighborRotations_invalid) {
+        // This is undefined behavior, but it's helpful for it to make sense.
+        H3Index origin = 0x811d7ffffffffffL;
```
dfellis, line 334: Anything special about this particular H3 index?

dfellis (approved):

src/apps/testapps/testGridDisk.c
isaacbrodsky: No - I just needed a known good index as I wanted the only rotations to be under test.

nrabinowitz: For my info, is there any difference between indexes with the `L` suffix or without? Should we be consistent?

src/h3lib/lib/algos.c
```diff
@@ -335,6 +335,9 @@ H3Error h3NeighborRotations(H3Index origin, Direction dir, int *rotations,
                             H3Index *out) {
     H3Index current = origin;
 
+    if (dir < CENTER_DIGIT || dir >= INVALID_DIGIT) {
```
nrabinowitz, line 338: For my info, what does `CENTER_DIGIT` even do? Returns the current index?

nrabinowitz (approved):

src/apps/testapps/testGridDisk.c
isaacbrodsky: All valid H3 indexes would only be valid integers as 64 bit, so no I don't believe so.

src/h3lib/lib/algos.c
isaacbrodsky: Yes, that is my expectation

PR #570, opened 2022-01-27
polygonToCells API accepts a flags argument

isaacbrodsky:
This supersedes and includes #519 and #549, but includes only the API changes, but not any specific implementations.

### DIFF
--- a/src/apps/testapps/testH3Memory.c
+++ b/src/apps/testapps/testH3Memory.c
@@ -189,31 +189,31 @@ SUITE(h3Memory) {
         sfGeoPolygon.numHoles = 0;
 
         int64_t numHexagons;
-        t_assertSuccess(
-            H3_EXPORT(maxPolygonToCellsSize)(&sfGeoPolygon, 9, &numHexagons));
+        t_assertSuccess(H3_EXPORT(maxPolygonToCellsSize)(&sfGeoPolygon, 9, 0,
+                                                         &numHexagons));
         H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
 
         resetMemoryCounters(0);
         failAlloc = true;
-        H3Error err = H3_EXPORT(polygonToCells)(&sfGeoPolygon, 9, hexagons);
+        H3Error err = H3_EXPORT(polygonToCells)(&sfGeoPolygon, 9, 0, hexagons);
         t_assert(err == E_MEMORY, "polygonToCells failed (1)");
         t_assert(actualAllocCalls == 1, "alloc called once");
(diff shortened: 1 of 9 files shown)


dfellis (approved):

src/apps/testapps/testH3Memory.c
```diff
@@ -189,31 +189,31 @@ SUITE(h3Memory) {
...
-        t_assertSuccess(
-            H3_EXPORT(maxPolygonToCellsSize)(&sfGeoPolygon, 9, &numHexagons));
+        t_assertSuccess(H3_EXPORT(maxPolygonToCellsSize)(&sfGeoPolygon, 9, 0,
+                                                         &numHexagons));
         H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
 
         resetMemoryCounters(0);
         failAlloc = true;
-        H3Error err = H3_EXPORT(polygonToCells)(&sfGeoPolygon, 9, hexagons);
+        H3Error err = H3_EXPORT(polygonToCells)(&sfGeoPolygon, 9, 0, hexagons);
```
nrabinowitz, line 198: Nit: Would prefer that tests use a const like `DEFAULT_FLAGS` instead of `0` - it makes the call clearer and it will be easier to update when we have meaningful input to test.

src/h3lib/lib/algos.c
```diff
@@ -749,7 +749,10 @@ H3Error H3_EXPORT(gridRingUnsafe)(H3Index origin, int k, H3Index *out) {
  * @return 0 (E_SUCCESS) on success.
  */
 H3Error H3_EXPORT(maxPolygonToCellsSize)(const GeoPolygon *geoPolygon, int res,
-                                         int64_t *out) {
+                                         uint32_t flags, int64_t *out) {
+    if (flags != 0) {
+        return E_FAILED;
```
nrabinowitz, line 754: Worth adding `E_UNSUPPORTED_FLAGS` or similar?

nrabinowitz (approved):

src/apps/testapps/testH3Memory.c
isaacbrodsky: The value 0 will be part of the API so we will not be modifying the value of `DEFAULT_FLAGS`. I can certainly add a constant for that. In the future when polygonToCells accepts more flags, we'll need to offer appropriate constants / defines for users to OR together.

src/h3lib/lib/algos.c
isaacbrodsky: I'm fine either way. I don't believe the specific error code is considered part of the API contract so I don't see an issue with it changing.

PR #571, opened 2022-01-28
Rename h3SetToLinkedGeo to cellsToLinkedMultiPolygon

isaacbrodsky:
Rename for consistency with the naming RFC. The name "LinkedMultiPolygon" was chosen so that in the future a "ToMultiPolygon" version could be added; multi was needed since this is inherently a multipolygon as differentiated from polygonToCells which deals only with polygons (non-multi)


dfellis (approved):
You are *FAST*

isaacbrodsky:
Partially automated refactoring :)

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
```diff
@@ -271,10 +271,12 @@ discussion requiring benchmarking, so we will defer that to a
...
-| *DNE*                             | `boundaryToLoop`         |                        |
-| `getH3UnidirectionalEdgeBoundary` | `directedEdgeToBoundary` | returns `CellBoundary` |
+|              3.x name             |       Proposed name         |         Notes              |
+|-----------------------------------|-----------------------------|----------------------------|
+| `h3ToGeoBoundary`                 | `cellToBoundary`            | returns `CellBoundary`     |
+| *DNE*                             | `cellToLoop`                | returns `GeoLoop`          |
+| *DNE*                             | `loopToBoundary`            |                            |
+| *DNE*                             | `boundaryToLoop`            |                            |
+| `getH3UnidirectionalEdgeBoundary` | `directedEdgeToBoundary`    | returns `CellBoundary`     |
+| `h3SetToLinkedGeo`                | `cellsToLinkedMultiPolygon` | returns `LinkedGeoPolygon` |
```
nrabinowitz, line 281: Should we also rename `LinkedGeoPolygon`? This struct is part of the external API, right?

nrabinowitz (approved):

dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md
isaacbrodsky: Per line 269 it looks like the rename should be `LinkedGeoMultiPolygon`? It's part of the external C API in `h3api.h`, although it is not needed for binary compatability

nrabinowitz: That shouldn't be necessary. The linked structures have a next pointer. So instead of a MultiPolygon with Polygon children, you have a Polygon that can optionally define a pointer to a Polygon sibling. 

Maybe the rename isn't necessary - I thought we were dropping the `Geo` in `GeoPolygon`, so I was suggesting this should follow suit. But if not, this is fin - though I guess we could update `LinkedGeoCoord` to `LinkedLatLng` for consistency.

Issue #572, opened 2022-01-31
h3togeo is not the centroid of the cell


francois-baptiste:
In the API documentation (https://h3geo.org/docs/api/indexing/#h3togeo) `h3togeo` is described as returning the index centroid.
However due to distortion of some cells, this is not entirely exact if we consider the standard definition of the centroid (as the projected center of the uniformly distributed mass of the cell index).

`h3togeo` should be better defined as the convergence point of the central child cell when the resolution goes finer.

For instance if I calculate the centroid of the index `85283473fffffff` (resolution 5), I get the point `POINT(-121.976343311047 37.345768562411)` while the `h3togeo` function returns the point `POINT(-121.976375972551 37.34576935375)`

The first point can be mapped to index "8f2834700000036" at resolution 15, while the second is mapped to "8f2834700000000" which is the central child of the initial cell.

sahrk:
Yes, that document is incorrect. The returned point is actually the center of the hexagon in the H3 projection space, inversely projected to the sphere.

isaacbrodsky:
Thanks for pointing this out and to @sahrk for confirming the correct explanation. I took a shot at updating the docs in #574 to fix this.

dfellis:
I am somewhat confused by this discussion.

On each icosahedron face the hexagons are uniform in size and the centroid of the hexagon *is* the exact center of the hexagon. This icosahedron is then projected back onto the Earth with the gnomonic projection system, so the centroid being shifted would be due to that projection and will vary based on where on the icosahedron it is: the central hexagon of each icosahedron *should* have a centroid that matches reality, while the hexagons closest to the vertices of the icosahedron faces should have the greatest discrepancy (and greatest distortion from a Regular hexagon).

So perhaps it would be best to state the centerpoint as reported by `h3ToGeo`/`cellToLatLng` is the center of the hexagon in grid space, and will drift versus the centroid of the hexagon on Earth due to distortion from the gnomonic projection within the icosahedron face it resides on and its distance from the center of the icosahedron face?

sahrkm:
[...]

PR #574, opened 2022-02-02
Fix cellToLatLng docs per #572

isaacbrodsky:
See #572; this is not the centroid of the cell geometry on the sphere but rather the centroid of the cell in the projected grid system, unprojected onto the sphere. These can differ so we should be exact.

### DIFF
--- a/website/docs/api/indexing.mdx
+++ b/website/docs/api/indexing.mdx
@@ -63,7 +63,7 @@ function example() {
 </Tabs>
 
 Indexes the location at the specified resolution, returning the index of the cell containing the location. This buckets
-the geographic point into the H3 grid.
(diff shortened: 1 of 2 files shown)


dfellis (approved):

nrabinowitz (approved):

sahrk (commented):
Might I suggest:

` Returns the geographical coordinates corresponding to the center point of the cell in the planar H3 projection space.`

This point is not necessarily centered on the center child, though I agree with @francois-baptiste that it is the convergence point of the center descendants as resolution goes to infinity. I’d also be wary of saying that it is “approximately the centroid”, unless that’s been quantified. Both these points feel to me like too much detail for an API function description. Maybe add more discussion to the [description in the algorithms section](https://h3geo.org/docs/core-library/h3ToGeoDesc/)?

ajfriend (approved):

PR #586, opened 2022-02-28
Rename experimentalLocalIjToH3 to localIjToCell

isaacbrodsky:
Renames `experimental...` functions to no longer be experimental, as they have been in the library for some time now. Also renames the functions to be in line with the naming RFC.


dfellis:
This does not include the mode that was discussed? Part of dropping the experimental part of the name was that with a mode we could add improved versions without breaking existing users.

isaacbrodsky:
> This does not include the mode that was discussed? Part of dropping the experimental part of the name was that with a mode we could add improved versions without breaking existing users.

Correct - I can certainly add that in this PR if you think it makes sense with the renaming.

dfellis:
> Correct - I can certainly add that in this PR if you think it makes sense with the renaming.

Yeah, I think they go together.

isaacbrodsky:
> Yeah, I think they go together.

Updated

src/h3lib/lib/localij.c
```diff
@@ -519,10 +519,15 @@ H3Error localIjkToCell(H3Index origin, const CoordIJK *ijk, H3Index *out) {
  *
  * @param origin An anchoring index for the ij coordinate system.
  * @param index Index to find the coordinates of
+ * @param flags Mode flags, must be 0
```
dfellis, line 522: Do we want this to be a set of bitflags or an integer mode? I thought we went for the latter because we couldn't think of any combinatorial flags that would make any sense?

isaacbrodsky: I'm treating it as an opaque mode identifier without structure until we decide further.

dfellis: So, I meant the docs all say "flags"? And I thought we had decided to just make it a mode enum/integer?

isaacbrodsky: I see what you mean. In my mind these were interchangeable (I could just say there's a single flags bit field of length 32 bits). I standardized on `mode`.

dfellis (approved):

src/h3lib/lib/localij.c
```diff
@@ -548,24 +548,24 @@ H3Error H3_EXPORT(experimentalH3ToLocalIj)(H3Index origin, H3Index h3,
...
-H3Error H3_EXPORT(experimentalLocalIjToH3)(H3Index origin, const CoordIJ *ij,
-                                           H3Index *out) {
-    // This function is currently experimental. Once ready to be part of the
-    // non-experimental API, this function (with the experimental prefix) will
-    // be marked as deprecated and to be removed in the next major version. It
-    // will be replaced with a non-prefixed function name.
+H3Error H3_EXPORT(localIjToCell)(H3Index origin, const CoordIJ *ij,
+                                 uint32_t mode, H3Index *out) {
+    if (mode != 0) {
+        return E_FAILED;
```
nrabinowitz, line 563: Suggestion: make a special `E_INVALID_MODE`? I imagine we'll need this for other cases too, and for `polygonToCells`

nrabinowitz (approved):

PR #587, opened 2022-03-07
Separate long function _geoToHex2d

wlbksy:
`_geoToHex2d` has two steps
1. determine which face this point is on
2. calculate the coordinates of this point on this face.

Here, it is better to separate this function into two functions.

### DIFF
--- a/src/h3lib/lib/faceijk.c
+++ b/src/h3lib/lib/faceijk.c
@@ -387,19 +387,9 @@ void _geoToFaceIjk(const LatLng *g, int res, FaceIJK *h) {
  * @param v The 2D hex coordinates of the cell containing the point.
  */
 void _geoToHex2d(const LatLng *g, int res, int *face, Vec2d *v) {
-    Vec3d v3d;
-    _geoToVec3d(g, &v3d);
-
     // determine the icosahedron face
-    *face = 0;
-    double sqd = _pointSquareDist(&faceCenterPoint[0], &v3d);
-    for (int f = 1; f < NUM_ICOSA_FACES; f++) {
-        double sqdT = _pointSquareDist(&faceCenterPoint[f], &v3d);
-        if (sqdT < sqd) {
-            *face = f;
-            sqd = sqdT;
-        }
-    }
+    double sqd;
+    _geoToClosestFace(g, face, sqd);
 
     // cos(r) = 1 - 2 * sin^2(r/2) = 1 - 2 * (sqd / 4) = 1 - sqd/2
     double r = acos(1 - sqd / 2);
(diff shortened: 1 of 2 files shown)


src/h3lib/lib/faceijk.c
```diff
@@ -387,19 +387,9 @@ void _geoToFaceIjk(const LatLng *g, int res, FaceIJK *h) {
...
-    double sqd = _pointSquareDist(&faceCenterPoint[0], &v3d);
-    for (int f = 1; f < NUM_ICOSA_FACES; f++) {
-        double sqdT = _pointSquareDist(&faceCenterPoint[f], &v3d);
-        if (sqdT < sqd) {
-            *face = f;
-            sqd = sqdT;
-        }
-    }
+    double sqd;
+    _geoToClosestFace(g, face, sqd);
```
nrabinowitz, line 392: I think this needs to be `&sqd`, right?

wlbksy: Good catch. I'll fix it soon.

wlbksy:
> I don't have any particular objection to this change, but can you explain the use case or problem you're trying to solve? 

Thanks for your comments, Nick.

I've using a modified version of h3 for quite a long time. I have area of interest which are near the h3 pentagons, this makes the hexagon grids distorted.
The idea of this modified version is quite simple: Rotate the h3 mesh grids, so that the nearest face center is near my area of interest. That's why I need to identify the nearest icosahedron face, so I could rotate this center and the whole h3 mesh grids

dfellis:
> > I don't have any particular objection to this change, but can you explain the use case or problem you're trying to solve?
> 
> Thanks for your comments, Nick.
> 
> I've been using a modified version of h3 for quite a long time. I have area of interest which are near the h3 pentagons, this makes the hexagon grids distorted. The idea of this modified version is quite simple: Rotate the h3 mesh grids, so that the nearest face center is near my area of interest. That's why I need to identify the nearest icosahedron face, so I could rotate this center and the whole h3 mesh grids

Is a modified H3 necessary, or could you just perform a `lat, lng` transform to rotate all of the coordinates to be within that center? (Basically, it could be a lot easier to do this with a preprocessing and postprocessing step than maintaining a fork?

wlbksy:
> Basically, it could be a lot easier to do this with a preprocessing and postprocessing step than maintaining a fork?

It is a preprocessing and postprocessing step. I'm thinking of move these api to h3 gradually.

It precomputes  the rotation matrix from new coordinates to h3, and the reverse one too. 
Coordinates to grid encodings: use this rotation first, then encoding the new coordinates to h3 grids.
Grid encodings to coordinates: decoding first, then use rotation

src/h3lib/lib/faceijk.c
```diff
@@ -934,3 +924,29 @@ Overage _adjustPentVertOverage(FaceIJK *fijk, int res) {
...
+void _geoToClosestFace(const LatLng *g, int *face, double *sqd) {
+    Vec3d v3d;
+    _geoToVec3d(g, &v3d);
+
+    // determine the icosahedron face
+    *face = 0;
+    // The distance between two farthest points is 2.0, therefore the square of
+    // the distance between two points should always be less or equal than 4.0 .
+    *sqd = 5.0;
+    for (int f = 0; f < NUM_ICOSA_FACES; ++f) {
```
isaacbrodsky, line 945: why not set sqd to `_pointSquareDist(&faceCenterPoint[0], &v3d)` and start `f` from 1, as before?

isaacbrodsky (approved):

src/h3lib/lib/faceijk.c
nrabinowitz: This is a bit of a wash IMO - on the one hand you have a duplicated function call, on the other you have specialized knowledge/assumptions about the output of `_pointSquareDist`. I think either is fine.

nrabinowitz (approved):

PR #599, opened 2022-04-14
polygonToCells: remove ring tracing step

kalenikaliaksandr:
polygonToCells algorithm basically seems to have following steps:
1. find hexagons lying on polygons outer and inner rings
2. do breadth-first search starting from every found hexagon and check if next neighbour hexagon center contained by polygon

First step with ring hexagons tracing seems unnecessary for me because in breadth-first search any point from polygon can be used as start point and all others will be found eventually.

### DIFF
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -776,72 +776,6 @@ H3Error H3_EXPORT(maxPolygonToCellsSize)(const GeoPolygon *geoPolygon, int res,
 }
 
 /**
- * _getEdgeHexagons takes a given geoloop ring (either the main geoloop or
- * one of the holes) and traces it with hexagons and updates the search and
- * found memory blocks. This is used for determining the initial hexagon set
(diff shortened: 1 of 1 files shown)


nrabinowitz:
Thank you for your submission. In the future, it might make sense to discuss in our Slack channel or in a GitHub issue before putting together a PR.

The reason we include the step of tracing the outlines is that the cells returned by `polyfill` may not be contiguous. The current `polyfill` algorithm only takes cells whose centers fall within the polygon, so there are many cases in which a simple flood fill won't work - imagine for example a "barbell" shape with two areas filled with cells, separated by a arbitrarily long "bar" with no cells. Depending on the scale of the shapes and the resolution of the cells, there may be many cases like this in a given dataset. Outline tracing was the best option we could come up with that handles all of these cases. We'd welcome other more efficient approaches here.

It looks like we don't have good test cases for these issues, however - I would have hoped the tests would fail in this PR, but they don't, which suggests a gap in our unit tests.

PR #614, opened 2022-06-09
Updated for BigQuery and Snowflake

jatorre:
Added the Open source Analytics toolbox for BigQuery and Snowflake with the Open Source CARTO Analytics Toolbox.

### DIFF
--- a/website/docs/community/bindings.md
+++ b/website/docs/community/bindings.md
@@ -17,7 +17,9 @@ As a C library, bindings can be made to call H3 functions from different program
 
 ## BigQuery
 
-- [cartodb/bigquery-jslibs](https://github.com/CartoDB/bigquery-jslibs)
+- [CARTO Analytics Toolbox](https://docs.carto.com/analytics-toolbox-bigquery/sql-reference/h3/)
+- [CARTO Analytics Toobox source code](https://github.com/CartoDB/analytics-toolbox-core)
+- [cartodb/bigquery-jslibs (Deprecated)](https://github.com/CartoDB/bigquery-jslibs) 
 
 ## C&#35;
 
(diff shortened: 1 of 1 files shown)


website/docs/community/bindings.md
```diff
@@ -17,7 +17,9 @@ As a C library, bindings can be made to call H3 functions from different program
 
 ## BigQuery
 
-- [cartodb/bigquery-jslibs](https://github.com/CartoDB/bigquery-jslibs)
+- [CARTO Analytics Toolbox](https://docs.carto.com/analytics-toolbox-bigquery/sql-reference/h3/)
```
nrabinowitz, line 20: I'd prefer to leave the docs out - many of the other linked projects have docs, but we link to the main page instead. I can see that you're trying to differentiate the DB-specific pieces, but this seems pretty clear from the main page - if we wanted this to be DB-specific, maybe link to https://github.com/CartoDB/analytics-toolbox-core#bigquery ?

website/docs/community/bindings.md
```diff
@@ -17,7 +17,9 @@ As a C library, bindings can be made to call H3 functions from different program
...
+- [CARTO Analytics Toobox source code](https://github.com/CartoDB/analytics-toolbox-core)
+- [cartodb/bigquery-jslibs (Deprecated)](https://github.com/CartoDB/bigquery-jslibs)
```
nrabinowitz, line 22: Is there a reason to include this if it's deprecated?

website/docs/community/bindings.md
jatorre: Well, the issue with Bigquery and snowflake is that you can have access to the functions without the need to install them, and that is probably what users want more. 

I have created a new section on the github repo on Getting Started so that the user knows they can start using them without doing anything, or they can install them directly on the DB.

https://github.com/CartoDB/analytics-toolbox-core/pull/250

My proposal is to link to their section, what do you think?

website/docs/community/bindings.md
jatorre: No, no reason, good point.

website/docs/community/bindings.md
jatorre: Let me know what you think of this

isaacbrodsky (approved):

PR #615, opened 2022-06-12
Add E_VALUE and E_MEMORY / E_MEMORY_ALLOC

ajfriend:
Just to make the discussion concrete, this is what I was thinking around adding some error cases to fill out the error hierarchy. Related to https://github.com/uber/h3-py/pull/260

- E_FAILED
    + E_VALUE
        * E_DOMAIN
            - E_LATLNG_DOMAIN
            - E_RES_DOMAIN
        * E_RES_DOMAIN
        * E_CELL_INVALID
        * ...
        * E_OPTION_INVALID
    + E_MEMORY
        * E_MEMORY_ALLOC
        * E_MEMORY_BOUNDS

### Benefits

- this will line up exactly with the Python Exception hierarchy, so we won't need to document each separately; we can just have the `h3-py` documentation point to the C docs

### TODO (if accepted)

- [ ] update documentation

### DIFF
--- a/src/h3lib/include/h3api.h.in
+++ b/src/h3lib/include/h3api.h.in
@@ -82,26 +82,30 @@ typedef enum {
     E_SUCCESS = 0,  // Success (no error)
     E_FAILED =
         1,  // The operation failed but a more specific error is not available
-    E_DOMAIN = 2,  // Argument was outside of acceptable range (when a more
+    E_VALUE =
+        2,  // Value error (when a more specific error code is not available)
+    E_DOMAIN = 3,  // Argument was outside of acceptable range (when a more
                    // specific error code is not available)
     E_LATLNG_DOMAIN =
-        3,  // Latitude or longitude arguments were outside of acceptable range
-    E_RES_DOMAIN = 4,    // Resolution argument was outside of acceptable range
-    E_CELL_INVALID = 5,  // `H3Index` cell argument was not valid
-    E_DIR_EDGE_INVALID = 6,  // `H3Index` directed edge argument was not valid
+        4,  // Latitude or longitude arguments were outside of acceptable range
+    E_RES_DOMAIN = 5,    // Resolution argument was outside of acceptable range
+    E_CELL_INVALID = 6,  // `H3Index` cell argument was not valid
+    E_DIR_EDGE_INVALID = 7,  // `H3Index` directed edge argument was not valid
     E_UNDIR_EDGE_INVALID =
-        7,                 // `H3Index` undirected edge argument was not valid
-    E_VERTEX_INVALID = 8,  // `H3Index` vertex argument was not valid
-    E_PENTAGON = 9,  // Pentagon distortion was encountered which the algorithm
-                     // could not handle it
-    E_DUPLICATE_INPUT = 10,  // Duplicate input was encountered in the arguments
+        8,                 // `H3Index` undirected edge argument was not valid
+    E_VERTEX_INVALID = 9,  // `H3Index` vertex argument was not valid
+    E_PENTAGON = 10,  // Pentagon distortion was encountered which the algorithm
+                      // could not handle it
+    E_DUPLICATE_INPUT = 11,  // Duplicate input was encountered in the arguments
                              // and the algorithm could not handle it
-    E_NOT_NEIGHBORS = 11,    // `H3Index` cell arguments were not neighbors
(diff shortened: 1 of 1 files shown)


ajfriend:
I might even suggest a new category of errors: errors due to pentagon distortion or hexagons being too far away. This concept overlaps with the `safe` and `unsafe` versions of functions. And these errors may go away for specific functions in the future if we update the algorithms to handle these trickier cases.

- E_GRID_PROBLEM
    - E_PENTAGON
    - E_TOO_FAR

Obviously, open to ideas for better names. :)

I'm imagining `E_TOO_FAR` might come up around functions like `gridDistance` or `gridPath`, where they've failed because the hexes were too far apart in terms of icosahedron faces, instead of pentagon distortion. However, looking at the code, i'm not sure that's still actually an issue or possible error. @nrabinowitz, maybe you're a bit more familiar with the code and would be able to comment on if something like `E_TOO_FAR` would make sense?

isaacbrodsky:
I am not sure what the purpose of having these codes in the C library is. While Python may implement errors as a hierarchy these codes would be unreferenced in C, since the errors are not hierarchical here. We could add the Python exception classes to the C docs. Since the intermediary exceptions would not be thrown, there would be no need to list them in that same table. Do we have examples of exception class hierarchies being aligned across bindings?

I agree with your comment that we should clearly document when E_PENTAGON means that a function should be retried with the `safe` version as that's critical for correct implementation. Maybe we could have E_PENTAGON (no retry possible) and E_RETRY_WITH_SAFE (regardless of cause for when to signal to the application to retry)?

Regarding E_TOO_FAR, this type of error should still be possible with anything using the LocalIJ system, including gridDistance and gridPathCells.

ajfriend:
In the C world, these errors would be hierarchical only in a logical sense. In Python, you'd have the logical sense and the class hierarchy sense.

However, even though Python may have been the motivation, I think this change makes sense even if we forget about Python C. So let's do that and just focus on wether this makes sense for the C lib alone.

(Also, I'm not necessarily even arguing for this specific change at this point. I'm more interested in clarifying my understanding of our error model.)

For me, the idea of the hierarchy comes out of the "when a more specific error code is not available" language we currently have. Doesn't that imply a hierarchy?

And if `E_MEMORY` doesn't make sense to add because it wouldn't be referenced, couldn't you make a similar argument for something like `E_FAILED` or `E_DOMAIN`?

ajfriend:
Let me back off even further from the specific changes in this PR. So forget the specific code changes for now.

Would we all agree, even if we don't add any new parent/grouping errors, on the following hierarchy/grouping, just in a **logical** sense:

- h3 errors: `E_FAILED`
    - value errors:
        - domain errors: `E_DOMAIN`
            - `E_LATLNG_DOMAIN`
            - `E_RES_DOMAIN`
        - `E_CELL_INVALID`
        - `E_DIR_EDGE_INVALID`
        - `E_UNDIR_EDGE_INVALID`
        - `E_VERTEX_INVALID`
        - `E_DUPLICATE_INPUT`
        - `E_NOT_NEIGHBORS`
        - `E_RES_MISMATCH`
        - `E_OPTION_INVALID`
    - grid problem errors:
        - `E_PENTAGON`
        - `E_TOO_FAR`
        - `E_RETRY_WITH_SAFE` (but this may overlap with the other errors in this group)
    - memory errors:
        - `E_MEMORY` (maybe renamed to something more specific about allocation)
        - `E_MEMORY_BOUNDS`

I think this grouping makes sense, even if we don't have a specific error code for each parent node. Would you all agree? For instance, this might just be a documentation adjustment to group the errors logically.

nrabinowitz:
I think this makes some sense, but I don't know if it needs to be in the lib. The key question for me (drawing on e.g. HTTP error codes) would be: Is this an input error (what you have as `VALUE`), a system error (e.g. `MEMORY`, not sure if there's anything else in this category), or a library limitation that might be removed in the future (what you have as "grid problems", though really this is closer to "not yet implemented" maybe?). I can see the benefit of this, since users might address these classes of error differently, either in code or while developing/debugging.

I don't think it's necessary to have these in order in the enum, and in fact this makes it less flexible (e.g. adding a new value/input error would change further enum values, which is a breaking change to the API). I think this is purely a binding/documentation concern, unless we think that the intermediate "unspecified" errors would actually be used in the library (which I can't really see, TBH - we generally throw these errors when we know what went wrong). The only core lib changes that seem useful here are possibly updating the enum names to reflect a hierarchy, and updating the docs to organize the errors.

PR #616, opened 2022-06-13
Added on the highlight section the fact that H3 cells are more or les…

jatorre:
…s equals size

One important reason to choose H3 over other global grid systems is that the cells are more or less equal size regardless of location. That makes it well suited for global analysis or global datasets.

### DIFF
--- a/website/docs/README.md
+++ b/website/docs/README.md
@@ -16,6 +16,7 @@ The H3 Core Library implements the H3 grid system. It includes functions for con
 * H3 is a hierarchical [geospatial index](/docs/highlights/indexing).
 * H3 was developed to address the [challenges of Uber's data science needs](/docs/highlights/aggregation).
 * H3 can be used to [join disparate data sets](/docs/highlights/joining).
+* H3 cells are more or less equal size at different latitudes
 * In addition to the benefits of the hexagonal grid shape, H3 includes features for [modeling flow](/docs/highlights/flowmodel).
 * H3 is well suited to apply [ML to geospatial data](/docs/highlights/ml).
 


isaacbrodsky (commented):
I think this isn't too far an assertion - H3 is certainly does not have the same issue as e.g. Mercator where things distort wildly near the poles. That said, H3 isn't equal area. The determination of whether areas are "roughly the same" is subjective. What do others think of adding this?

ajfriend:
We could add a link to https://h3geo.org/docs/next/core-library/restable#hexagon-min-and-max-areas

ajfriend:
I'm a concerned with the fuzziness of the phrasing "more or less". I don't think a new user, reading this summary, would know what to make of that. What about phrasing that tries to quantify the similarity/distortion?

Something along the lines of:

H3 cells have similar surface area everywhere on the globe ([no more than 2x difference](https://h3geo.org/docs/next/core-library/restable/#hexagon-min-and-max-areas))

Also, the phrase "at different latitudes" suggests a contrast with a different system (S2?). I'd either drop "at different latitudes" or add some extra explanation as to why that's especially important in our case. (But totally open to hear other opinions!)

nrabinowitz:
> I'm a concerned with the fuzziness of the phrasing "more or less". I don't think a new user, reading this summary, would know what to make of that. What about phrasing that tries to quantify the similarity/distortion?
> 
> Something along the lines of:
> 
> H3 cells have similar surface area everywhere on the globe ([no more than 2x difference](https://h3geo.org/docs/next/core-library/restable/#hexagon-min-and-max-areas))
> 
> Also, the phrase "at different latitudes" suggests a contrast with a different system (S2?), or, at least, a specific property that is typically not satisfied. I'd either drop "at different latitudes" or add some extra explanation as to why that's especially important in our case. (But totally open to hear other opinions!)

I prefer this phrasing, and the link to the table. Presumably the latitude issue is referring to graticule-based grids (i.e. using lat/lng as Cartesian coords), which typically have distortion based on latitude.

isaacbrodsky:
I think "more or less" may be too vague for the website. I saw there was a paper published recently that compares various DGGS including on this criteria - https://dx.doi.org/10.1080/20964471.2022.2094926 - perhaps we could look to that to guide how to express this.

PR #619, opened 2022-07-03
Add E_TOO_FAR and E_RETRY_WITH_SAFE

ajfriend:


ajfriend:
Question: Whenever we'd consider returning `E_RETRY_WITH_SAFE`, it it the case that the reason the error happened is that we ran into a pentagon? Or, are there multiple reasons, and we might want to also relay that information along in the error codes somehow?

src/h3lib/lib/algos.c
```diff
@@ -1142,7 +1142,7 @@ H3Error H3_EXPORT(cellsToLinkedMultiPolygon)(const H3Index *h3Set,
     }
     _vertexGraphToLinkedGeo(&graph, out);
     if (normalizeMultiPolygon(out)) {
-        return E_FAILED;
+        return E_FAILED; // Q: normalizeMultiPolygon returns error codes, should we forward those to the public API?
```
nrabinowitz, line 1145: Yes, I assume that would be better. It looks like these didn't get rolled into the generalized error enum, so we'd need to move them there. It also looks like at least `normalizeMultiPolygon` and maybe some others return `int` instead of `H3Error`, which we should also update.

src/h3lib/lib/h3Index.c
```diff
@@ -784,7 +784,7 @@ H3Error H3_EXPORT(latLngToCell)(const LatLng *g, int res, H3Index *out) {
     if (*out) {
         return E_SUCCESS;
     } else {
-        return E_FAILED;
+        return E_FAILED;  // todo: under what circumstances would this failure happen?
```
nrabinowitz, line 787: It's not clear to me - looks like it will fail if the `faceIjk` object has out-of-bounds coordinates, but I can't easily tell whether that's a real case with bad coordinates or a fail-safe.

src/h3lib/lib/linkedGeo.c
```diff
@@ -294,7 +294,7 @@ int normalizeMultiPolygon(LinkedGeoPolygon *root) {
     // We assume that the input is a single polygon with loops;
     // if it has multiple polygons, don't touch it
     if (root->next) {
-        return NORMALIZATION_ERR_MULTIPLE_POLYGONS;
+        return NORMALIZATION_ERR_MULTIPLE_POLYGONS; // Do we want to add these error codes to the public API?
```
nrabinowitz, line 297: I think we should, though I think at present the user will never encounter these. The current input for `normalizeMultiPolygon` comes from `h3SetToVertexGraph`, so I don't think these error codes are reachable through the public API unless there's an error within the library. I'm not sure what to do with these internal errors, but I think it makes sense to unify them with external errors in order to a) have one approach to error handling, and b) not require a refactor if this ever is exposed in a way that could be prone to user input errors. 

That said, we might want to note in the docs that these errors are internal fail-safes, and if you hit them you should file a bug.

src/h3lib/lib/localij.c
```diff
@@ -204,7 +204,7 @@ H3Error cellToLocalIjk(H3Index origin, H3Index h3, CoordIJK *out) {
                 // TODO: We may be unfolding the pentagon incorrectly in this
                 // case; return an error code until this is guaranteed to be
                 // correct.
-                return E_FAILED;
+                return E_PENTAGON;  // invalid inputs or algo improvements needed?
```
nrabinowitz, line 207: I feel like this needs a different error, e.g. `E_UNREACHABLE_CELL` (could replace `E_TOO_FAR`). In other contexts, `E_PENTAGON` means we hit a pentagon and the algo can't handle it - in this case it's more specific, only certain directions around a pentagon are unreachable, and pentagons themselves are not failure cases.

src/h3lib/lib/localij.c
```diff
@@ -319,14 +319,14 @@ H3Error localIjkToCell(H3Index origin, const CoordIJK *ijk, H3Index *out) {
...
-            return E_FAILED;
+            return E_DOMAIN;
         }
 
         const Direction dir = _unitIjkToDigit(ijk);
         const int newBaseCell = _getBaseCellNeighbor(originBaseCell, dir);
         if (newBaseCell == INVALID_BASE_CELL) {
             // Moving in an invalid direction off a pentagon.
-            return E_FAILED;
+            return E_DOMAIN;  // i'm assuming this means the input values were invalid?
```
nrabinowitz, line 329: No, I think this is `E_UNREACHABLE_CELL` again - a direction the algo can't handle.

src/h3lib/lib/linkedGeo.c
ajfriend: Makes sense.

src/h3lib/lib/localij.c
ajfriend: What would you think about breaking `E_PENTAGON` into two different errors, to express either:

1. We hit a pentagon, and the math was tricky, so there might be a "safe" function you can use to recover from this, or such a function may get implemented in the future
2. We hit a pentagon, but something is actually wrong here and this error is not recoverable

Would either of those be covered by what you were thinking with `E_UNREACHABLE_CELL`?

src/h3lib/lib/localij.c
ajfriend: Would we just want an error denoting exactly what this comment is saying? `E_INVALID_PENTAGON_DIRECTION`? Or something like that?

Or is that exactly what you mean by `E_UNREACHABLE_CELL`?

src/h3lib/lib/localij.c
nrabinowitz: See my other comment - in this case, the pentagon is an unnecessary detail to the main issue: We can't reach the cell you want. I don't want to expose the exact reasons here, because they are not fundamental to the grid, only to our partial implementation. All the user needs/wants to know is that some cases aren't handled, and this is one of them - we actually don't want to encourage them to try to find weird workarounds, because they would be totally dependent on our current, flawed implementation.

src/h3lib/lib/algos.c
ajfriend: Moved discussion to https://github.com/uber/h3/issues/628

src/h3lib/lib/linkedGeo.c
ajfriend: Moved discussion to https://github.com/uber/h3/issues/628

src/h3lib/lib/localij.c
ajfriend: https://github.com/uber/h3/issues/627

src/h3lib/lib/localij.c
ajfriend: https://github.com/uber/h3/issues/627

src/h3lib/lib/algos.c
```diff
@@ -563,7 +563,7 @@ H3Error H3_EXPORT(gridDiskDistancesUnsafe)(H3Index origin, int k, H3Index *out,
 
     if (H3_EXPORT(isPentagon)(origin)) {
         // Pentagon was encountered; bail out as user doesn't want this.
-        return E_PENTAGON;
```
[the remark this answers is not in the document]
ajfriend: Moved discussion to https://github.com/uber/h3/issues/627

PR #633, opened 2022-08-03
Add changelog for v4.0.0 release

isaacbrodsky:
H3 version 4.0.0 is expected to be released on ~~Tuesday August 16, 2022~~ Tuesday August 23, 2022.

PR #635, opened 2022-08-03
Set v4 as a released version on the website

isaacbrodsky:
Depends on #633.


nrabinowitz (approved):

dfellis:
I presume this also depends on the various bindings releasing v4.0.0?

isaacbrodsky:
> I presume this also depends on the various bindings releasing v4.0.0?

I think so? Otherwise most of the code samples it would direct users to will not be correct.

dfellis:
> > I presume this also depends on the various bindings releasing v4.0.0?
> 
> I think so? Otherwise most of the code samples it would direct users to will not be correct.

It wasn't explicitly called out, so I wasn't sure if I :heavy_check_mark:'ed this PR that it would immediately go out when that's probably not the right thing to do, yet.

dfellis (approved):

ajfriend (approved):

isaacbrodsky:
> > > I presume this also depends on the various bindings releasing v4.0.0?
> > 
> > 
> > I think so? Otherwise most of the code samples it would direct users to will not be correct.
> 
> It wasn't explicitly called out, so I wasn't sure if I ✔️'ed this PR that it would immediately go out when that's probably not the right thing to do, yet.

Yes, the idea is to merge #633 first on August 16, tag & release uber/h3 v4.0.0, update & release binding updates, and then merge this PR. (Releasing binding updates and this PR might be parallelized)

PR #636, opened 2022-08-08
Fix undefined behavior in polygonToCells test

isaacbrodsky:
* Fixes a test that had an invalid read. This was found by running the tests with `ENABLE_FUZZERS`, which enabled ASAN etc. We don't run this in CI but perhaps should.
* Added a couple tests that exercise what happens when non-finite values are passed into polygonToCells. We previously had some casts from double to int64_t that were unsafe if the value was not finite because that cast would be undefined behavior.

Edit: happy to break this out into two PRs if that's easier.

Edit 2: This does not block #633

### DIFF
--- a/src/apps/testapps/testPolygonToCells.c
+++ b/src/apps/testapps/testPolygonToCells.c
@@ -476,13 +476,16 @@ SUITE(polygonToCells) {
         int64_t numHexagons;
         t_assertSuccess(H3_EXPORT(maxPolygonToCellsSize)(&pointGeoPolygon, 9, 0,
                                                          &numHexagons));
-        t_assert(numHexagons == 13, "Point has expected 1 hexagon");
+        // 0 estimation, plus 1 vertex, plus 12 buffer
+        t_assert(numHexagons == 13, "Point has expected 13 hexagons");
 
         H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
         t_assertSuccess(
             H3_EXPORT(polygonToCells)(&pointGeoPolygon, 9, 0, hexagons));
-        t_assert(hexagons[0] == H3_NULL,
-                 "no results for polygonToCells of a single point");
+        for (int i = 0; i < numHexagons; i++) {
+            t_assert(hexagons[i] == H3_NULL,
+                     "no results for polygonToCells of a single point");
+        }
         free(hexagons);
     }
 }
--- a/src/apps/testapps/testPolygonToCells.c
+++ b/src/apps/testapps/testPolygonToCells.c
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include <assert.h>
+#include <math.h>
 #include <stdlib.h>
 
 #include "algos.h"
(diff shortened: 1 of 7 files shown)


src/h3lib/lib/bbox.c
```diff
@@ -95,9 +95,10 @@ double _hexRadiusKm(H3Index h3Index) {
...
  * @param res the resolution of the H3 hexagons to fill the bounding box
- * @return the estimated number of hexagons to fill the bounding box
+ * @param out the estimated number of hexagons to fill the bounding box
+ * @return E_SUCCESS (0) on success, or another value otherwise.
  */
-int64_t bboxHexEstimate(const BBox *bbox, int res) {
+H3Error bboxHexEstimate(const BBox *bbox, int res, int64_t *out) {
     // Get the area of the pentagon as the maximally-distorted area possible
     H3Index pentagons[12] = {0};
     // TODO: Return error here
```
nrabinowitz, line 104: Can we address this TODO as well? Might be unreachable, but we can now return the error if any

src/h3lib/lib/bbox.c
```diff
@@ -123,30 +124,42 @@ int64_t bboxHexEstimate(const BBox *bbox, int res) {
...
+ * trace the line
+ * @return E_SUCCESS (0) on success or another value otherwise.
  */
-int64_t lineHexEstimate(const LatLng *origin, const LatLng *destination,
-                        int res) {
+H3Error lineHexEstimate(const LatLng *origin, const LatLng *destination,
+                        int res, int64_t *out) {
     // Get the area of the pentagon as the maximally-distorted area possible
     H3Index pentagons[12] = {0};
     // TODO: Return error here
```
nrabinowitz, line 152: Ditto

src/apps/testapps/testPolygonToCells.c
```diff
@@ -420,4 +436,53 @@ SUITE(polygonToCells) {
...
+                                           hexagons) == E_FAILED,
+                 "Invalid geo polygon cannot be evaluated");
+        free(hexagons);
+    }
+
+    TEST(polygonToCellsPoint) {
+        int64_t numHexagons;
+        t_assertSuccess(H3_EXPORT(maxPolygonToCellsSize)(&pointGeoPolygon, 9, 0,
+                                                         &numHexagons));
+        t_assert(numHexagons == 13, "Point has expected 1 hexagon");
```
nrabinowitz, line 479: Nit: Assertion and message don't seem to line up?

src/apps/testapps/testPolygonToCells.c
```diff
@@ -420,4 +436,53 @@ SUITE(polygonToCells) {
...
+
+        H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
+        t_assertSuccess(
+            H3_EXPORT(polygonToCells)(&pointGeoPolygon, 9, 0, hexagons));
+        t_assert(hexagons[0] == H3_NULL,
```
nrabinowitz, line 484: Nit: Do we need to check all of the 13 indexes?

nrabinowitz (approved):

dfellis (approved):

PR #638, opened 2022-08-10
Fix comment of M_180_PI

isbang:
I fixed comment of M_180_PI. Should I update CHANGELOG.md?

```
// calculated using golang
180 * pi : 565.4866776461628
180 / pi : 57.29577951308232
```


isaacbrodsky (approved):

nrabinowitz (approved):

nrabinowitz:
Thanks for the fix! I don't think a CHANGELOG entry is needed on this one, since it doesn't change anything consumer-facing.

isaacbrodsky:
> Thanks for the fix! I don't think a CHANGELOG entry is needed on this one, since it doesn't change anything consumer-facing.

That's also what I did in a recent comment and test-only change.

PR #639, opened 2022-08-15
exactEdgeLength -> edgeLength

ajfriend:
Fix a naming consistency issue. `edgeLength` now lines up with `cellArea`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -232,7 +232,7 @@ set(OTHER_SOURCE_FILES
     src/apps/fuzzers/fuzzerCellsToLinkedMultiPolygon.c
     src/apps/fuzzers/fuzzerDistances.c
     src/apps/fuzzers/fuzzerCellArea.c
-    src/apps/fuzzers/fuzzerExactEdgeLength.c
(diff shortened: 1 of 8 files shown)


nrabinowitz:
I'm okay with this change - we'll need to update the bindings as well. For consistency, we ought to offer `getEdgeLengthAvg*` functions as well, but def not required for launch.

nrabinowitz (approved):

ajfriend:
> I'm okay with this change - we'll need to update the bindings as well. For consistency, we ought to offer `getEdgeLengthAvg*` functions as well, but def not required for launch.

We do already have `getHexagonEdgeLengthAvgM()`. Is that what you mean?

nrabinowitz:
> > I'm okay with this change - we'll need to update the bindings as well. For consistency, we ought to offer `getEdgeLengthAvg*` functions as well, but def not required for launch.
> 
> We do already have `getHexagonEdgeLengthAvgM()`. Is that what you mean?

Yes, nvm 🤦‍♂️

isaacbrodsky (approved):
There are other places in the docs the function rename needs to be noted. I'll address them in a PR to release the next RC.

PR #642, opened 2022-08-19
Cover getPentagons res argument being invalid

isaacbrodsky:
Mainly-test only change to cover an additional line. The core library change is adding a coverage exclusion comment on a line we already excluded the branch coverage for. I plan to merge after v4.0.0.

### DIFF
--- a/src/apps/testapps/testPentagonIndexes.c
+++ b/src/apps/testapps/testPentagonIndexes.c
@@ -56,6 +56,16 @@ SUITE(getPentagons) {
         }
     }
 
+    TEST(getPentagonsInvalid) {
+        H3Index h3Indexes[PADDED_COUNT] = {0};
+        t_assert(H3_EXPORT(getPentagons)(16, h3Indexes) == E_RES_DOMAIN,
+                 "getPentagons of invalid resolutions fails");
+        t_assert(H3_EXPORT(getPentagons)(100, h3Indexes) == E_RES_DOMAIN,
+                 "getPentagons of invalid resolutions fails");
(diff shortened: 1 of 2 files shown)


dfellis (approved):

src/apps/testapps/testPentagonIndexes.c
```diff
@@ -56,6 +56,16 @@ SUITE(getPentagons) {
         }
     }
 
+    TEST(getPentagonsInvalid) {
+        H3Index h3Indexes[PADDED_COUNT] = {0};
+        t_assert(H3_EXPORT(getPentagons)(16, h3Indexes) == E_RES_DOMAIN,
+                 "getPentagons of invalid resolutions fails");
+        t_assert(H3_EXPORT(getPentagons)(100, h3Indexes) == E_RES_DOMAIN,
```
nrabinowitz, line 63: For my info, is there any difference between this test and the one above? Having two tests implies two cases, but I think there's only one here.

nrabinowitz (approved):

src/apps/testapps/testPentagonIndexes.c
isaacbrodsky: They should be the same code path, just different order of magnitude of the resolution parameter.

dfellis: Well, we do bit-packing in this library and many values are constrained well below an integer, so I can see a difference between a 5-bit value (the 16) and a 7-bit value (the 100) for our testing. If we simply lopped off the high bits they'd be turned into res 0 and res 4, respectively.

Issue #670, opened 2022-09-04
Signed integer overflow in fuzzerGridDisk


isaacbrodsky:
```
$ ./bin/fuzzerGridDisk
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3121615734
INFO: Loaded 1 modules   (6373 inline 8-bit counters): 6373 [0x5599618a3b68, 0x5599618a544d), 
INFO: Loaded 1 PC tables (6373 PCs): 6373 [0x5599618a5450,0x5599618be2a0), 
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
#2	INITED cov: 2 ft: 2 corp: 1/1b exec/s: 0 rss: 31Mb
	NEW_FUNC[1/1]: 0x5599617a8410 in maxGridDiskSize /home/rocky/oss/h3/src/h3lib/lib/algos.c:161
#1331	NEW    cov: 7 ft: 8 corp: 2/17b lim: 17 exec/s: 0 rss: 32Mb L: 16/16 MS: 4 ChangeByte-InsertRepeatedBytes-CopyPart-InsertRepeatedBytes-
#1334	NEW    cov: 13 ft: 14 corp: 3/33b lim: 17 exec/s: 0 rss: 32Mb L: 16/16 MS: 3 CrossOver-ChangeByte-ChangeByte-
	NEW_FUNC[1/10]: 0x5599617a8720 in gridDisk /home/rocky/oss/h3/src/h3lib/lib/algos.c:182
	NEW_FUNC[2/10]: 0x5599617a87b0 in gridDiskDistances /home/rocky/oss/h3/src/h3lib/lib/algos.c:204
#1367	NEW    cov: 87 ft: 95 corp: 4/49b lim: 17 exec/s: 0 rss: 38Mb L: 16/16 MS: 3 EraseBytes-InsertRepeatedBytes-EraseBytes-
	NEW_FUNC[1/2]: 0x5599617e6610 in _h3LeadingNonZeroDigit /home/rocky/oss/h3/src/h3lib/lib/h3Index.c:586
	NEW_FUNC[2/2]: 0x5599617ea870 in isResolutionClassIII /home/rocky/oss/h3/src/h3lib/lib/h3Index.c:1050
[...]

PR #671, opened 2022-09-04
Prevent negative sizes in fuzzerGridDisk

isaacbrodsky:
Fixes the progression of the fuzzer blocked in #670.

### DIFF
--- a/src/apps/fuzzers/fuzzerGridDisk.c
+++ b/src/apps/fuzzers/fuzzerGridDisk.c
@@ -26,7 +26,7 @@ typedef struct {
     int64_t k;
 } inputArgs;
 
-const int64_t MAX_GRID_DISK_SIZE = 0x10000000;
+const int64_t MAX_GRID_DISK_SIZE = 0x1000000;
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     if (size < sizeof(inputArgs)) {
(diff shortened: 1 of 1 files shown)


dfellis (approved):

src/apps/fuzzers/fuzzerGridDisk.c
```diff
@@ -36,8 +36,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 
     int64_t sz;
     H3Error err = H3_EXPORT(maxGridDiskSize)(args->k, &sz);
-    if (err || sz > MAX_GRID_DISK_SIZE) {
-        // Can't allocate
+    if (err || sz > MAX_GRID_DISK_SIZE || sz < 0) {
```
nrabinowitz, line 39: For my info - is any corresponding check needed in `gridDisk` itself for bad input?

nrabinowitz (approved):

src/apps/fuzzers/fuzzerGridDisk.c
isaacbrodsky: Partially yes. If the `k` value is such that an overflow will occur, we need to guard against it. There are however `k` values which are valid but simply too large to practically use, which I don't think we should guard against. We should instead document the time complexity so users can be aware of this.

PR #672, opened 2022-09-04
Add a combined fuzzer

isaacbrodsky:
Creates a fuzzer that runs against the entire H3 API surface; not sure how useful this will be as I would like to get all the fuzzers currently in the H3 repo run by oss-fuzz which would take care of ensuring these are all run at once.


isaacbrodsky:
Closing the combined fuzzer out as it doesn't seem to be needed for OSS-Fuzz.

Issue #669, opened 2022-09-04
Leak detected in cellsToLinkedMultiPolygon fuzzer


isaacbrodsky:
This case is detected very quickly using the `fuzzerCellsToLinkedMultiPolygon` fuzzer.
```
=================================================================
==108804==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x55c5fad43d58 in __interceptor_calloc (/home/rocky/oss/h3/build/bin/fuzzerCellsToLinkedMultiPolygon+0x129d58) (BuildId: d7f817ee8f60e0189139805158a515806db6cf39)
    #1 0x55c5fadf04f7 in initVertexGraph /home/rocky/oss/h3/src/h3lib/lib/vertexGraph.c:39:26
    #2 0x55c5fad9cbc7 in h3SetToVertexGraph /home/rocky/oss/h3/src/h3lib/lib/algos.c:1069:5
    #3 0x55c5fad9dce8 in cellsToLinkedMultiPolygon /home/rocky/oss/h3/src/h3lib/lib/algos.c:1147:19
    #4 0x55c5fad7ea5a in LLVMFuzzerTestOneInput /home/rocky/oss/h3/src/apps/fuzzers/fuzzerCellsToLinkedMultiPolygon.c:34:19
    #5 0x55c5faca74f3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/rocky/oss/h3/build/bin/fuzzerCellsToLinkedMultiPolygon+0x8d4f3) (BuildId: d7f817ee8f60e0189139805158a515806db6cf39)
    #6 0x55c5faca6c49 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) (/home/rocky/oss/h3/build/bin/fuzzerCellsToLinkedMultiPolygon+0x8cc49) (BuildId: d7f817ee8f60e0189139805158a515806db6cf39)
    #7 0x55c5faca8439 in fuzzer::Fuzzer::MutateAndTestOne() (/home/rocky/oss/h3/build/bin/fuzzerCellsToLinkedMultiPolygon+0x8e439) (BuildId: d7f817ee8f60e0189139805158a515806db6cf39)
[...]

PR #673, opened 2022-09-04
Add test for cellsToLinkedMultipolygon leak

isaacbrodsky:
Adds a test #669.

### DIFF
--- a/src/h3lib/lib/linkedGeo.c
+++ b/src/h3lib/lib/linkedGeo.c
@@ -290,20 +290,20 @@ static const LinkedGeoPolygon *findPolygonForHole(
  * @param root Root polygon including all loops
  * @return     0 on success, or an error code > 0 for invalid input
  */
-int normalizeMultiPolygon(LinkedGeoPolygon *root) {
+H3Error normalizeMultiPolygon(LinkedGeoPolygon *root) {
     // We assume that the input is a single polygon with loops;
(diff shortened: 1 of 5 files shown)


dfellis (approved):

src/h3lib/lib/linkedGeo.c
```diff
@@ -290,20 +290,20 @@ static const LinkedGeoPolygon *findPolygonForHole(
  * @param root Root polygon including all loops
  * @return     0 on success, or an error code > 0 for invalid input
  */
-int normalizeMultiPolygon(LinkedGeoPolygon *root) {
+H3Error normalizeMultiPolygon(LinkedGeoPolygon *root) {
     // We assume that the input is a single polygon with loops;
     // if it has multiple polygons, don't touch it
     if (root->next) {
-        return NORMALIZATION_ERR_MULTIPLE_POLYGONS;
```
nrabinowitz, line 297: Should we make new `H3Error` codes here? I'm not sure if it matters for the end user at this point, but this change drops information about the source of the error.

nrabinowitz (approved):

src/h3lib/lib/linkedGeo.c
isaacbrodsky: We could - that information was not exposed to the user before.

PR #674, opened 2022-09-05
Increase fuzzer runs

isaacbrodsky:
Depends on #673, #671.


nrabinowitz (approved):

isaacbrodsky:
It looks like 10,000 runs is taking too long for some fuzzer, so perhaps it's better to give it a time budget per-fuzzer.

dfellis (approved):

PR #679, opened 2022-09-08
fix: handle compacting base cell children

jogly:
Children of a res 0 cell should compact back to that cell successfully.

### DIFF
--- a/src/apps/testapps/testCompactCells.c
+++ b/src/apps/testapps/testCompactCells.c
@@ -85,9 +85,12 @@ SUITE(compactCells) {
             t_assert(children[i] != 0, "got expected child");
         }
 
-        H3Index *compressed = calloc(1, sizeof(H3Index));
+        H3Index *compressed = calloc(arrSize, sizeof(H3Index));
         t_assertSuccess(H3_EXPORT(compactCells(children, compressed, arrSize)));
         t_assert(compressed[0] == parent, "got expected parent");
+        t_assert(compressed[1] == 0, "expected only 1 cell");
+
+        free(compressed);
     }
 
     TEST(res0) {
--- a/src/apps/testapps/testCompactCells.c
+++ b/src/apps/testapps/testCompactCells.c
@@ -70,6 +70,26 @@ SUITE(compactCells) {
         free(sunnyvaleExpanded);
     }
 
+    TEST(res0children) {
+        H3Index parent;
+        setH3Index(&parent, 0, 0, 0);
+
+        int64_t arrSize;
+        t_assertSuccess(H3_EXPORT(cellToChildrenSize)(parent, 1, &arrSize));
+
+        H3Index *children = calloc(arrSize, sizeof(H3Index));
+        t_assertSuccess(H3_EXPORT(cellToChildren)(parent, 1, children));
+
+        t_assert(arrSize > 0, "got expected number of children");
+        for (int i = 0; i < arrSize; i++) {
+            t_assert(children[i] != 0, "got expected child");
+        }
+
+        H3Index *compressed = calloc(1, sizeof(H3Index));
+        t_assertSuccess(H3_EXPORT(compactCells(children, compressed, arrSize)));
+        t_assert(compressed[0] == parent, "got expected parent");
+    }
+
     TEST(res0) {
         int hexCount = NUM_BASE_CELLS;
 
--- a/src/apps/testapps/testCompactCells.c
+++ b/src/apps/testapps/testCompactCells.c
@@ -86,6 +86,7 @@ SUITE(compactCells) {
         t_assert(compressed[1] == 0, "expected only 1 cell");
 
         free(compressed);
+        free(children);
     }
 
     TEST(res0) {
--- a/src/apps/testapps/testCompactCells.c
+++ b/src/apps/testapps/testCompactCells.c
@@ -80,11 +80,6 @@ SUITE(compactCells) {
         H3Index *children = calloc(arrSize, sizeof(H3Index));
         t_assertSuccess(H3_EXPORT(cellToChildren)(parent, 1, children));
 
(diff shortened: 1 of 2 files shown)


src/apps/testapps/testCompactCells.c
```diff
@@ -70,6 +70,26 @@ SUITE(compactCells) {
...
+
+        H3Index *children = calloc(arrSize, sizeof(H3Index));
+        t_assertSuccess(H3_EXPORT(cellToChildren)(parent, 1, children));
+
+        t_assert(arrSize > 0, "got expected number of children");
+        for (int i = 0; i < arrSize; i++) {
+            t_assert(children[i] != 0, "got expected child");
+        }
+
+        H3Index *compressed = calloc(1, sizeof(H3Index));
```
isaacbrodsky, line 88: Doesn't `compressed` need to be of size arrSize per the function contract? https://h3geo.org/docs/api/hierarchy/#compactcells

Also, please add corresponding `free` calls for `calloc`.

jogly: good call, I wrote this fast. it should be yes, and I will change that but that doesn't affect the integrity of the test as there *should* only be 1 cell returned.

jogly: updated

isaacbrodsky:
It looks like around h3Index.c line 344 it tries to take the parent of a res 0 cell at res -1 (res -1 was calculated on line 325), which is where the failure is coming from.

jogly:
I'm uncertain of this solution, but it passes existing tests and the new failing test.

Basically if the compaction process reaches a base cell, we skip the parent reserved-bit-hit processing and move on to the rest of the logic.

src/h3lib/lib/h3Index.c
```diff
@@ -323,75 +323,84 @@ H3Error H3_EXPORT(compactCells)(const H3Index *h3Set, H3Index *compactedSet,
...
+                        H3_MEMORY(free)(remainingHexes);
+                        H3_MEMORY(free)(hashSetArray);
+                        return parentError;
+                    }
+                    // Modulus hash the parent into the temp array
+                    int loc = (int)(parent % numRemainingHexes);
+                    int loopCount = 0;
+                    while (hashSetArray[loc] != 0) {
+                        if (loopCount >
+                            numRemainingHexes) {  // LCOV_EXCL_BR_LINE
```
isaacbrodsky, line 362: Coveralls seems to be complaining of coverage drops because some of the branch exclusions, like these, got reformatted onto different lines, which changes how coverage is computed.

src/h3lib/lib/h3Index.c
```diff
@@ -323,75 +323,84 @@ H3Error H3_EXPORT(compactCells)(const H3Index *h3Set, H3Index *compactedSet,
...
-                int loopCount = 0;
-                while (hashSetArray[loc] != 0) {
-                    if (loopCount > numRemainingHexes) {  // LCOV_EXCL_BR_LINE
-                        // LCOV_EXCL_START
-                        // This case should not be possible because at most one
-                        // index is placed into hashSetArray per
-                        // numRemainingHexes.
+        // If parentRes is less than zero, we've compacted all the way up to the
+        // base cells. Time to process the remaining cells.
+        if (parentRes >= 0) {
```
isaacbrodsky, line 329: This seems like a good check to have, although I haven't reviewed this in full because of how complicated this function is. I guess the alternative is breaking at the end of the loop if we get to res 0.

This seems like a good target for reducing how nested the function is, or splitting out into more manageable blocks of code.

src/apps/testapps/testCompactCells.c
```diff
@@ -70,6 +70,24 @@ SUITE(compactCells) {
...
     }
 
+    TEST(res0children) {
+        H3Index parent;
+        setH3Index(&parent, 0, 0, 0);
+
+        int64_t arrSize;
+        t_assertSuccess(H3_EXPORT(cellToChildrenSize)(parent, 1, &arrSize));
+
+        H3Index *children = calloc(arrSize, sizeof(H3Index));
```
isaacbrodsky, line 80: This one also needs a `free` (which is why Valgrind is complaining)

jogly: just saw it ;)

src/h3lib/lib/h3Index.c
jogly: agree on all points, but I do not volunteer for that refactor <3

jogly: except breaking at the end of the loop. I'm not sure that's a safe thing to do. Just because this iteration's parentRes is < 0 doesnt mean the next iteration's will be too....I think.  I have no idea, but it seems possible and this way I didn't have to worry about it.

isaacbrodsky: Yeah no need to do a full refactor here.

src/h3lib/lib/h3Index.c
```diff
@@ -323,75 +323,84 @@ H3Error H3_EXPORT(compactCells)(const H3Index *h3Set, H3Index *compactedSet,
...
+                        H3_MEMORY(free)(remainingHexes);
+                        H3_MEMORY(free)(hashSetArray);
+                        return parentError;
+                    }
+                    // Modulus hash the parent into the temp array
+                    int loc = (int)(parent % numRemainingHexes);
+                    int loopCount = 0;
+                    while (hashSetArray[loc] != 0) {
+                        if (loopCount >           // LCOV_EXCL_BR_LINE
+                            numRemainingHexes) {  // LCOV_EXCL_BR_LINE
```
isaacbrodsky, line 362: I believe we only need the first LCOV_EXCL_BR_LINE but I can look at that later.

isaacbrodsky (approved):

dfellis (approved):

PR #685, opened 2022-09-11
Fix memory leak in compactCells

isaacbrodsky:
This branch was previously thought unreachable, but evidently is. It resulted in a memory leak which allowed us to detect it. I added a test based on a specific bad input and assert that it returns failure, plus rely on Valgrind to detect if a leak happens; it may be possible to minify the test case even further.

We should convert these "unreachable" branches to assertions, at least for debug builds so we can be sure to find other such cases in e.g. fuzzing.

### DIFF
--- a/src/apps/testapps/testCompactCells.c
+++ b/src/apps/testapps/testCompactCells.c
@@ -355,9 +355,7 @@ SUITE(compactCells) {
                          0x2100000000,
                          0x7,
                          0x400000000};
-        H3Index output[] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-                            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-                            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+        H3Index output[43] = {0};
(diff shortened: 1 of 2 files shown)


src/apps/testapps/testCompactCells.c
```diff
@@ -307,6 +307,62 @@ SUITE(compactCells) {
...
+                         0x20000000,
+                         0x5000000000,
+                         0x100321,
+                         0x20000000,
+                         0x5000000000,
+                         0x100321,
+                         0x2100000000,
+                         0x7,
+                         0x400000000};
+        H3Index output[] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
nrabinowitz, line 358: Nit: Maybe more legible as `H3Index output[43] = {0}`?

nrabinowitz (approved):

src/apps/testapps/testCompactCells.c
```diff
@@ -307,6 +307,60 @@ SUITE(compactCells) {
             "compactCells returns E_RES_MISMATCH on bad input (parent error)");
     }
 
+    TEST(compactCells_parentError2) {
+        // This test primary ensures memory is not leaked upon invalid input,
```
dfellis, line 311: Nit

```suggestion
        // This test primarily ensures memory is not leaked upon invalid input,
```

dfellis (approved):

PR #686, opened 2022-09-12
Prevent signed integer overflow in maxGridDiskSize

isaacbrodsky:
Signed integer overflow is undefined behavior in C, so we cannot add, multiply, etc. arbitrary integer inputs. This adds a hard cap on the maxGridDiskSize function based on knowledge of the grid system itself.

### DIFF
--- a/src/apps/testapps/testCompactCells.c
+++ b/src/apps/testapps/testCompactCells.c
@@ -83,7 +83,9 @@ SUITE(compactCells) {
         H3Index *compressed = calloc(arrSize, sizeof(H3Index));
         t_assertSuccess(H3_EXPORT(compactCells(children, compressed, arrSize)));
         t_assert(compressed[0] == parent, "got expected parent");
-        t_assert(compressed[1] == 0, "expected only 1 cell");
+        for (int idx = 1; idx < arrSize; idx++) {
+            t_assert(compressed[idx] == 0, "expected only 1 cell");
+        }
 
         free(compressed);
         free(children);
--- a/src/h3lib/lib/algos.c
+++ b/src/h3lib/lib/algos.c
@@ -152,6 +152,12 @@ static const Direction NEW_ADJUSTMENT_III[7][7] = {
      CENTER_DIGIT, IJ_AXES_DIGIT}};
 
 /**
+ * k value which will encompass all cells at resolution 15.
+ * This is the largest possible k in the H3 grid system.
+ */
+static const int K_ALL_CELLS_AT_RES_15 = 13780510;
+
+/**
  * Maximum number of cells that result from the gridDisk algorithm with the
  * given k. Formula source and proof: https://oeis.org/A003215
  *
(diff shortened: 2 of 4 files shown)


src/h3lib/lib/algos.c
```diff
@@ -162,6 +162,17 @@ H3Error H3_EXPORT(maxGridDiskSize)(int k, int64_t *out) {
     if (k < 0) {
         return E_DOMAIN;
     }
+    if (k >= 13780510) {
```
nrabinowitz, line 165: Nit: Should we assign this to a const?

nrabinowitz (approved):

src/h3lib/lib/algos.c
dfellis: And/or a collection of constants for each resolution? I assume a much lower `k` at res 0 will run into similar problems?

isaacbrodsky: As this function does not accept the resolution we have no way of knowing if the `k` value will be larger than the grid. This would require an API change.

`k` that generates >122 cells at resolution 0 is also not the immediate issue. The issue is `k` values like INT32_MAX which would result in signed integer overflow, which is undefined behavior.

dfellis: We can do that without breaking backwards compatibility by creating a new function that does take the resolution, and recommending it over the old one? That would help prevent memory over-allocation for the smaller resolutions, and we could mark the current one as deprecated (but still with this fix).

I do agree with @nrabinowitz in any case that this would be better as a self-descriptive constant variable name.

isaacbrodsky: Added a comment for the constant. I'd defer a new function which limits the size to a future PR.

src/apps/testapps/testCompactCells.c
```diff
@@ -83,7 +83,9 @@ SUITE(compactCells) {
         H3Index *compressed = calloc(arrSize, sizeof(H3Index));
         t_assertSuccess(H3_EXPORT(compactCells(children, compressed, arrSize)));
         t_assert(compressed[0] == parent, "got expected parent");
-        t_assert(compressed[1] == 0, "expected only 1 cell");
+        for (int idx = 1; idx < arrSize; idx++) {
+            t_assert(compressed[idx] == 0, "expected only 1 cell");
+        }
```
jogly, line 88: just curious why checking the second element isn't sufficient?  i'd expect the first `0` to "terminate" the array, no much how big it is.

src/h3lib/lib/h3Index.c
```diff
@@ -357,8 +357,8 @@ H3Error H3_EXPORT(compactCells)(const H3Index *h3Set, H3Index *compactedSet,
                     int loc = (int)(parent % numRemainingHexes);
                     int loopCount = 0;
                     while (hashSetArray[loc] != 0) {
-                        if (loopCount >           // LCOV_EXCL_BR_LINE
-                            numRemainingHexes) {  // LCOV_EXCL_BR_LINE
+                        if (loopCount >  // LCOV_EXCL_BR_LINE
+                            numRemainingHexes) {
```
jogly, line 361: thanks!

jogly (approved):

src/apps/testapps/testCompactCells.c
isaacbrodsky: I'm not sure that is a valid assumption for compactCells. We should document that clearly if it is. I don't believe that assumption holds for gridDisk specifically due to the use of a hash function.

jogly: good point! this has ramifications for bindings (at least for h3-go) where I "post-process" the internally allocated C buffer to fit the language's paradigms, e.g. I resize the returned slice of indexes to the first zero'd index for compaction...

PR #689, opened 2022-09-14
Add vcpkg installation instructions

Cheney-W:
`H3` available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `H3` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `H3`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/h3/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,17 @@ brew install h3
 ```
 Otherwise, to build H3 from source, please see the following instructions.
 
+### Installing from vcpkg
+You can download and install H3 using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+```sh
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.sh #.\bootstrap-vcpkg.bat(for windows)
+./vcpkg integrate install
+./vcpkg install h3
+```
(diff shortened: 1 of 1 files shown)


dfellis:
I don't know if I want to endorse this "vcpkg" tool in the repo like this. [The H3 version it installs is several years out of date](https://github.com/microsoft/vcpkg/blob/master/ports/h3/portfile.cmake#L6) and it is making opinionated decisions for users like deleting all of the utility binaries.

nrabinowitz:
> I don't know if I want to endorse this "vcpkg" tool in the repo like this. [The H3 version it installs is several years out of date](https://github.com/microsoft/vcpkg/blob/master/ports/h3/portfile.cmake#L6) and it is making opinionated decisions for users like deleting all of the utility binaries.

I don't see how this is significantly different from including Homebrew installation instructions? This seems fine to me, users who don't want a somewhat opinionated package manager have other options.

I agree that it would be great to at least install v3.7.2 (the latest H3 v3 release).

Cheney-W:
I saw the latest release of h3 is v4.0.0, I will update this port to this version.

PR #692, opened 2022-09-17
Check subtraction in fuzzerHierarchy

isaacbrodsky:
Prevent a possible signed integer underflow in fuzzerHierarchy.c.


dfellis (approved):

nrabinowitz (approved):
LG - does this imply we need a negative res check in some portion of the live code?

isaacbrodsky:
> LG - does this imply we need a negative res check in some portion of the live code?

This should already be checked in https://github.com/uber/h3/blob/619a2a7ff3b54c45d456c838dcceefb30163f062/src/h3lib/lib/iterators.c#L219. Strictly speaking we just need to ensure the subtraction doesn't underflow, so I can update this PR to pass in non-underflowing negative values.

PR #706, opened 2022-10-06
Fix a possible signed integer overflow in localIjToCell

isaacbrodsky:
This protects against a particular instance of signed integer overflow that would have occurred in `ijkNormalize`.

### DIFF
--- a/src/h3lib/include/coordijk.h
+++ b/src/h3lib/include/coordijk.h
@@ -108,7 +108,7 @@ Direction _rotate60ccw(Direction digit);
 Direction _rotate60cw(Direction digit);
 int ijkDistance(const CoordIJK *a, const CoordIJK *b);
 void ijkToIj(const CoordIJK *ijk, CoordIJ *ij);
-void ijToIjk(const CoordIJ *ij, CoordIJK *ijk);
+H3Error ijToIjk(const CoordIJ *ij, CoordIJK *ijk);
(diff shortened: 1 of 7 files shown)


src/h3lib/include/coordijk.h
```diff
@@ -108,7 +108,7 @@ Direction _rotate60ccw(Direction digit);
 Direction _rotate60cw(Direction digit);
 int ijkDistance(const CoordIJK *a, const CoordIJK *b);
 void ijkToIj(const CoordIJK *ijk, CoordIJ *ij);
-void ijToIjk(const CoordIJ *ij, CoordIJK *ijk);
+H3Error ijToIjk(const CoordIJ *ij, CoordIJK *ijk);
```
nrabinowitz, line 111: This makes me suspicious that we may need something similar for `ijkToIj` or the cube functions as well, might be worth adding tests.

nrabinowitz (approved):

src/h3lib/include/coordijk.h
isaacbrodsky: I am not sure if ijkToIj can be overflowed assuming that the input coordinates are all non-negative.

isaacbrodsky:
I don't think my original test was correct (coverage decreased) and I updated. TBD if coverage is complete on the new test.

dfellis (approved):

PR #710, opened 2022-10-12
ratio for calculation of bbox area should always be > 1 (#708)

sebisteiner:
Description and discussion is here: https://github.com/uber/h3/issues/708

PR #711, opened 2022-10-12
add Clang sanitizer checks in CI

isaacbrodsky:


dfellis (approved):

src/apps/testapps/testPolygonToCells.c
```diff
@@ -40,19 +40,27 @@ static GeoLoop holeGeoLoop = {.numVerts = 3, .verts = holeVerts};
 static GeoPolygon holeGeoPolygon;
 
 static LatLng emptyVerts[] = {{0.659966917655, -2.1364398519394},
-                              {0.659966917655, -2.1364398519395},
-                              {0.659966917655, -2.1364398519396}};
+                              {0.659966917656, -2.1364398519395},
+                              {0.659966917657, -2.1364398519396}};
```
nrabinowitz, line 44: Why did these need to be changed?

src/h3lib/lib/bbox.c
```diff
@@ -123,7 +123,12 @@ H3Error bboxHexEstimate(const BBox *bbox, int res, int64_t *out) {
     double d = H3_EXPORT(greatCircleDistanceKm)(&p1, &p2);
     // Derived constant based on: https://math.stackexchange.com/a/1921940
     // Clamped to 3 as higher values tend to rapidly drag the estimate to zero.
-    double a = d * d / fmin(3.0, fabs((p1.lng - p2.lng) / (p1.lat - p2.lat)));
+    double lngDiff = p1.lng - p2.lng;
+    double latDiff = p1.lat - p2.lat;
+    if (lngDiff == 0 || latDiff == 0) {
```
nrabinowitz, line 128: For my info, how are `NAN` and `Infinity` values handled here? Do we need to check for something like `isFinite` as well?

nrabinowitz (approved):

nrabinowitz:
Does this supersede https://github.com/uber/h3/pull/710?

src/h3lib/lib/bbox.c
isaacbrodsky: isfinite is checked below

src/apps/testapps/testPolygonToCells.c
isaacbrodsky: Possibly they didn't need to - I was trying to adjust some test cases to get better coverage although I don't think I really accomplished that.

isaacbrodsky:
> Does this supersede #710?

No, that PR appears to be addressing a different issue in the ratio calculation, it doesn't address the divide by zero problem.

src/apps/testapps/testPolygonToCells.c
isaacbrodsky: My mistake, they do need to be changed because as defined this is a straight line. polygonToCells will reject straight lines now - even if a cell centroid is along the line - because it would have attempted to divide by zero when estimating.

PR #715, opened 2022-10-14
[Docs] Ruby bindings GitHub URL has changed

seanhandley:
Moved from [StuartApp/h3_ruby](https://github.com/StuartApp/h3_ruby) to [seanhandley/h3_ruby](https://github.com/seanhandley/h3_ruby)


nrabinowitz (approved):
Thanks for the update!

PR #719, opened 2022-10-24
Feature: Add cellToChildPos and childPosToCell

nrabinowitz:
Adds two new functions:

- `cellToChildPos` gives the position of the child in an ordered list of all the children of the cell's parent at a given resolution. The "ordered list" here is the same order as returned by `cellToChildren`, which I believe is numerically sorted by index. Essentially, this function allows you to determine the numeric index of the cell among its siblings, given a parent resolution that defines the sibling list.
- `childPosToCell` offers the reverse functionality, returning a child cell given a parent, a child resolution, and a child index.

This pair of functions opens up several new use cases:
- Easy iteration through all children of some parent, with no need to allocate memory for `cellToChildren`. I think we can do this internally in the library using iterators, but this offers an ergonomic API for external consumers (simply iterate `for (int i = 0; i < cellToChildrenSize(parent, res); i++)` and take `childPosToCell(i, parent, res)`).
- Compact representations for sets of common-ancestor cells, including arrays of small bit-width integers (e.g. `uint16` if the parent/child res difference is <= 5) and bit arrays (e.g. 16807 bits / 2101 bytes to indicate yes/no whether each child 5 steps down is included in the set). This is likely more effective than `compact` if there's no particular likelihood that the cells are contiguous.
- Association between data arrays and H3 indexes without the need to store the H3 indexes themselves (e.g. a list of 16807 values corresponding by position to each child 5 steps down from some parent)

**Notes**
- I'm open to suggestions on the name. `childPosition` might be clearer (but longer), `childIndex` makes sense but risked confusion with H3 indexes.
- I included exhaustive tests for 0-3 resolution steps for all cells up to res 2. Let me know if you think any further tests are needed here.


src/h3lib/lib/h3Index.c
```diff
@@ -1052,3 +1052,113 @@ H3Error H3_EXPORT(getPentagons)(int res, H3Index *out) {
...
+    int parentIsPentagon = H3_EXPORT(isPentagon)(parent);
+
+    // Walk up the resolution digits, incrementing the index
+    *out = 0;
+    if (parentIsPentagon) {
+        // Pentagon logic. Pentagon parents skip the 1 digit, so the offsets are
+        // different from hexagons
+        for (int res = childRes; res > parentRes; res--) {
+            // It shouldn't be possible to get an error here, so we don't check
+            H3_EXPORT(cellToParent(child, res - 1, &parent));
```
nrabinowitz, line 1077: This is probably a good spot for `NEVER` once #720 lands

src/apps/testapps/testCellToChildPos.c
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright 2017-2021 Uber Technologies, Inc.
```
isaacbrodsky, line 2:
```suggestion
 * Copyright 2022 Uber Technologies, Inc.
```

src/h3lib/lib/h3Index.c
```diff
@@ -1052,3 +1052,113 @@ H3Error H3_EXPORT(getPentagons)(int res, H3Index *out) {
...
+    } else {
+        // Hexagon tile logic. Offsets are simple powers of 7
+        for (int res = 1; res <= resOffset; res++) {
+            int64_t resWidth = _ipow(7, resOffset - res);
+            H3_SET_INDEX_DIGIT(*child, parentRes + res, idx / resWidth);
+            idx %= resWidth;
+        }
+    }
+
+    return E_SUCCESS;
```
isaacbrodsky, line 1163: What happens if the passed in childPos is abnormally high or low? E.g. if `idx > 0` here or if `childPos < 0`?

src/h3lib/lib/h3Index.c
```diff
@@ -1052,3 +1052,113 @@ H3Error H3_EXPORT(getPentagons)(int res, H3Index *out) {
...
+    *out = 0;
+    if (parentIsPentagon) {
+        // Pentagon logic. Pentagon parents skip the 1 digit, so the offsets are
+        // different from hexagons
+        for (int res = childRes; res > parentRes; res--) {
+            // It shouldn't be possible to get an error here, so we don't check
+            H3_EXPORT(cellToParent(child, res - 1, &parent));
+            parentIsPentagon = H3_EXPORT(isPentagon)(parent);
+            int rawDigit = H3_GET_INDEX_DIGIT(child, res);
+            int digit =
```
isaacbrodsky, line 1080: What happens for invalid indexes into the deleted K subsequence of a pentagon, or indexes with INVALID_DIGIT (7)?

src/h3lib/lib/h3Index.c
```diff
@@ -1052,3 +1052,113 @@ H3Error H3_EXPORT(getPentagons)(int res, H3Index *out) {
...
+
+    return E_SUCCESS;
+}
+
+/**
+ * Returns the child cell at a given position within an ordered list of all
+ * children at the specified resolution */
+H3Error H3_EXPORT(childPosToCell)(int64_t childPos, H3Index parent,
+                                  int childRes, H3Index *child) {
+    if (childRes < 0 || childRes > MAX_H3_RES) {
```
isaacbrodsky, line 1113: This would also be a good place for `testcase` from #720 to ensure boundary conditions work as expected.

isaacbrodsky (commented):
I understand how these functions are used so they look helpful.

Please add website API documentation for these functions.

src/apps/testapps/testCellToChildPos.c
dfellis:
```suggestion
 * Copyright 2017-2022 Uber Technologies, Inc.
```

src/h3lib/lib/h3Index.c
nrabinowitz: Good questions:
- Invalid indexes into the K subsequence will be treated as `0` per the logic on the next line
- INVALID_DIGIT will result in an offset that's too high, and the resulting number might be greater than the number of max children

This _might_ be okay, on the principle of GIGO, but I could imagine attacks on systems _using_ H3 that would use poor output here to trigger random memory access. Possible failsafes:
- We could validate the digit and return an error
- We could clamp output to the max possible child index for the given parent

Validating the digits is probably worthwhile. If we do that we don't need to clamp output, but we could (after #720) add an assertion at the end of the function like
```
if (NEVER(out > cellToChildrenSize(parent, childRes))) {
  return E_FAILED;
}
```

src/h3lib/lib/h3Index.c
nrabinowitz: Another good question. You'd get an invalid digit for output; I'm not sure there would be other issues. Probably worth validating `childPos` and returning an error for bad cases.

dfellis (approved):

website/docs/api/hierarchy.mdx
```diff
@@ -220,7 +220,118 @@ function example() {
...
+    {label: 'C', value: 'c'},
+    {label: 'Python', value: 'python'},
+    {label: 'Java', value: 'java'},
+    {label: 'JavaScript (Live)', value: 'javascript'},
+  ]
+}>
+<TabItem value="c">
+
+```c
+H3Index cellToChildPos(H3Index child, int parentRes, int64_t *out);
```
isaacbrodsky, line 242: All the other bindings say "cell" - maybe they should also have the parameter named "child"?

src/h3lib/lib/h3Index.c
```diff
@@ -1050,3 +1050,162 @@ H3Error H3_EXPORT(getPentagons)(int res, H3Index *out) {
...
+/**
+ * Validate a child position in the context of a given parent, returning
+ * an error if validation fails.
+ */
+static H3Error validateChildPos(int64_t childPos, H3Index parent,
+                                int childRes) {
+    int64_t maxChildCount;
+    H3Error sizeError =
+        H3_EXPORT(cellToChildrenSize)(parent, childRes, &maxChildCount);
+    if (NEVER(sizeError)) {
```
isaacbrodsky, line 1063: In the future, we should require fuzzers be added too to try to exercise these. I am fine with merging this without the fuzzer and then adding it in a follow up PR.

isaacbrodsky (approved):

website/docs/api/hierarchy.mdx
nrabinowitz: This is all prospective, of course, as the binding owners may have other ideas, but I can update all to `child` for now

PR #720, opened 2022-10-25
Add ALWAYS and NEVER macros (from SQLite)

isaacbrodsky:
Adds ALWAYS, NEVER, and related macros from SQLite (as linked in the code comments) to help ensure fuzzers or unit tests that reach expected-unreachable code appropriately assert. This also excludes defensive code blocks from coverage metrics.


CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ else()
     set(ENABLE_REQUIRES_ALL_SYMBOLS OFF)
 endif()
 
-option(ENABLE_COVERAGE "Enable compiling tests with coverage." ON)
+option(ENABLE_COVERAGE "Enable compiling tests with coverage." OFF)
```
nrabinowitz, line 48: For my info, why toggle this?

Oh, I see - this now changes the production build as well? So locally, we should make sure to toggle this ON when developing, right?

src/h3lib/include/alloc.h
```diff
@@ -33,10 +33,10 @@
...
 #endif
 
-void* H3_MEMORY(malloc)(size_t size);
-void* H3_MEMORY(calloc)(size_t num, size_t size);
-void* H3_MEMORY(realloc)(void* ptr, size_t size);
-void H3_MEMORY(free)(void* ptr);
+void *H3_MEMORY(malloc)(size_t size);
+void *H3_MEMORY(calloc)(size_t num, size_t size);
+void *H3_MEMORY(realloc)(void *ptr, size_t size);
+void H3_MEMORY(free)(void *ptr);
```
nrabinowitz, line 39: Why wasn't this previously formatted via clang?

src/h3lib/include/h3Assert.h
```diff
@@ -0,0 +1,125 @@
...
+ * Adapted from https://www.sqlite.org/testing.html and
+ * https://www.sqlite.org/assert.html
+ *
+ * Used under the terms of the SQLite3 project, reproduced below:
+ * The author disclaims copyright to this source code.  In place of
+ * a legal notice, here is a blessing:
+ *
+ *    May you do good and not evil.
+ *    May you find forgiveness for yourself and forgive others.
+ *    May you share freely, never taking more than you give.
```
nrabinowitz, line 38: This is amazing 😮

src/h3lib/include/h3Assert.h
```diff
@@ -0,0 +1,125 @@
...
+#else
+#define DEFENSEONLY(X)
+#endif
+
+/*
+** The ALWAYS and NEVER macros surround boolean expressions which
+** are intended to always be true or false, respectively.  Such
+** expressions could be omitted from the code completely.  But they
+** are included in a few cases in order to enhance the resilience
+** of SQLite to unexpected behavior - to make the code "self-healing"
```
nrabinowitz, line 104:
```suggestion
** of the H3 library to unexpected behavior - to make the code "self-healing"
```
?

src/h3lib/include/h3Assert.h
```diff
@@ -0,0 +1,125 @@
...
+** When doing coverage testing ALWAYS and NEVER are hard-coded to
+** be true and false so that the unreachable code they specify will
+** not be counted as untested code.
+*/
+#if defined(H3_OMIT_AUXILIARY_SAFETY_CHECKS)
+#define ALWAYS(X) (1)
+#define NEVER(X) (0)
+#elif !defined(NDEBUG)
+#define ALWAYS(X) ((X) ? 1 : (assert(0), 0))
+#define NEVER(X) ((X) ? (assert(0), 1) : 0)
```
nrabinowitz, line 119: So I understand, the expression `(assert(0), 1)` means: first assert 0 (i.e. fail), then return 1, which is unused in practice but necessary for the compiler?

src/h3lib/lib/algos.c
```diff
@@ -602,10 +603,11 @@ H3Error H3_EXPORT(gridDiskDistancesUnsafe)(H3Index origin, int k, H3Index *out,
             // the end of this ring.
             H3Error neighborResult = h3NeighborRotations(
                 origin, NEXT_RING_DIRECTION, &rotations, &origin);
-            if (neighborResult) {  // LCOV_EXCL_BR_LINE
+            if (neighborResult) {
                 // Should not be possible because `origin` would have to be a
                 // pentagon
-                return neighborResult;  // LCOV_EXCL_LINE
+                // TODO: Reachable via fuzzer
```
nrabinowitz, line 609: For my info: Were these TODOs uncovered by trying to use the new macros here and triggering assertion errors?

src/h3lib/lib/algos.c
```diff
@@ -364,8 +365,7 @@ H3Error h3NeighborRotations(H3Index origin, Direction dir, int *rotations,
 
     int newRotations = 0;
     int oldBaseCell = H3_GET_BASE_CELL(current);
-    if (oldBaseCell < 0 ||  // LCOV_EXCL_BR_LINE
-        oldBaseCell >= NUM_BASE_CELLS) {
+    if (NEVER(oldBaseCell < 0) || oldBaseCell >= NUM_BASE_CELLS) {
```
nrabinowitz, line 368: For my info: This branch is now considered covered in our coverage metrics?

src/h3lib/lib/vertex.c
```diff
@@ -68,13 +69,17 @@ static H3Error vertexRotations(H3Index cell, int *out) {
...
-        for (int p = 0; p < NUM_PENTAGONS; p++) {  // LCOV_EXCL_BR_LINE
+        // We never hit the end condition
+        int p = 0;
+        for (; p < NUM_PENTAGONS; p++) {
             if (pentagonDirectionFaces[p].baseCell == baseCell) {
                 dirFaces = pentagonDirectionFaces[p];
                 break;
             }
         }
+        if (p == NUM_PENTAGONS) {
```
nrabinowitz, line 80: Should this be `NEVER(p == NUM_PENTAGONS)`? This case is supposedly unreachable, right?

website/docs/core-library/testing.md
```diff
@@ -5,8 +5,7 @@ sidebar_label: Testing strategy
 slug: /core-library/testing
 ---
 
-The H3 developers try to ensure the robustness and reliability of the H3 library. Tools used to do this include unit tests with coverage reporting
-and fuzzing.
+The H3 developers try to ensure the robustness and reliability of the H3 library. Tools used to do this include defensive code, unit tests with coverage reporting and fuzzing.
```
nrabinowitz, line 8:
```suggestion
The H3 developers try to ensure the robustness and reliability of the H3 library. Tools used to do this include defensive code, unit tests with coverage reporting, and fuzzing.
```

nrabinowitz (approved):
This looks great, I find the macros very readable and clear. Just to confirm:

- Using the `NEVER` or `ALWAYS` macros on a branch condition will properly exclude that branch from coverage?
- The drop in coverage in this PR is due to the macros exposing reachable branches we had previously excluded from coverage?

CMakeLists.txt
isaacbrodsky: I believe this should be OFF while developing in order to get the benefit of the assertions.

src/h3lib/include/alloc.h
isaacbrodsky: It was not included in the list of source files, presumably an oversight.

src/h3lib/include/h3Assert.h
isaacbrodsky: Conceptually that is correct.

src/h3lib/lib/algos.c
isaacbrodsky: Yes, because for coverage it just sees `0 || oldBaseCell >= NUM_BASE_CELLS` (constant folds to `oldBaseCell >= NUM_BASE_CELLS`, which is covered).

src/h3lib/lib/algos.c
isaacbrodsky: Yes, I generally tried to replace LCOV exclusions first with the macros and had to replace them with TODOs, since this PR makes it clear they are reachable.

src/h3lib/lib/vertex.c
isaacbrodsky: I'm not sure that would satisfy the compile error about dirFaces being uninitialized. We could check.

isaacbrodsky:
> This looks great, I find the macros very readable and clear. Just to confirm:
> 
>     * Using the `NEVER` or `ALWAYS` macros on a branch condition will properly exclude that branch from coverage?
Yes, you can find an example here: https://coveralls.io/builds/53603547/source?filename=src%2Fh3lib%2Flib%2FdirectedEdge.c#L282
>     * The drop in coverage in this PR is due to the macros exposing reachable branches we had previously excluded from coverage?
Yes, I had hoped this PR would increase coverage by being able to exclude some more branches we had not marked with exclusions, but rather it exposed they are reachable, and we should cover them via unit testing.

website/docs/core-library/compilation-options.md
```diff
@@ -56,7 +56,7 @@ Should be set to `Release` for production builds, and `Debug` in development.
 
 ## ENABLE_COVERAGE
 
-Whether to compile `Debug` builds with coverage instrumentation (compatible with GCC, Clang, and lcov).
+Whether to compile `Debug` builds with coverage instrumentation (compatible with GCC, Clang, and lcov). This also elides defensive code in the library.
```
ajfriend, line 59: I had to look up "elides". Maybe "bypasses"? 😅 

And maybe even a quick note on why:

"This also bypasses defensive code in the library, since it should typically not be reachable by coverage."

website/docs/core-library/testing.md
```diff
@@ -18,17 +17,27 @@ Github Actions are used to run the H3 test suite for every commit. A variety of
...
+| Mac OS           | Apple Clang | Debug, Release | x64                    |
+| Windows          | MSVC        | Debug, Release | x64                    | Static library
+| Windows          | MSVC        | Debug, Release | x86                    | Static library
+| Windows          | MSVC        | Debug, Release | x64                    | Dynamic library; testing is not run in this configuration
+| Windows          | MSVC        | Debug, Release | x86                    | Dynamic library; testing is not run in this configuration
+
+## Defensive code
+
+H3 uses preprocessor macros borrowed from [SQLite's testing methodology](https://www.sqlite.org/testing.html) to include defensive code in the library. Defensive code is code that handles error conditions for which there are no known test cases to demonstrate it. The lack of known test cases means that without the macros, the defensive cases could inappropriately reduce coverage metrics, disincentivizing including them.
+
```
ajfriend, line 35: nit/suggestion: What about presenting the next three paragraphs as an unordered list, introduced by something like

```
The macros will behave differently, depending on the build type being "release", "debug", or "coverage":
```

It is probably a matter of personal choice, but I find it can make documentation a little more scannable.

ajfriend: One styling idea:

The macros will behave differently, depending on the library build type being "release", "debug", or "coverage":

- **release builds**: The defensive code is included without modification. These branches are intended to be very simple (usually only `return`ing an error code and possibly `free`ing some resources) and to be visually inspectable.
- **debug builds**: The defensive code is included and `assert` calls are included if the defensive code is invoked. Any unit test or fuzzer which can demonstrate the defensive code is actually reached will trigger a test failure and the developers can be alerted to cover the defensive code in unit tests.
- **coverage builds**: The defensive code is not included by replacing its condition with a constant value. The compiler removes the defensive code and it is not counted in coverage metrics.

ajfriend:
These macros are awesome!

ajfriend (approved):

ajfriend:
@isaacbrodsky, this is maybe more of a conceptual question, but what if there were a user who was risk-tolerant but very interested in performance. Would it make sense for them to want to build a release version of the library that bypasses the defensive code (like we do in the coverage build)? Is that configuration possible, or should we allow for that?

Would that provide any potential performance benefit by skipping the defensive checks?

isaacbrodsky:
> @isaacbrodsky, this is maybe more of a conceptual question, but what if there were a user who was risk-tolerant but very interested in performance. Would it make sense for them to want to build a release version of the library that bypasses the defensive code (like we do in the coverage build)? Is that configuration possible, or should we allow for that?
> 
> Would that provide any potential performance benefit by skipping the defensive checks?

I don't think that configuration is supported here. It would be possible to eliminate some conditional checks, but I would imagine the performance gain to be modest. We could add another CMake option that controls whether those are included or not that's separate from (probably defaulted from) the coverage option.

ajfriend:
> I don't think that configuration is supported here. It would be possible to eliminate some conditional checks, but I would imagine the performance gain to be modest. We could add another CMake option that controls whether those are included or not that's separate from (probably defaulted from) the coverage option.

Makes sense. And I don't think we need to make any changes here. That was just a conceptual question.

PR #733, opened 2022-12-05
Fix possible signed integer overflow in ijToIjk

isaacbrodsky:
Follow up to #706.

PR #735, opened 2022-12-06
Fix upAp7 signed integer overflow

isaacbrodsky:
Adds versions of _upAp7/_upAp7r that check for signed integer overflow on integer operations, for use in the LocalIJ functions. Replaces #733.


src/h3lib/lib/coordijk.c
```diff
@@ -272,6 +307,92 @@ Direction _unitIjkToDigit(const CoordIJK *ijk) {
     return digit;
 }
 
+/**
+ * Returns non-zero if _upAp7 with the given input could have a signed integer
+ * overflow.
+ *
+ * Assumes ijk is IJK+ coordinates (no negative numbers).
+ */
+H3Error _upAp7Checked(CoordIJK *ijk) {
```
isaacbrodsky, line 316: Note that these functions are missing fully branch coverage

src/h3lib/include/mathExtensions.h
```diff
@@ -27,6 +27,14 @@
  */
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+// TODO: Verify
+#define SUM_INT32S_OVERFLOWS(a, b) \
+    ((a) > 0 ? (INT32_MAX - (a) < (b)) : (INT32_MIN - (a) > (b)))
+
+// TODO: Verify
+#define SUB_INT32S_OVERFLOWS(a, b) \
+    ((a) > 0 ? (INT32_MIN + (a) > (b)) : (INT32_MAX + (a) < (b)))
```
isaacbrodsky, line 36: Need to add some tests to demonstrate these macros work as expected.

src/apps/fuzzers/fuzzerInternalCoordIjk.c
```diff
@@ -0,0 +1,52 @@
...
+    if (args->ijk.i >= 0 && args->ijk.j >= 0 && args->ijk.k >= 0) {
+        CoordIJK ijkCopy1 = args->ijk;
+        _upAp7Checked(&ijkCopy1);
+        CoordIJK ijkCopy2 = args->ijk;
+        _upAp7rChecked(&ijkCopy2);
+    }
+    // This function needs a guard check to be safe and that guard
+    // check assumes k = 0.
+    CoordIJK ijkCopy3 = args->ijk;
+    if (ijkCopy3.k == 0 && !_ijkNormalizeCouldOverflow(&ijkCopy3)) {
```
nrabinowitz, line 45: Would it make more sense to actually set `ijkCopy3.k = 0`, rather than discarding all random input where this was not the case?

nrabinowitz (approved):

nrabinowitz:
Is there a strong reason to keep the unchecked versions of these functions? Is there a significant perf difference?

src/h3lib/include/mathExtensions.h
```diff
@@ -27,6 +27,14 @@
  */
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+// TODO: Verify
```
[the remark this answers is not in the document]
isaacbrodsky: This comment seems to be on an outdated version of the file. That comment was removed when I added tests for these macros.

isaacbrodsky:
> Is there a strong reason to keep the unchecked versions of these functions? Is there a significant perf difference?

I didn't evaluate the performance difference. I don't believe there is a strong reason to keep the unchecked versions of the functions, although in use in cell indexing it should not be possible to overflow them, so they would have a bunch of `if (NEVER(_upAp7(...)))` guards.

src/h3lib/lib/coordijk.c
```diff
@@ -207,6 +208,41 @@ void _ijkScale(CoordIJK *c, int factor) {
...
+        // Only if the min is less than 0 will the resulting number be larger
+        // than max. If min is positive, then max is also positive, and a
+        // positive signed integer minus another positive signed integer will
+        // not overflow.
+        if (SUM_INT32S_OVERFLOWS(max, min)) {
+            // max + min would overflow
+            return true;
+        }
+        if (SUB_INT32S_OVERFLOWS(0, min)) {
+            // 0 - INT32_MIN would overflow
```
dfellis, line 235: This comment should be:

```suggestion
            // 0 - min would overflow
```

Right?

src/h3lib/include/mathExtensions.h
```diff
@@ -27,6 +27,14 @@
  */
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+/** Evaluates to true if a + b would overflow for int32 */
+#define SUM_INT32S_OVERFLOWS(a, b) \
```
dfellis, line 31: Nit: `SUM` and `SUB` look nearly identical so I didn't even notice they were different macros in the conditionals below at first. Maybe this one can be:

```suggestion
#define ADD_INT32S_OVERFLOWS(a, b) \
```

src/h3lib/lib/coordijk.c
```diff
@@ -272,6 +308,96 @@ Direction _unitIjkToDigit(const CoordIJK *ijk) {
...
+    int j = ijk->j - ijk->k;
+
+    if (SUM_INT32S_OVERFLOWS(i, i)) {
+        return E_FAILED;
+    }
+    int i2 = i + i;
+    if (SUM_INT32S_OVERFLOWS(i2, i)) {
+        return E_FAILED;
+    }
+    int i3 = i2 + i;
```
dfellis, line 329: Maybe it would make sense to make a `MUL_INT32S_OVERFLOWS(i, 3)` check and skip this intermediate addition and check?

src/h3lib/lib/coordijk.c
```diff
@@ -272,6 +308,96 @@ Direction _unitIjkToDigit(const CoordIJK *ijk) {
...
+    }
+    int i2 = i + i;
+    if (SUM_INT32S_OVERFLOWS(j, j)) {
+        return E_FAILED;
+    }
+    int j2 = j + j;
+    if (SUM_INT32S_OVERFLOWS(j2, j)) {
+        return E_FAILED;
+    }
+    int j3 = j2 + j;
```
dfellis, line 378: A `MUL_INT32S_OVERFLOWS` check would work here, too.

src/h3lib/lib/coordijk.c
isaacbrodsky: In this case, `min` would be `INT32_MIN`, which is I believe the only case that would hit this condition.

src/h3lib/include/mathExtensions.h
isaacbrodsky: That's reasonable. Renamed to ADD_...

src/h3lib/lib/coordijk.c
isaacbrodsky: Is there a reference for a `MUL_INT32S_OVERFLOWS` we can use? I looked at the implementation you pointed to in a conversation last week but that involved floating point math, and I was skeptical there would be a significant difference with the intermediate addition+check approach.

dfellis (approved):

Issue #666, opened 2022-08-31
Version 4.x cell stats documentation: edge length missing


holgerteichgraeber:
The cell stats table is currently missing edge length, the section is marked by a TODO: https://h3geo.org/docs/core-library/restable 

Would you mind updating this page in the documentation with edge lengths? It must have disappeared with v4, and it used to be extremely helpful in quickly looking up and comparing different resolutions and making modeling decisions.

ajfriend:
IIRC, those original edge length values may have actually been incorrect. We planned to update them, but never got around to it. Also, it wasn't clear that there was an efficient way to compute the exact average without enumerating all of them (unlike the cell area case), so the proposal was to compute exact values for a few resolutions and then extrapolate. Something like this: https://gist.github.com/ajfriend/200299f857ed22bf34bde4dfbc562100

I can maybe take a look at some point, but I think we'd also be more than happy if someone took this on and put up a PR with a source notebook :)

PR #749, opened 2023-01-31
Update average edge lengths

### LINKED ISSUE #526: Average Hexagon Length value incorrect in restable
It seems that the column Average Hexagon Edge Length (km) in the [restable](https://h3geo.org/docs/core-library/restable/) is undershooting the actual length of the hexagon edges. 

The Average Hexagon Area (km2) column seems correct when calculating in QGIS (with correct projection), and also measuring the edge in QGIS gives a different number. 

Moreover, the edge length does not correspond if we calculate a hexagon based on it's formula. 

For example, at res7 the edge length should be ~1.5 km, not 1.22 km. 

image

mciethan:
I recomputed average edge lengths for all H3 resolutions, building off of the proposed methodology and example notebook provided by @ajfriend in #666 (which involves calculating exact averages for resolutions 0 through 6 and then extrapolating the rest). This would help fix #526 and other issues related to average edge lengths.

Source notebook: https://gist.github.com/mciethan/3e10802c1f41972831c325994d97ef27

My initial commit to this PR only modifies the restable docs.  I haven't attempted to modify the average edge length function itself, although it would make sense to do that before updating the docs. I assume that would just be a matter of updating the hard-coded values, but I am a novice of H3 and also don't know any C programming, so I don't have a good understanding of exactly which files would need to be updated, how to test the changes, etc.


nrabinowitz (approved):

nrabinowitz:
This looks great, thanks for the contribution! I'll let @ajfriend check the approach here, but it LGTM. I believe you're correct that all we would need to do in code is modify the hard-coded values, feel free to add that to this PR.

dfellis (approved):

ajfriend:
Looks good to me! And just as a quick sanity check, I made a plot with the interpolated values:

Screenshot 2023-03-23 at 2 20 06 PM

ajfriend (approved):

PR #762, opened 2023-03-11
Reorganize tests into public/internal

isaacbrodsky:
Reorganizes tests into public and internal categories. The intention is that the public part of the tests should be runnable even when the internal library functions are not available (for example, on Windows, or using [HydroniumLabs/h3o](https://github.com/HydroniumLabs/h3o)). This is not possible today in this PR because there are still some internal functions that are used in various parts of tests.

This PR is intended to only reorganize the tests - there should not be any coverage change or meaningful change to the test suite (or library for that matter).

I'd suggest a combination or replacing function calls with constants or data files (since the data file tests are easiest to port to different bindings or implementations), and when it really helps the test suite, marking some functions as exported for tests. These functions would not be part of the public API for versioning purposes and are not intended to be called by applications, but would be exposed by the library. An example would be `setH3Index`.

Also included in this PR is a new CMake option `AUDIT_SOURCE_FILE_LIST`, which is used in CI to ensure the file lists in CMakeLists.txt actually match what's in the repository.


isaacbrodsky:
> [Coverage Status](https://coveralls.io/builds/57754630)
> 
> Coverage: 98.227% (-0.4%) from 98.657% when pulling **[e9532a7](https://github.com/uber/h3/commit/e9532a75f993b13a4bc1467fc57ca25e1ec15e8e) on isaacbrodsky:reorg-internal-tests** into **[f0336b5](https://github.com/uber/h3/commit/f0336b52967034eeec4245aa99b1c2361e97fad9) on uber:master**.

Coverage decreasing probably means a test got dropped somewhere in the refactor.

src/apps/testapps/testVertexInternal.c
```diff
@@ -0,0 +1,88 @@
...
+/** @file
+ * @brief tests H3 vertex functions.
+ */
+
+#include "test.h"
+#include "utility.h"
+#include "vertex.h"
+
+SUITE(Vertex) {
+    TEST(vertexNumForDirection_hex) {
```
isaacbrodsky, line 25: @dfellis this should be where the test you commented on moved to?

src/apps/testapps/testCellToLocalIjInternal.c
```diff
@@ -0,0 +1,118 @@
...
+#include "algos.h"
+#include "baseCells.h"
+#include "constants.h"
+#include "h3Index.h"
+#include "h3api.h"
+#include "localij.h"
+#include "test.h"
+#include "utility.h"
+
+SUITE(h3ToLocalIj) {
```
nrabinowitz, line 36: Nit: Doesn't much matter, but should these have the `-Internal` suffix in the suite name?

nrabinowitz (approved):
LGTM

src/apps/testapps/testCellToLocalIjInternal.c
isaacbrodsky: We can -- although maybe we should just use a macro like `__FILE__` or an extension macro like `__FILE_NAME__` instead.

src/apps/testapps/testVertexInternal.c
dfellis: Got it.

dfellis (approved):

Issue #775, opened 2023-07-20
Add additional modes for polygonToCells


warnes:
**TLDR:**  Allow specification of 'contained', 'intersects', 'covers' for `polygonToCells`.

**Details:**

For my project, I need to be able to convert arbitrary multipolygons to the set of hexes which completely cover the enclosed area.

The current implementation of `polygonToCells` only generates a list of cells whose centroids are located inside of the (multi)polygon.

My current workaround is:
1. expand the original multipolygon by `edge_length(res)` using `st_buffer`
2. use `polygonToCells` to generate the list of cell ids
3. Use `st_overlaps` and `st_contains` to flag all the cells that cover the original multipolygon.

Unfortunately, this is somewhat slow, so I'm looking for a faster way to accomplish this.

(Note that the algorithm described at https://github.com/uber/h3/issues/608#issuecomment-1137852267 won't work if any area of the original polygon is too small/narrow to contain a cell centroid.).

KirkDybvik:
One somewhat crude option (but effective and avoids some extra st_ calls) would be the following:
1. If desired h3 resolution is R, first use R+2 for generating the indexes for "covering" the polygon
2. For each of the cell indexes identified in step 1, find the unique set of indexes at resolution R 
3. Review what you end up with, it might be a bit "extra chunky" and have some cells that barely touch the original shape - but gets around the challenge of losing coverage cells that did not get included due to cell centroid not within the shape
4. Review what you end up with in the case where the original polygon/multipolygon contains holes, you may have some custom considerations on when you want to retain or eliminate holes

"Your mileage may vary" with this approach, but hopefully it helps

nrabinowitz:
> One somewhat crude option (but effective and avoids some extra st_ calls) would be the following:
> 
> 1. If desired h3 resolution is R, first use R+2 for generating the indexes for "covering" the polygon
> 2. For each of the cell indexes identified in step 1, find the unique set of indexes at resolution R
> 3. Review what you end up with, it might be a bit "extra chunky" and have some cells that barely touch the original shape - but gets around the challenge of losing coverage cells that did not get included due to cell centroid not within the shape
> 4. Review what you end up with in the case where the original polygon/multipolygon contains holes, you may have some custom considerations on when you want to retain or eliminate holes
> 
> "Your mileage may vary" with this approach, but hopefully it helps

This approach would work, but caveat emptor: 
- You could still end up with missed cells that intersect the polygon
[...]

Issue #521, opened 2021-10-05
polygonToCells fails for some simple large polygons


mutability:
(Originally: uber/h3-py#202)

This polygon (nb: in degrees, not radians) causes `polygonToCells` to return `E_FAILED` (at resolution 5, probably true for other resolutions):

```
{
  {-74, -120},
  {-74, 0},
  {-74, 120},
  {-84, 120},
  {-84, 0},
  {-84, -120}
}
```

I poked at this a bit and the problem AIUI is:

* `polygonToCells` ends up generating more than the "maximum" number of cells; because
* `maxPolygonToCellsSize` underestimates the number of cells needed for this polygon; because
* `bboxHexEstimate` underestimates the number of cells needed

For this poly `bboxHexEstimate` returns 16037 cells at resolution 5. By messing with `POLYGON_TO_CELLS_BUFFER` I got `polygonToCells` to run to completion, and it generated 20337 cells.

My intuitive guess at what's going wrong is that `bboxHexEstimate` naively measures the distance between the two corners of the bounding box using `distanceKm`. But this always returns the _shorter_ great-circle arc, which for large bounding boxes is wrong (for this polygon, the shorter arc is _not_ the diagonal of the bounding box)

mutability:
Test case showing the problem:

```
#include <h3api.h>
#include <inttypes.h>
#include <stdio.h>
#include <math.h>
#include <assert.h>

static LatLng goodVerts[] = { {-74, -120},
                             {-74, 0},
                             {-74, 120},
                             {-82, 120},
                             {-82, 0},
                             {-82, -120} };
static GeoLoop goodLoop = { .numVerts = 6, .verts = goodVerts };
static GeoPolygon goodPoly;

static LatLng badVerts[] = { {-74, -120},
                             {-74, 0},
                             {-74, 120},
                             {-84, 120},
                             {-84, 0},
                             {-84, -120} };
static GeoLoop badLoop = { .numVerts = 6, .verts = badVerts };
static GeoPolygon badPoly;

static void initPoly(GeoPolygon *poly, GeoLoop *loop)
{
    for (int i = 0; i < loop->numVerts; ++i) {
        loop->verts[i].lat = loop->verts[i].lat * M_PI / 180.0;
        loop->verts[i].lng = loop->verts[i].lng * M_PI / 180.0;
    }
    poly->numHoles = 0;
    poly->geoloop = *loop;
}

static void fill(const char *what, GeoPolygon *poly)
{
    H3Error error;

[...]

Issue #275, opened 2019-08-21
Support different polyfill mode 


boyxgc:
Currently, the polyfill function only includes hexagons whose center is within the polygon. Would be good to support different polyfill needs:
1. center within the input polygon
2. input polygon fully contained
3. only hexagons fully contained in the input polygon

bkowshik-rapido:
> input polygon fully contained

+1 this issue. Wanted to do something like ^ today. But, later realized about the below line from the documentation. Looked around for any related issue and found this one to be appropriate.

Ref: https://github.com/uber/h3-js#module_h3.polyfill
> Get all hexagons with centers contained in a given polygon.

<img width="979" alt="Screenshot 2019-12-26 at 11 26 10 AM" src="https://user-images.githubusercontent.com/58680555/71465987-2c734500-27e4-11ea-8c67-dbf0a71eb758.png">

bkowshik-rapido:
Had an interesting case at work today:

- There are two clusters, colored green and blue.
- Now, the task was to cover both these clusters with `hex8` polygons
- It so happened that there was this `hex8` who's center was in the space between the two clusters 😬 

<img width="1680" alt="Screenshot 2020-02-10 at 9 18 26 AM" src="https://user-images.githubusercontent.com/58680555/74119651-5cd14000-4be6-11ea-896a-d2d4c97301dd.png">

dfellis:
Well, those two polygons do overlap. Polygons meant to be adjacent to each other should be using the exact same points on the shared sides, that's the only way the promises of this current polyfill algorithm will be kept.

kevinschaich:
For option (2):

> 2. input polygon fully contained

I've put together some code to do this (introduces dependency to Shapely):

```python
import json
import h3
from shapely import geometry
from shapely.geometry import Point, MultiPoint, LineString, MultiLineString, Polygon, MultiPolygon
from pyspark.sql import functions as F


def _point_to_h3(point: Point, resolution: int):
    '''
    Return the set of H3 cells at the specified resolution which completely cover the input point.
    '''
    result_set = set()

    # Hexes for point
    result_set.update(h3.geo_to_h3(t[1], t[0], resolution) for t in list(point.coords))
    return result_set


def _line_to_h3(line: LineString, resolution: int):
    '''
    Return the set of H3 cells at the specified resolution which completely cover the input line.
    '''
    result_set = set()
    # Hexes for endpoints
[...]

PR #785, opened 2023-10-03
Alternate polygonToCells algorithm

nrabinowitz:
This PR implements an alternate algorithm for `polygonToCells`, which provides `polygonToCompactCells` as a byproduct.

### Goals
- [x] Offer a stream/iterator-based algorithm to avoid memory issues when converting polygons to resolutions with high-cardinality output (#521, https://github.com/uber/h3-java/issues/21)
- [x] Provide compact output without a separate `compactCells` step
- [x] Improve performance of `polygonToCells`
- [ ] Add new polyfill modes for `contains` and `intersects` (#775, #275, https://github.com/uber/h3-java/issues/41)

### Algorithm Overview
```
let cell = base cell 0
while cell:
  res = getResolution(cell)
  if res == target res:
     if cell is in polygon:
        set output to cell and return
  if res < target res:
     let bbox = boundingBox(cell) // where bbox is padded to contain all descendants
     if bbox intersects polygon:
        if bbox is contained by polygon:
          set output to cell and return
        cell = first child of cell
        continue
  cell = next cell // where "next" is the next sibling, or next sibling of parent, etc
when cell is null we're done
```

### Results
 -  This PR includes an iterator-based algo with full test parity to the current algorithm. In its iterator form, it takes very little memory and that memory does not scale with the size of the output - so we can stream cells out of a very large polyfill process without worrying about memory issues.
- The new algo gives compact (or nearly compact) output by default, so we get `polygonToCellsCompact` effectively for free.
- The performance isn't exactly better or worse than the current algorithm - it just has different characteristics. In benchmarks, it's faster than the current algorithm for large numbers of cells (because we can use a small number of compact parent cells) but slower for large numbers of polygon vertices (because the containment check scales with the number of vertices). See benchmarks below.
- The new polyfill modes are not included in this PR, but should be simple to add based on the work done here (particularly `cellBoundaryInsidePolygon` and `cellBoundaryCrossesGeoLoop`).

### For Discussion
- Are the benefits sufficient to replace the existing `polygonToCells` algo, even if performance degrades in some cases?
- In order to get the memory benefits, consumers need to use the iterator directly, or we need to implement some sort of batch mode that can fill chunks of memory provided by the caller (might be worthwhile to avoid costs in bindings associated with an FFI per output cell). Which option is better? What if any clean up do we need in the iterator interface if we want to make it part of the public API? Should we make the `cellToChildren` iterator public as well?

### Benchmarks

In the following benchmakrks, `SF` is a 6-vertex polygon covering San Francisco, `Alameda` is a 49-vertex polygon covering a similar area, and `SouthernExpansion` is a 22-vertex polygon covering an area 2x-3x larger than SF. 
- The old `polygonToCells` performs better when the ratio of output cells to vertexes is low (few cells, complex polygons)
- The new `polygonToCells2` performs better when the ratio of output cells to vertexes is high (many cells, simple polygons)
- Weirdly, the `polygonToCellsCompact` function, which is just `polygonToCells2` without the uncompact step, performs about the same as `polygonToCells2` 🤔. I went over this several times, but it seems like the benchmark is correct? 

**Res 9**
```
Built target benchmarkPolygon
	-- polygonToCellsSF: 2264.810000 microseconds per iteration (500 iterations)
	-- polygonToCellsSF2: 3530.106000 microseconds per iteration (500 iterations)
	-- polygonToCellsSFCompact: 3530.478000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlameda: 3182.120000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlameda2: 5960.740000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlamedaCompact: 6027.128000 microseconds per iteration (500 iterations)
	-- polygonToCellsSouthernExpansion: 103521.500000 microseconds per iteration (10 iterations)
	-- polygonToCellsSouthernExpansion2: 112462.900000 microseconds per iteration (10 iterations)
	-- polygonToCellsSouthernExpansionCompact: 104374.600000 microseconds per iteration (10 iterations)
```
**Res 10**
```
Built target benchmarkPolygon
	-- polygonToCellsSF: 12268.826000 microseconds per iteration (500 iterations)
	-- polygonToCellsSF2: 11435.998000 microseconds per iteration (500 iterations)
	-- polygonToCellsSFCompact: 11363.694000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlameda: 15037.306000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlameda2: 25583.798000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlamedaCompact: 25478.888000 microseconds per iteration (500 iterations)
	-- polygonToCellsSouthernExpansion: 693673.900000 microseconds per iteration (10 iterations)
	-- polygonToCellsSouthernExpansion2: 672981.900000 microseconds per iteration (10 iterations)
	-- polygonToCellsSouthernExpansionCompact: 659994.200000 microseconds per iteration (10 iterations)
```
**Res 11**
```
Built target benchmarkPolygon
	-- polygonToCellsSF: 81978.584000 microseconds per iteration (500 iterations)
	-- polygonToCellsSF2: 50297.930000 microseconds per iteration (500 iterations)
	-- polygonToCellsSFCompact: 50223.350000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlameda: 96910.940000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlameda2: 140856.702000 microseconds per iteration (500 iterations)
	-- polygonToCellsAlamedaCompact: 140873.162000 microseconds per iteration (500 iterations)
	-- polygonToCellsSouthernExpansion: 5245777.500000 microseconds per iteration (10 iterations)
	-- polygonToCellsSouthernExpansion2: 4613408.400000 microseconds per iteration (10 iterations)
	-- polygonToCellsSouthernExpansionCompact: 4607957.500000 microseconds per iteration (10 iterations)
```


src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,422 @@
...
+/**
+ * Parity implementation for polygonToCells
+ */
+H3Error H3_EXPORT(polygonToCells2)(const GeoPolygon *polygon, int res,
+                                   uint32_t flags, H3Index *out) {
+    IterCellsPolygon iter = iterInitPolygon(polygon, res, flags);
+    int64_t i = 0;
+    for (; iter.cell; iterStepPolygon(&iter)) {
+        out[i++] = iter.cell;
+    }
```
isaacbrodsky, line 407: you could also assert we don't go over the estimation size here (which would be an out of bounds write)

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,422 @@
...
+
+    while (cell) {
+        int cellRes = H3_GET_RESOLUTION(cell);
+
+        // Target res: Do a fine-grained check
+        if (cellRes == iter->_res) {
+            // Check if the cell is in the polygon
+            // TODO: Handle other polyfill modes here
+            LatLng center;
+            H3_EXPORT(cellToLatLng)(cell, &center);
```
isaacbrodsky, line 251: this should have a NEVER check for the error

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    // Adjust the BBox to handle poles
+    H3Index poleTest;
+    // North pole
+    err = H3_EXPORT(latLngToCell)(&NORTH_POLE, H3_GET_RESOLUTION(cell),
+                                  &poleTest);
+    if (NEVER(err != E_SUCCESS)) {
+        return err;
+    }
```
isaacbrodsky, line 74: It looks like Clang warnings want these stores to be to different variables https://github.com/uber/h3/actions/runs/6408852991/job/17398764952?pr=785 rather than a reused `err` variable. Not 100% sure I agree (since this is really complaining about something that only affects the coverage build) although I think it is marginally clearer than reusing `err`.

src/h3lib/lib/polygon.c
```diff
@@ -83,3 +83,103 @@ bool pointInsidePolygon(const GeoPolygon *geoPolygon, const BBox *bboxes,
...
+ * @param  b2 End of line B
+ * @return    Whether the lines intersect
+ */
+bool lineIntersectsLine(const LatLng *a1, const LatLng *a2, const LatLng *b1,
+                        const LatLng *b2) {
+    double test;
+    test = ((b2->lat - b1->lat) * (a1->lng - b1->lng) -
+            (b2->lng - b1->lng) * (a1->lat - b1->lat)) /
+           ((b2->lng - b1->lng) * (a2->lat - a1->lat) -
+            (b2->lat - b1->lat) * (a2->lng - a1->lng));
```
isaacbrodsky, line 177: In CI, testPolygonInternal hits an unguarded division by zero on this line:
```
TEST polygonInternal
/Users/isaac/oss/h3/src/h3lib/lib/polygon.c:175:56: runtime error: division by zero
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/isaac/oss/h3/src/h3lib/lib/polygon.c:175:56 in
......................................................................................zsh: abort      ./bin/testPolygonInternal
```

(to configure: `cmake -DCMAKE_C_FLAGS="-fsanitize=undefined,float-divide-by-zero -fno-sanitize-recover=undefined,float-divide-by-zero" ..`)

src/h3lib/lib/polygon.c
```diff
@@ -83,3 +83,103 @@ bool pointInsidePolygon(const GeoPolygon *geoPolygon, const BBox *bboxes,
...
+    if (test < 0 || test > 1) return false;
+
+    test = ((a2->lat - a1->lat) * (a1->lng - b1->lng) -
+            (a2->lng - a1->lng) * (a1->lat - b1->lat)) /
+           ((b2->lng - b1->lng) * (a2->lat - a1->lat) -
+            (b2->lat - b1->lat) * (a2->lng - a1->lng));
```
isaacbrodsky, line 183: same as above, presumably this could result in division by zero

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,422 @@
...
+                cell = child;
+                continue;
+            }
+        }
+
+        // Find the next cell in the sequence of all cells and continue
+        cell = nextCell(cell);
+    }
+    // If we make it out of the loop, we're done
+    iterDestroyPolygonCompact(iter);
```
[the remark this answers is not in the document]
dfellis: That seems like a reasonable way to think about it. Perhaps instead of calling it a `destroyIterator` or etc, it's `cancelIterator`?

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,422 @@
...
+
+    CellBoundary boundary;
+    err = H3_EXPORT(cellToBoundary)(cell, &boundary);
+    if (err) {
+        return err;
+    }
+    // Convert to GeoLoop
+    GeoLoop loop;
+    loop.numVerts = boundary.numVerts;
+    loop.verts = (LatLng *)&boundary.verts;
```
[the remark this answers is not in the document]
nrabinowitz: Yes - otherwise the compiler complains about passing a fixed-size array (`*LatLng[10]` vs `*LatLng`)

src/h3lib/lib/polyfill.c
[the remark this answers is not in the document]
nrabinowitz: 🤷 I like `destroy`, as it clearly indicates "release memory" - we use it for `destroyLinkedMultiPolygon` as well. I went back and forth about doing this for the caller, but prefer this option:
- In general, the component allocating memory should be responsible for releasing it, and in this case that's the iterator functions
- This is more ergonomic for the most common use cases, and ensures proper release in those cases, reducing the potential for bugs

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,422 @@
...
+/**
+ * Destroy an iterator, releasing any allocated memory. Iterators destroyed in
+ * this manner are safe to use but will always return H3_NULL.
+ * @param  iter Iterator to destroy
+ */
+void iterDestroyPolygon(IterCellsPolygon *iter) {
+    iterDestroyPolygonCompact(&(iter->_cellIter));
+    iter->cell = H3_NULL;
+    iter->error = E_SUCCESS;
+}
```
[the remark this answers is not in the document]
nrabinowitz: Will do

src/h3lib/lib/polyfill.c
nrabinowitz: Will add

src/h3lib/lib/polyfill.c
nrabinowitz: Ok, will update - I thought I was being clever by declaring this at the top of each function that could throw, but clearly not.

src/h3lib/lib/polyfill.c
nrabinowitz: I don't think we'd want this in production code, since we *assume* the caller has sized the memory correctly, but the `*out` array could be any size, so our assertion would only be useful as an internal check that we were meeting our contract. But I think can use the assertion macros to check this in tests.

src/h3lib/lib/polygon.c
nrabinowitz: Thanks! Will fix

nrabinowitz:
With the various optimizations I've added so far, I've closed the gaps in the original benchmarks, and I think the new algorithm is likely to be faster most of the time, though still slower as the ratio of vertexes to cells increases (numbers shown below are ratio of old / new):

**Res 9**
```
Original:
  Simple: 0.64x
  Complex: 0.53x
  Large: 0.92x

Now:
  Simple: 1.00x
  Complex: 0.81x
  Large: 1.3x
```
**Res 10**
```
Original:
  Simple: 1.07x
  Complex: 0.58x
  Large: 1.03x

Now:
  Simple: 1.56x
  Complex: 0.83x
  Large: 1.36x
```

nrabinowitz:
Some more benchmarking data - when transpiled into `h3-js`, the performance looks almost exactly the same, so close that for a while I was worried I wasn't testing correctly:
```
polygonToCells_all_countries_0 x 3.98 ops/sec ±9.40% (30 runs sampled)
polygonToCells2_all_countries_0 x 3.78 ops/sec ±11.69% (30 runs sampled)
polygonToCells_all_countries_1 x 0.74 ops/sec ±3.42% (21 runs sampled)
polygonToCells2_all_countries_1 x 0.75 ops/sec ±1.33% (21 runs sampled)
polygonToCells_all_countries_2 x 0.53 ops/sec ±2.04% (21 runs sampled)
polygonToCells2_all_countries_2 x 0.55 ops/sec ±1.08% (21 runs sampled)
polygonToCells_all_countries_3 x 0.20 ops/sec ±10.64% (20 runs sampled)
polygonToCells2_all_countries_3 x 0.20 ops/sec ±2.62% (20 runs sampled)
polygonToCells_all_countries_4 x 0.04 ops/sec ±4.17% (20 runs sampled)
polygonToCells2_all_countries_4 x 0.04 ops/sec ±3.42% (20 runs sampled)
polygonToCells_FRANCE_0 x 36.45 ops/sec ±140.62% (28 runs sampled)
polygonToCells2_FRANCE_0 x 30.89 ops/sec ±151.18% (30 runs sampled)
polygonToCells_FRANCE_1 x 26.86 ops/sec ±133.33% (74 runs sampled)
polygonToCells2_FRANCE_1 x 60.46 ops/sec ±61.84% (73 runs sampled)
polygonToCells_FRANCE_2 x 38.19 ops/sec ±78.95% (55 runs sampled)
polygonToCells2_FRANCE_2 x 73.25 ops/sec ±0.60% (56 runs sampled)
polygonToCells_FRANCE_3 x 21.73 ops/sec ±112.03% (43 runs sampled)
polygonToCells2_FRANCE_3 x 27.79 ops/sec ±82.78% (51 runs sampled)
polygonToCells_FRANCE_4 x 21.06 ops/sec ±1.18% (54 runs sampled)
polygonToCells2_FRANCE_4 x 22.21 ops/sec ±0.77% (56 runs sampled)
```
The `all_countries` suite uses multipolygons for the 258 countries in the [Natural Earth 1:110m vector data](https://www.naturalearthdata.com/downloads/110m-cultural-vectors/110m-admin-0-countries/), which I figured should offer a wide range of shapes and sizes. I wasn't able to benchmark past res 4, but performance looks almost identical for this dataset. Looking at an individual country like France highlights some differences (in this case, the new version performs slightly better), but overall it's a wash.

Given the memory benefits, I'm considering this a win, though I was hoping it would actually be faster 🤷.

nrabinowitz:
Even more benchmarking here (now in C) suggests that the resolution of the output cells also makes a difference, which makes sense given the hierarchical search involved in the new algo. This means that the performance of the new algo decreases compared to the old algo as the res gets finer. The following benchmark is for Djibouti:
```
res 5
  -- polygonToCells_1: 15230.800000 microseconds per iteration (5 iterations)
  -- polygonToCells_2: 14093.200000 microseconds per iteration (5 iterations)
res 6
  -- polygonToCells_1: 46217.600000 microseconds per iteration (5 iterations)
  -- polygonToCells_2: 87146.400000 microseconds per iteration (5 iterations)
res 7
  -- polygonToCells_1: 222787.600000 microseconds per iteration (5 iterations)
  -- polygonToCells_2: 552836.200000 microseconds per iteration (5 iterations)
res 8
  -- polygonToCells_1: 1393634.200000 microseconds per iteration (5 iterations)
  -- polygonToCells_2: 3721000.000000 microseconds per iteration (5 iterations)
res 9
  -- polygonToCells_1:  8811737.600000 microseconds per iteration (5 iterations)
  -- polygonToCells_2: 28252274.800000 microseconds per iteration (5 iterations)
```
By res 9 it's 3x worse than the current implementation 😢

nrabinowitz:
Further investigation shows that there's something else going on, rather than just resolution, but I'm not sure what.

The benchmark for all countries has the old algo better after res 5, but I can't run it for finer resolutions:
```
Res 0
	-- polygonToCellsAllCountries_1: 1846956.400000 microseconds per iteration (5 iterations)
	-- polygonToCellsAllCountries_2: 147822.200000 microseconds per iteration (5 iterations)
Res 1
	-- polygonToCellsAllCountries_1: 3200132.800000 microseconds per iteration (5 iterations)
	-- polygonToCellsAllCountries_2: 1224828.200000 microseconds per iteration (5 iterations)
Res 2
	-- polygonToCellsAllCountries_1: 2814667.400000 microseconds per iteration (5 iterations)
	-- polygonToCellsAllCountries_2: 1650428.200000 microseconds per iteration (5 iterations)
Res 3
	-- polygonToCellsAllCountries_1: 5786302.800000 microseconds per iteration (5 iterations)
	-- polygonToCellsAllCountries_2: 3390271.200000 microseconds per iteration (5 iterations)
Res 4
	-- polygonToCellsAllCountries_1: 14079108.200000 microseconds per iteration (5 iterations)
	-- polygonToCellsAllCountries_2: 13361490.400000 microseconds per iteration (5 iterations)
Res 5
	-- polygonToCellsAllCountries_1: 63288210.400000 microseconds per iteration (5 iterations)
	-- polygonToCellsAllCountries_2: 73714663.800000 microseconds per iteration (5 iterations)
```
Testing Indonesia shows the same pattern, old algo is almost twice as fast by res 7. But testing a smaller set of small Indonesian islands at finer resolutions shows the opposite pattern from res 8 onward:
```
Res 8
	-- polygonToCellsIndonesia_1: 7767.200000 microseconds per iteration (5 iterations)
	-- polygonToCellsIndonesia_2: 10843.000000 microseconds per iteration (5 iterations)
Res 9
	-- polygonToCellsIndonesia_1: 22478.400000 microseconds per iteration (5 iterations)
	-- polygonToCellsIndonesia_2: 25132.200000 microseconds per iteration (5 iterations)
Res 10
	-- polygonToCellsIndonesia_1: 109577.000000 microseconds per iteration (5 iterations)
	-- polygonToCellsIndonesia_2: 94577.200000 microseconds per iteration (5 iterations)
Res 11
	-- polygonToCellsIndonesia_1: 754375.400000 microseconds per iteration (5 iterations)
	-- polygonToCellsIndonesia_2: 472660.000000 microseconds per iteration (5 iterations)
Res 12
	-- polygonToCellsIndonesia_1: 4869990.800000 microseconds per iteration (5 iterations)
	-- polygonToCellsIndonesia_2: 2903927.600000 microseconds per iteration (5 iterations)
```
So again, it may be a wash 🤷‍♂️. As far as I can tell, simpler shapes with better compaction yield faster times for the new algo, complex shapes and poor compaction are faster with the old.

isaacbrodsky:
I put together a little animation of this process (code: https://github.com/isaacbrodsky/h3/tree/updated-polyfill-animation), to aid in visualizing what it's doing. I intentionally cropped out the first and last iterations were it's going over the base cells because the scale is so different compared to where most of the work happens.

30fps:
sf_polyfill

15fps:
sf_polyfill_15fps

nrabinowitz:
> I put together a little animation of this process (code: https://github.com/isaacbrodsky/h3/tree/updated-polyfill-animation), to aid in visualizing what it's doing. I intentionally cropped out the first and last iterations were it's going over the base cells because the scale is so different compared to where most of the work happens.

These are great! I was thinking about making an Observable notebook for this, but it's a little tough without reimplementing the algo, or just animating a log of the output

nrabinowitz:
Latest commit speeds up the bbox calculation, gaining maybe 20% perf, but still slower for 75% of countries in the benchmark :(.

nrabinowitz:
Added a lookup table for the res 0 cell bounding boxes, since we check all of them every time, and this seems to have done it! Now in my benchmark (all countries, res 1-5) the new polyfill algo is faster 80% of the time! The main issue here is that the pre-calculated res 0 bboxes seem to wipe out the advantage of the old algorithm for smaller, low-vertex shapes:

image

nrabinowitz:
One additional note here: My new approach to `NORMALIZE_LNG` seems to result in a lot of cases where we lose branch coverage (and possibly no branch coverage is possible, e.g. if we set the `normalization` argument explicitly we'll never hit most branches). Open to better options here - I'm also not thrilled that my current implementation hits the most common case, `NORMALIZE_NONE`, last. In other languages I'd probably pass around a function here, but that's harder and more costly in C. Let me know if you have a better option.

src/h3lib/lib/latLng.c
```diff
@@ -137,6 +137,24 @@ double constrainLng(double lng) {
...
+double normalizeLng(const double lng,
+                    const LongitudeNormalization normalization) {
+    switch (normalization) {
+        case NORMALIZE_NONE:
+            return lng;
+        case NORMALIZE_EAST:
+            return lng < 0 ? lng + (double)M_2PI : lng;
+        case NORMALIZE_WEST:
+            return lng > 0 ? lng - (double)M_2PI : lng;
+    }
```
dfellis, line 155: [GCC is complaining that there's no `default` case here](https://github.com/uber/h3/actions/runs/6658832157/job/18096562714?pr=785#step:6:71) so it's possible to reach the end of the function without returning a number.

Perhaps this could be rewritten as:

```suggestion
    switch (normalization) {
        case NORMALIZE_EAST:
            return lng < 0 ? lng + (double)M_2PI : lng;
        case NORMALIZE_WEST:
            return lng > 0 ? lng - (double)M_2PI : lng;
        default:
            return lng;
    }
```

nrabinowitz: Updated, thanks

dfellis (approved):

src/h3lib/include/polyfill.h
```diff
@@ -0,0 +1,86 @@
...
+    IterCellsChildren _childIter;       // sub-iterator for cell children
+} IterCellsPolygon;
+
+DECLSPEC IterCellsPolygon iterInitPolygon(const GeoPolygon *polygon, int res,
+                                          uint32_t flags);
+DECLSPEC void iterStepPolygon(IterCellsPolygon *iter);
+DECLSPEC void iterDestroyPolygon(IterCellsPolygon *iter);
+
+DECLSPEC H3Error H3_EXPORT(polygonToCellsExperimental)(
+    const GeoPolygon *polygon, int res, uint32_t flags, H3Index *out);
```
isaacbrodsky, line 82: For clarity, this is not intended to be part of the public API when this PR is merged (as it is not in h3api.h.in?)

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,593 @@
...
+        out->west = -M_PI;
+    }
+
+    return E_SUCCESS;
+}
+
+/**
+ * Get a base cell by number, or H3_NULL if out of bounds
+ */
+static H3Index getBaseCell(int baseCellNum) {
```
isaacbrodsky, line 269: This function name looks like it could do the opposite and accept and index and return a number, maybe rename to `indexFromBaseCell` or something?

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,593 @@
...
+    if (NEVER(baseCellNum < 0) || baseCellNum >= NUM_BASE_CELLS) {
```
isaacbrodsky, line 270: I don't think this needs `NEVER` if the function is marked non-static and a unit test exercises this case.

src/h3lib/lib/polyfill.c
```diff
@@ -0,0 +1,593 @@
...
+    }
+
+    if (flags != 0) {
+        iterErrorPolygonCompact(&iter, E_OPTION_INVALID);
+        return iter;
+    }
+
+    // Initialize bounding boxes for polygon and any holes. Memory allocated
+    // here must be released through iterDestroyPolygonCompact
+    iter._bboxes = H3_MEMORY(malloc)((polygon->numHoles + 1) * sizeof(BBox));
```
isaacbrodsky, line 364:
```suggestion
    iter._bboxes = H3_MEMORY(calloc)(polygon->numHoles + 1, sizeof(BBox));
```

isaacbrodsky (approved):

src/apps/testapps/testPolygonToCellsExperimental.c
```diff
@@ -0,0 +1,488 @@
...
+
+    pointGeoPolygon.geoloop = pointGeoLoop;
+    pointGeoPolygon.numHoles = 0;
+
+    lineGeoPolygon.geoloop = lineGeoLoop;
+    lineGeoPolygon.numHoles = 0;
+
+    // --------------------------------------------
+    // maxPolygonToCellsSize
+    // --------------------------------------------
```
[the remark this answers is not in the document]
nrabinowitz: These tests are duplicated. My intent was to have copy/paste versions of the old tests with the new function, so that when we drop the old function it would just be removing the `-Experimental` files. But I guess you're right that there's no point duplicating the `maxPolygonToCellsSize` tests, I can remove.

src/h3lib/include/polyfill.h
nrabinowitz: Correct - I added `DECLSPEC` so that it would be available for export, but I don't expect us to expose it in the API until we transparently replace the old function.

src/h3lib/lib/polyfill.c
nrabinowitz: Makes sense, will rename to `baseCellNumToIndex`

src/h3lib/lib/polyfill.c
nrabinowitz: True, can update. It's not clear to me whether `static` offers any compilation advantages, or whether it just indicates a local function. We settled on `static` to avoid `_underscore` names to denote local functions, but I guess it makes sense to test this separately.

src/h3lib/lib/polyfill.c
nrabinowitz: Will update, thanks

nrabinowitz:
Thanks everyone for the reviews and support on this!

PR #788, opened 2023-10-10
explicitly state parameters are , #787

### LINKED ISSUE #787: Replace empty function parameters with `void`
from: http://port70.net/~nsz/c/c99/n1256.html#6.11.6

> The use of function declarators with empty parentheses (not prototype-format parameter type declarators) is an obsolescent feature.

and [this SO answer](https://stackoverflow.com/a/41805712/5977215) suggest you should use `void` in the function parameter definition to explicitly say the function does not take any arguments.

Therefore, should these two functions be changed to use `void`, as in:
```c
//h3Index.c
int H3_EXPORT(pentagonCount)(void) { return NUM_PENTAGONS; }

//baseCells.c
int H3_EXPORT(res0CellCount)(void) { return NUM_BASE_CELLS; }
```

---

isaacbrodsky:
Fixes #787.


nrabinowitz (approved):

dcooley:
is it also required in `iterators.c` L45?

```c
static IterCellsChildren _null_iter(void)
```

dfellis (approved):

PR #790, opened 2023-10-20
aarch64 performance fix: replace long double with double

heshpdx:
On aarch64 systems, we are seeing a significant overhead in libgcc - __multf3.
```
26.28%  h3geo  libgcc_s.so.1      [.] __multf3
```
This overhead is due to the usage of lroundl in coordijk.c. To prevent this overhead, we are proposing to replace long double type with double type and use lround instead of lroundl. This should have no impact on other systems than aarch64. We have verified the behavior matches on some test cases. For coherency sake, we also changed long double constants to double type.

This was found by the SPEC CPU committee, where h3 is being considered as a candidate benchmark for [SPEC CPU v8](https://spec.org/cpuv8/) and undergoing testing across a wide variety of systems and compilers.

Feel free to squash these commits.

### DIFF
--- a/src/h3lib/lib/coordijk.c
+++ b/src/h3lib/lib/coordijk.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018, 2020-2022 Uber Technologies, Inc.
+ * Copyright 2016-2018, 2020-2023 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -67,7 +67,7 @@ void _hex2dToCoordIJK(const Vec2d *v, CoordIJK *h) {
 
     // first do a reverse conversion
     x2 = a2 / M_SIN60;
-    x1 = a1 + x2 / 2.0L;
+    x1 = a1 + x2 / 2.0;
 
     // check if we have the center of a hex
     m1 = x1;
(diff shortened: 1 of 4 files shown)


isaacbrodsky (approved):

dfellis (approved):
Thank you! This might provide a small perf boost on other platforms since it looks like all uses of long double get casted back into regular double, and none of them look like the extra 16 bits should impact the intermediate values except *maybe* the branches in `_hex2dToCoordIJK`? (But the comparisons of long-double intermediate values look like they are always against regular doubles, so I presume they get casted during the comparison and the extra bits of precision are dropped.)

grim7reaper:
FWIW, h3o only uses `double` (since `long double` no longer exists in Rust), so it's likely possible to get rid of `long double` on every platform without impacting the correctness

isaacbrodsky:
Thanks for the PR!

I remember removing some uses of `long double` before (I was checking if that was in the Git history and didn't see it, maybe it was before we open sourced the library?). I am pretty comfortable with removing remaining `long double` usage.

sahrk:
> Thanks for the PR!
> 
> I remember removing some uses of `long double` before (I was checking if that was in the Git history and didn't see it, maybe it was before we open sourced the library?). I am pretty comfortable with removing remaining `long double` usage.

Everything was `long double` when we started, until it became clear that was unnecessary. When we switched to `double` some of them must have been missed.

src/h3lib/lib/coordijk.c
```diff
@@ -346,8 +346,8 @@ H3Error _upAp7Checked(CoordIJK *ijk) {
     }
 
     // TODO: Do the int math parts here in long double?
```
nrabinowitz, line 348: I'm a little late here, but it looks like there are some TODO comments that are now out of date

Issue #489, opened 2021-06-28
Feature proposal: circleToCells


nrabinowitz:
[This SO question](https://stackoverflow.com/a/68166990/380487), and others where people asked for a radius lookup, made me think that there might be value in a circle-fill function that produced more accurate "cells in radius" output than `kRing`. One option here, which could make this fairly fast, is to calculate:

- the minimum grid radius `r1` required to contain the circle
- the maximum distance `r2` _guaranteed_ to be in the circle

Then you could use `kRingDistances(cell, r1)` and check only cells with a grid distance greater than `r2` using a true geo distance calculation to determine whether they should be in the output set.

dfellis:
The same containment configuration that we want to add to polyfill also applies here. Fully contained, centerpoint containment, and any part of the hexagon intersecting the circle are all things that different people will want for different reasons.

It may also make sense to make this reuse a lot of the upcoming polyfill changes for the above but just swapping out the point-in-polygon function for a point-in-circle function?

ajfriend:
+1

I've had plenty of situations where this function would have been super useful. :)

PR #796, opened 2023-11-04
Add support for full containment and overlapping modes in polygonToCellsExperimental

nrabinowitz:
Adds support for full containment mode and overlapping modes in `polygonToCellsExperimental`.

- Defines an enum for the containment modes. My assumption here is that the lowest 4 bits of the 32-bit flag are reserved for containment modes (not sure that we'd need more than 7, but wanted to be on the safe side), and that we'd use them as an integer, not as bitwise flags, since the modes are mutually exclusive.
- Implements the check for containment in `iterStepPolygonCompact`. This was very straightforward as we already have `cellBoundaryInsidePolygon` polygon available.
- Implements the check for overlapping cells in `iterStepPolygonCompact`. This checks center point inclusion first, then shares code with the full containment check, substituting a new `cellBoundaryCrossesPolygon` check for `cellBoundaryInsidePolygon`.
- Adds some very basic tests, which essentially just duplicate existing polyfill tests with full containment and overlapping turned on, substituting an empirically determined result and checking that it's reasonable compared to the center containment mode.
- Updates all tests to use modes instead of `0` for `flags`.

TODO:
- [x] We should probably clean up the tests and use `CENTER_CONTAINMENT` instead of `0` for the flag value in existing tests.
- [x] There are a couple of error branches here that I think we can get coverage on by sticking invalid cells into the iterator object.
- [ ] I'd love better tests here, though I'm not sure what would make sense.

### DIFF
--- a/scripts/make_countries.js
+++ b/scripts/make_countries.js
@@ -131,6 +131,7 @@ async function makeCountries(sourceUrl, targetPath) {
 #include "h3api.h"
 #include "mathExtensions.h"
 #include "polyfill.h"
+#include "polygon.h"
 
 const GeoPolygon COUNTRIES[${polygons.length}] = {${
   polygons.map((poly, i) => formatGeoPolygon(poly, names[i])).join(',')
@@ -149,13 +150,25 @@ for (int res = 0; res < MAX_RES + 1; res++) {
 
   BENCHMARK(polygonToCells_AllCountries1, 5, {
     for (int index = 0; index < ${polygons.length}; index++) {
-      H3_EXPORT(polygonToCells)(&COUNTRIES[index], res, 0, hexagons);
+      H3_EXPORT(polygonToCells)(&COUNTRIES[index], res, CENTER_CONTAINMENT, hexagons);
     }
   });
 
   BENCHMARK(polygonToCells_AllCountries2, 5, {
     for (int index = 0; index < ${polygons.length}; index++) {
-      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, 0, hexagons);
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, CENTER_CONTAINMENT, hexagons);
+    }
+  });
+
+  BENCHMARK(polygonToCells_AllCountries3, 5, {
+    for (int index = 0; index < ${polygons.length}; index++) {
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, FULL_CONTAINMENT, hexagons);
+    }
+  });
+
+  BENCHMARK(polygonToCells_AllCountries4, 5, {
+    for (int index = 0; index < ${polygons.length}; index++) {
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, OVERLAPPING, hexagons);
     }
   });
 
--- a/scripts/make_countries.js
+++ b/scripts/make_countries.js
@@ -150,25 +150,25 @@ for (int res = 0; res < MAX_RES + 1; res++) {
 
   BENCHMARK(polygonToCells_AllCountries1, 5, {
     for (int index = 0; index < ${polygons.length}; index++) {
-      H3_EXPORT(polygonToCells)(&COUNTRIES[index], res, CENTER_CONTAINMENT, hexagons);
+      H3_EXPORT(polygonToCells)(&COUNTRIES[index], res, CONTAINMENT_CENTER, hexagons);
     }
   });
 
   BENCHMARK(polygonToCells_AllCountries2, 5, {
     for (int index = 0; index < ${polygons.length}; index++) {
-      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, CENTER_CONTAINMENT, hexagons);
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, CONTAINMENT_CENTER, hexagons);
     }
   });
 
   BENCHMARK(polygonToCells_AllCountries3, 5, {
     for (int index = 0; index < ${polygons.length}; index++) {
-      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, FULL_CONTAINMENT, hexagons);
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, CONTAINMENT_FULL, hexagons);
     }
   });
 
   BENCHMARK(polygonToCells_AllCountries4, 5, {
     for (int index = 0; index < ${polygons.length}; index++) {
-      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, OVERLAPPING, hexagons);
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, CONTAINMENT_OVERLAPPING, hexagons);
     }
   });
 
--- a/src/h3lib/include/polygon.h
+++ b/src/h3lib/include/polygon.h
@@ -40,7 +40,22 @@
 /** Macro: Whether a GeoLoop is empty */
 #define IS_EMPTY_GEOFENCE(geoloop) geoloop->numVerts == 0
 
+/**
+ * Values representing polyfill containment modes, to be used in
+ * the `flags` bit field.
+ */
+typedef enum {
+    CENTER_CONTAINMENT = 0,
+    FULL_CONTAINMENT = 1,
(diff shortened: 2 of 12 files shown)


scripts/make_countries.js
```diff
@@ -149,13 +150,25 @@ for (int res = 0; res < MAX_RES + 1; res++) {
...
+
+  BENCHMARK(polygonToCells_AllCountries3, 5, {
+    for (int index = 0; index < ${polygons.length}; index++) {
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, FULL_CONTAINMENT, hexagons);
+    }
+  });
+
+  BENCHMARK(polygonToCells_AllCountries4, 5, {
+    for (int index = 0; index < ${polygons.length}; index++) {
+      H3_EXPORT(polygonToCellsExperimental)(&COUNTRIES[index], res, OVERLAPPING, hexagons);
```
nrabinowitz, line 171: The new modes are slower than center containment, unsurprisingly, but still manage to be faster than the old algo in some cases:
```
Res 0	-- polygonToCells_AllCountries1: 198222.000000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries2: 42474.800000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries3: 247580.600000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries4: 282227.200000 microseconds per iteration (5 iterations)
Res 1	-- polygonToCells_AllCountries1: 324889.800000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries2: 18640.000000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries3: 72821.200000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries4: 89494.400000 microseconds per iteration (5 iterations)
Res 2	-- polygonToCells_AllCountries1: 257009.400000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries2: 66020.600000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries3: 143866.400000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries4: 181394.600000 microseconds per iteration (5 iterations)
Res 3	-- polygonToCells_AllCountries1: 424776.400000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries2: 280514.200000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries3: 442108.400000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries4: 545078.600000 microseconds per iteration (5 iterations)
Res 4	-- polygonToCells_AllCountries1: 2035545.200000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries2: 1462475.200000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries3: 1969331.600000 microseconds per iteration (5 iterations)
	-- polygonToCells_AllCountries4: 2559085.000000 microseconds per iteration (5 iterations)
```

src/h3lib/include/polygon.h
```diff
@@ -40,18 +40,38 @@
...
+/**
+ * Values representing polyfill containment modes, to be used in
+ * the `flags` bit field.
+ */
+typedef enum {
+    CENTER_CONTAINMENT = 0,
+    FULL_CONTAINMENT = 1,
+    OVERLAPPING = 2,
+    INVALID_CONTAINMENT = 3
+} ContainmentMode;
```
isaacbrodsky, line 52: we may want to prefix the enun names with something here (`POLYGON_CENTER_CONTAINMENT`?) so that it is clear a valid enum is being fed into `polygonToCells`.

src/h3lib/include/polygon.h
```diff
@@ -40,18 +40,38 @@
...
 
+/**
+ * Values representing polyfill containment modes, to be used in
+ * the `flags` bit field.
+ */
+typedef enum {
+    CENTER_CONTAINMENT = 0,
+    FULL_CONTAINMENT = 1,
+    OVERLAPPING = 2,
+    INVALID_CONTAINMENT = 3
```
isaacbrodsky, line 51:
```suggestion
    CENTER_CONTAINMENT = 0, ///< Cells are included if their center is in the polygon
    FULL_CONTAINMENT = 1, ///< Cells are included if they are entirely contained in the polygon
    OVERLAPPING = 2, ///< Cells are included if any part of them overlaps the polygon
    INVALID_CONTAINMENT = 3 ///< This mode is invalid and should not be used.
```

src/h3lib/include/polygon.h
nrabinowitz: That makes sense - leaning towards `CONTAINMENT_` as the prefix, so we'd have

- `CONTAINMENT_CENTER`
- `CONTAINMENT_FULL`
- `CONTAINMENT_OVERLAPPING`
- `CONTAINMENT_INVALID`

One thought here is that we might apply the same modes to [circles](https://github.com/uber/h3/issues/489) or bounding boxes at some point, so I don't necessarily think we want to bake `POLYGON` into the name.

src/h3lib/lib/polyfill.c
```diff
@@ -406,23 +407,78 @@ void iterStepPolygonCompact(IterCellsPolygonCompact *iter) {
...
-            if (NEVER(centerErr != E_SUCCESS)) {
-                iterErrorPolygonCompact(iter, centerErr);
-                return;
+            if (mode == CONTAINMENT_CENTER || mode == CONTAINMENT_OVERLAPPING) {
+                // Check if the cell center is inside the polygon
+                LatLng center;
+                H3Error centerErr = H3_EXPORT(cellToLatLng)(cell, &center);
+                if (centerErr != E_SUCCESS) {
+                    iterErrorPolygonCompact(iter, centerErr);
+                    return;
```
isaacbrodsky, line 423: This (and similar case below) are uncovered in test, but look like they should be testable by setting an invalid cell on the iterator and then calling this function.

src/h3lib/lib/polyfill.c
```diff
@@ -406,23 +407,78 @@ void iterStepPolygonCompact(IterCellsPolygonCompact *iter) {
...
+                if (boundaryErr != E_SUCCESS) {
+                    iterErrorPolygonCompact(iter, boundaryErr);
+                    return;
+                }
+                BBox bbox;
+                H3Error bboxErr = cellToBBox(cell, &bbox, false);
+                if (bboxErr) {
+                    iterErrorPolygonCompact(iter, bboxErr);
+                    return;
+                }
```
isaacbrodsky, line 463: This case may not be testable because presumably the boundary error above would not be E_SUCCESS? Perhaps there exists some index for which cellToBoundary returns 0 but cellToBbox returns non-0?

isaacbrodsky: Related question, do we have fuzzer question of these functions?

src/h3lib/lib/polyfill.c
nrabinowitz: Yes, I'll try to add coverage here today or tomorrow.

src/h3lib/lib/polyfill.c
nrabinowitz: Not sure if this is coverable, will switch to `NEVER` if I can't figure out an approach.

We don't have a fuzzer for `polygonToCellsExperimental`, but I can add one - that's probably best in a separate PR.

isaacbrodsky: I would be happy to help getting a fuzzer added. Agreed doing that in a separate PR would make the most sense.

nrabinowitz:
@isaacbrodsky I believe the latest commit should close the holes in coverage.

isaacbrodsky (approved):

michaelbrichko:
Thanks for adding this functionality !!
How mature is it? Are there any knows issues for overlapping functionality? Any known gaps ?

nrabinowitz:
> Thanks for adding this functionality !! How mature is it? Are there any knows issues for overlapping functionality? Any known gaps ?

Well, you can see it was merged two weeks ago, so that's how mature it is :). But we've done a fair amount of testing against the older algo, and it seems to behave as expected. Overlapping mode seems to work as expected, we aren't aware of any issues, but it's hard to extensively test this. The main (expected) issue is that overlapping mode is slower than the other modes, as discussed in this PR.

michaelbrichko:
Ok, thanks for quick answer !

BrendanJM:
Look great! Has this been released yet? If not, is there a stable tag we could reference? We use the h3-py bindings and it looks like it’s been a bit since the last release.

PR #800, opened 2023-11-16
Polygon to cells experimental fuzzer

isaacbrodsky:
This adds new fuzzers for the `polygonToCellsExperimental` function, based on the existing functions. Since #796 (this PR is based on that branch) adds containment modes, this updates the existing fuzzers to exercise those options too.

When I run the `fuzzerPolygonToCellsExperimental`, I see some runtime errors about misaligned addresses, which doesn't cause an issue on Apple M1 architecture but might indicate trouble on other architectures. When I run `fuzzerPolygonToCellsExperimentalNoHoles` I get a crash about `heap-buffer-overflow` around `polygonToCellsExperimental polyfill.c:646` (looks like the output exceeds the allocated output buffer in some case?)


src/apps/fuzzers/fuzzerPolygonToCells.c
```diff
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 /** @file
- * @brief Fuzzer program for polygonToCells and related functions
+ * @brief Fuzzer program for polygonToCellsExperimental and related functions
```
nrabinowitz, line 17: This is incorrect, right? This fuzzes the old function

nrabinowitz (commented):
My guess is that `maxPolygonToCellsSize` doesn't allocate enough cells for the new modes in some cases. I've seen similar behavior here in benchmarks - this could be an issue that affects the old `polygonToCells` as well.

I have a plan for a new `maxPolygonToCellsSize` algo that I'll try out, this may fix the fuzzer errors.

nrabinowitz:
I think the/one issue here is polar shapes. When I run all countries, I see differences in output for Antarctica (not sure why I didn't catch this earlier):

image

My new estimator seems to cover this case, though it has other issues I'm still debugging. I haven't had a chance yet to check what the old/new output looks like, though I'd guess that the new output is closer to correct due to better handling of the pole itself.

nrabinowitz:
Old polyfill:
image
New polyfill:
image

The new one may be correct, I'd have to import the original polygon to check

src/apps/fuzzers/fuzzerPolygonToCells.c
isaacbrodsky: Yes, that's a mistake

isaacbrodsky:
> The new one may be correct, I'd have to import the original polygon to check

Has this polygon been added as a test case to ensure we aren't dropping a lot of cells there?

nrabinowitz:
> > The new one may be correct, I'd have to import the original polygon to check
> 
> Has this polygon been added as a test case to ensure we aren't dropping a lot of cells there?

I'd rather add some simplified version, as the Antarctica polygon is too large to be easily added to a test file. I can add a new PR with more tests for this.

nrabinowitz:
Still seeing errors here with the new estimator. Is there an easy-ish way to get the input for the failure cases?

src/h3lib/lib/polygon.c
```diff
@@ -130,10 +130,12 @@ bool cellBoundaryInsidePolygon(const GeoPolygon *geoPolygon, const BBox *bboxes,
 
     // Check for line intersections with, or containment of, any hole
     for (int i = 0; i < geoPolygon->numHoles; i++) {
-        if (pointInsideGeoLoop(&boundaryLoop, boundaryBBox,
-                               &geoPolygon->holes[i].verts[0]) ||
-            cellBoundaryCrossesGeoLoop(&(geoPolygon->holes[i]), &bboxes[i + 1],
-                                       boundary, boundaryBBox)) {
+        // If the hole has no verts, it is not possible to intersect with it.
+        if (geoPolygon->holes[i].numVerts > 0 &&
```
isaacbrodsky, line 134: This probably needs an additional test to be added

isaacbrodsky:
I added a test case that demonstrates this. (If you run under valgrind it will show an issue.) It seems we do not actually test a polygon with a single vertex being passed in. If you do so with res=0, flags=2 (CONTAINMENT_OVERLAPPING), the estimate will be 0 but the function will try to write the buffer.

src/h3lib/lib/polyfill.c
```diff
@@ -418,6 +418,12 @@ void iterStepPolygonCompact(IterCellsPolygonCompact *iter) {
         iter->_started = true;
     }
 
+    if (iter->_polygon->geoloop.numVerts == 0) {
```
isaacbrodsky, line 421: Should this be changed to 0 or 1 returns nothing? @nrabinowitz

nrabinowitz: I think the case for 1 and 2 vertexes is equivalent - in both cases, we have input that cannot be a 2D polygon, and we either have to discard it or treat it like a shape. While it's obviously a pathological case, I think it would be reasonable to assume that If you have a one-vertex "polygon" and you use `CONTAINMENT_OVERLAPPING` that you'd get the cell that contains the point. There's also a potential use case for line segments in the 2-vertex case. I'm also concerned that the estimate and the implementation don't line up.

Should I take a shot at a PR to fix this, which we can land before this one?

isaacbrodsky: I agree that seems like the expected behavior, if you have an idea for addressing please take a look.

I can help apply these tests to another branch to help validate.

isaacbrodsky:
Here is the hexdump of a new crash case for `fuzzerPolygonToCellsExperimental`:
```
$ hexdump ./crash-7bee58752754302154762e5e3a399235e773c80c
0000000 0000 4400 0005 0000 0001 0000 0000 0000
0000010 0000 0000 0000 0000 0000 0000 0009 0000
0000020 0000 0000 0000 0000 0000 9f00 9f9f 9f9f
0000030 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 6069
0000040 6060 6060 6060 9f9f 9f9f 9f9f 9f9f 9f9f
0000050 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
*
0000090 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 3939 3939
00000a0 3939 3939 3939 3939 0039 0000 0000 0000
00000b0 0000 0000 0000 0000 0000 0000 0000 0000
*
0000180 9f00 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
0000190 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
*
00001c0 9f9f 9f9f 9f9f 9f2a 9f9f 9f9f 9f9f 9f9f
00001d0 9f9f 9f9f 9f9f 399f 3939 3939 3939 3939
00001e0 3939 3939 0000 0000 0000 0000 0000 0000
00001f0 0000 0000 0000 0000 0000 0000 0000 0000
0000200 0000 0000 0544 0000 0000 0000 0000 0000
0000210 0000 0040 0000 0000 0000 0000 9f9f 9f9f
0000220 9f9f 9f9f 9f9a 9f9f 9f9f 9f9f 9f9f 9f9f
0000230 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
*
0000260 9f9f 9f9f 9f9f 9f9f 9f9f 399f 3939 3939
0000270 3b39 3939 3939 3939 3939 3939 3939 3939
0000280 3939 3939 3939 3939 3939 3939 3939 3939
0000290 3939 3939 3939 3939 3939 9f38 9f9f 9f9f
00002a0 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
*
00002c0 9f9f 9f9f 9f9f 9f89 9f9f 9f9f 409f 9f9f
00002d0 9f9f fc9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
00002e0 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
00002f0 9f9f 9f9f 9f9f 0000 0000 0000 0000 0000
0000300 0000 0000 0000 0000 0000 0000 0000 9f9f
0000310 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
*
0000340 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 009f
0000350 0000 0000 0000 0000 9f9f 9f9f 9f9f 9f9f
0000360 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
*
0000380 3939 3939 3939 3939 3939 3939 0039 0000
0000390 0000 0000 0800 0000 0000 0000 0000 0000
00003a0 0000 0000 0000 0000 9f9f 9f9f 9f9f 9f9f
00003b0 9fe7 9f9f 9f9f 9f9f 9f9f 0c01 0000 0000
00003c0 0000 0000 0000 0000 9f00 9f9f 9f9f 9f9f
00003d0 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f 9f9f
*
0000430 9f9f 9f9f 9f9f 9f9f 0000 0800 0000 0000
0000440 0000 0000 0000 9f00 9f9f 0000          
000044b
```

isaacbrodsky:
I tried adding an additional check today to recompute the expected array size and bail if the function is about to write past the end of the array. That seems to, at the cost of recomputing the size, avoid any remaining crashes. Unfortunately there is a timeout with some input which is currently failing the test.

src/h3lib/lib/polyfill.c
```diff
@@ -433,7 +433,8 @@ void iterStepPolygonCompact(IterCellsPolygonCompact *iter) {
 
         // Target res: Do a fine-grained check
         if (cellRes == iter->_res) {
-            if (mode == CONTAINMENT_CENTER || mode == CONTAINMENT_OVERLAPPING) {
+            if (mode == CONTAINMENT_CENTER || mode == CONTAINMENT_OVERLAPPING ||
+                mode == CONTAINMENT_OVERLAPPING_BBOX) {
```
nrabinowitz, line 437: Hm. These additional checks might get the fuzzer to pass, but they miss the point of `CONTAINMENT_OVERLAPPING_BBOX`, which is to do a much faster check than `CONTAINMENT_OVERLAPPING`. The perf impact here is likely significant, since we're going from a fast bbox check to a slow set of polygon-based checks.

Are all of these additional checks needed to make the fuzzer pass? Or can we narrow down to find the check that's actually missed by `CONTAINMENT_OVERLAPPING_BBOX`?

isaacbrodsky: All but the change around line 511 seem to be necessary to prevent crashes.

nrabinowitz: I admit that I'm not seeing any significant differences in the benchmarks after this change, assuming I'm running them correctly. I think I'm ok with this to get the new algo out the door, though I'd really like to look through the logic here and understand what's missing in the bbox check - it doesn't make a lot of sense to me conceptually that these would work when the bbox doesn't, so we must be messing up the bbox check in some way.

src/h3lib/lib/polyfill.c
```diff
@@ -692,9 +696,27 @@ void iterDestroyPolygon(IterCellsPolygon *iter) {
 H3Error H3_EXPORT(polygonToCellsExperimental)(const GeoPolygon *polygon,
                                               int res, uint32_t flags,
                                               H3Index *out) {
+#ifdef H3_POLYGON_TO_CELLS_ASSERT
```
nrabinowitz, line 699: Should we comment on how to set this variable?  Can we use  the `TESTONLY` macro here instead?

nrabinowitz (approved):
As noted elsewhere, this seems ok for now, but we should definitely investigate why it's necessary, as the changes to make the fuzzer pass could have significant perf impact.

ajfriend:
> > > The new one may be correct, I'd have to import the original polygon to check
> > 
> > 
> > Has this polygon been added as a test case to ensure we aren't dropping a lot of cells there?
> 
> I'd rather add some simplified version, as the Antarctica polygon is too large to be easily added to a test file. I can add a new PR with more tests for this.

Do we still need to add a simplified Antartica-like polygon to the tests? I could help with that.

ajfriend (approved):

isaacbrodsky:
I'm overriding the CLA check since the only people with commits in this branch are myself and @nrabinowitz (who may have used an old email address, but has certainly signed the CLA before)

PR #812, opened 2024-01-17
`hierarchy.mdx`: Fix call to nonexistent `compact` in `js` example for `uncompactCells`

ericbf:
The `uncompactCells` example for `js` makes a call to `compact`, which no longer exists. It should use `compactCells`.


isaacbrodsky (approved):

nrabinowitz:
Thanks for the fix!

PR #816, opened 2024-02-09
Add fixes and tests for 0-, 1-, and 2-vertex geoloops and holes

nrabinowitz:
This includes the fixes from #800, plus one more fix for estimating a 1-vertex polygon. Includes the following tests:
- 0-vertex polygon, all modes
- 1-vertex polygon, all modes
- 2-vertex polygon, all modes
- polygon with 0-vertex hole, all modes
- polygon with 1-vertex hole, all modes
- polygon with 2-vertex hole, all modes

Hopefully #800 will be green once rebased on top of this.

### DIFF
--- a/src/apps/testapps/testPolygonToCellsExperimental.c
+++ b/src/apps/testapps/testPolygonToCellsExperimental.c
@@ -622,48 +622,24 @@ SUITE(polygonToCells) {
     }
 
     TEST(polygonToCellsNullPolygon) {
-        int64_t numHexagons;
-        t_assertSuccess(H3_EXPORT(maxPolygonToCellsSizeExperimental)(
-            &nullGeoPolygon, 9, CONTAINMENT_CENTER, &numHexagons));
-        t_assert(numHexagons == 0, "got expected estimated size");
-        H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
+        for (int res = 0; res < MAX_H3_RES; res++) {
+            for (uint32_t flags = 0; flags < CONTAINMENT_INVALID; flags++) {
+                int64_t numHexagons;
+                t_assertSuccess(H3_EXPORT(maxPolygonToCellsSizeExperimental)(
+                    &nullGeoPolygon, res, flags, &numHexagons));
+                t_assert(numHexagons == 0, "got expected estimated size");
+                H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
 
-        t_assertSuccess(H3_EXPORT(polygonToCellsExperimental)(
-            &nullGeoPolygon, 9, CONTAINMENT_CENTER, hexagons));
-        int64_t actualNumIndexes = countNonNullIndexes(hexagons, numHexagons);
+                t_assertSuccess(H3_EXPORT(polygonToCellsExperimental)(
+                    &nullGeoPolygon, res, flags, hexagons));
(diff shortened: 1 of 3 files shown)


src/apps/testapps/testPolygonToCellsExperimental.c
```diff
@@ -598,6 +621,282 @@ SUITE(polygonToCells) {
...
+        t_assert(numHexagons == 0, "got expected estimated size");
+        H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
+
+        t_assertSuccess(H3_EXPORT(polygonToCellsExperimental)(
+            &nullGeoPolygon, 9, CONTAINMENT_OVERLAPPING, hexagons));
+        int64_t actualNumIndexes = countNonNullIndexes(hexagons, numHexagons);
+
+        t_assert(actualNumIndexes == 0, "got expected polygonToCells size");
+        free(hexagons);
+    }
```
isaacbrodsky, line 667: maybe these could be collapsed to the same test and just loop through which flags to test? Might be easier to read that way?

isaacbrodsky (approved):

src/apps/testapps/testPolygonToCellsExperimental.c
nrabinowitz: True, that's what you did in #800. I can collapse this one, and the null hole, but not the other two (where FULL behaves differently)

nrabinowitz: Oh, actually this is the only one I can collapse, the others have different results per mode

isaacbrodsky (approved):

src/apps/testapps/testPolygonToCellsExperimental.c
```diff
@@ -598,6 +621,258 @@ SUITE(polygonToCells) {
...
+        t_assert(numHexagons == 1, "got expected estimated size");
+        H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
+
+        t_assertSuccess(H3_EXPORT(polygonToCellsExperimental)(
+            &pointGeoPolygon, 9, CONTAINMENT_OVERLAPPING, hexagons));
+        int64_t actualNumIndexes = countNonNullIndexes(hexagons, numHexagons);
+
+        t_assert(actualNumIndexes == 1, "got expected polygonToCells size");
+        free(hexagons);
+    }
```
isaacbrodsky, line 688: There is a test missing in this PR for polygonToCellsPointPolygon_overlappingBbox (and same for the below), which is why the fuzzer is still failing

src/h3lib/lib/polyfill.c
```diff
@@ -700,6 +706,17 @@ static double getAverageCellArea(int res) {
...
+    // Special case: 0-vertex polygon
+    if (polygon->geoloop.numVerts == 0) {
+        *out = 0;
+        return E_SUCCESS;
+    }
+    // Special case: 1-vertex polygon
+    if (polygon->geoloop.numVerts == 1) {
+        *out = 1;
+        return E_SUCCESS;
+    }
```
isaacbrodsky, line 718: In overlapping BBOX mode, this will not be correct as more than one cell bounding box may overlap a vertex.

nrabinowitz: Gotcha. I'll have a new PR up for this soon.

PR #817, opened 2024-02-12
Fix for edge cases with bbox overlap and out-of-bounds polygons

nrabinowitz:
More follow-up for #800:
- Adds tests for `CONTAINMENT_OVERLAPPING_BBOX` and fixes the out-of-bounds write when using bbox overlap on a one-vertex polygon
- Adds a simpler version of [this test case](https://github.com/uber/h3/pull/800/commits/673046900ad0eabaa65bcd66c508047ce24af806) and fixes `CONTAINMENT_OVERLAPPING` results for polygons outside of the normal lat/lng range. The issue here was the test for cell-contains-polygon, which used `latLngToCell` - this function will return results for out-of-bounds input, presumably wrapping around the sphere, but because the rest of the logic (including the estimator) in `polygonToCells` is using cartesian coords, these false positives were unexpected. I updated this to include a bbox check first, which will exclude polygons with all points outside of the normal range.

### DIFF
--- a/src/apps/testapps/testPolyfillInternal.c
+++ b/src/apps/testapps/testPolyfillInternal.c
@@ -36,14 +36,6 @@ static GeoPolygon sfGeoPolygon = {
                                     {0.6599990002976, -2.1376771158464}}},
     .numHoles = 0};
 
-static GeoPolygon invalidGeoPolygon = {
-    .geoloop = {.numVerts = 4,
-                .verts = (LatLng[]){{NAN, -2.1364398519396},
-                                    {0.6595011102219, NAN},
-                                    {NAN, -2.1354884206045},
-                                    {0.6581220034068, NAN}}},
-    .numHoles = 0};
-
 SUITE(polyfillInternal) {
     TEST(iterInitPolygonCompact_errors) {
         IterCellsPolygonCompact iter;
@@ -139,24 +131,6 @@ SUITE(polyfillInternal) {
         iterStepPolygonCompact(&iter);
         t_assert(iter.error == E_RES_DOMAIN,
                  "Got expected error for too-fine cell");
-        t_assert(iter.cell == H3_NULL, "Got null output for invalid cell");
-    }
-
-    TEST(iterStepPolygonCompact_invalidPolygonErrors) {
-        IterCellsPolygonCompact iter;
-
-        // Start with a good polygon, otherwise we error out early
-        iter =
-            iterInitPolygonCompact(&sfGeoPolygon, 5, CONTAINMENT_OVERLAPPING);
-        t_assertSuccess(iter.error);
-
-        // Give the iterator a bad polygon and a cell at target res
-        iter._polygon = &invalidGeoPolygon;
-        iter.cell = 0x85283473fffffff;
-
-        iterStepPolygonCompact(&iter);
-        t_assert(iter.error == E_LATLNG_DOMAIN,
-                 "Got expected error for invalid polygon");
         t_assert(iter.cell == H3_NULL, "Got null output for invalid cell");
     }
 
--- a/src/apps/testapps/testPolygonToCellsExperimental.c
+++ b/src/apps/testapps/testPolygonToCellsExperimental.c
@@ -50,6 +50,11 @@ static LatLng invalidVerts[] = {{INFINITY, INFINITY}, {-INFINITY, -INFINITY}};
 static GeoLoop invalidGeoLoop = {.numVerts = 2, .verts = invalidVerts};
 static GeoPolygon invalidGeoPolygon;
 
+static LatLng outOfBoundsVert[] = {{-2000, -2000}};
+static GeoLoop outOfBoundsVertGeoLoop = {.numVerts = 1,
+                                         .verts = outOfBoundsVert};
+static GeoPolygon outOfBoundsVertGeoPolygon;
+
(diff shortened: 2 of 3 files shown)


src/apps/testapps/testPolyfillInternal.c
```diff
@@ -142,24 +134,6 @@ SUITE(polyfillInternal) {
         t_assert(iter.cell == H3_NULL, "Got null output for invalid cell");
     }
 
-    TEST(iterStepPolygonCompact_invalidPolygonErrors) {
```
nrabinowitz, line 145: This test covered a block that AFAIK is now unreachable.

src/apps/testapps/testPolygonToCellsExperimental.c
```diff
@@ -643,48 +707,105 @@ SUITE(polygonToCells) {
...
-        int64_t actualNumIndexes = countNonNullIndexes(hexagons, numHexagons);
-
-        t_assert(actualNumIndexes == 0, "got expected polygonToCells size");
-        free(hexagons);
+        for (int res = 0; res < MAX_H3_RES; res++) {
+            int64_t numHexagons;
+            t_assertSuccess(H3_EXPORT(maxPolygonToCellsSizeExperimental)(
+                &pointGeoPolygon, res, CONTAINMENT_CENTER, &numHexagons));
+            t_assert(numHexagons >= 1 && numHexagons <= 5,
+                     "got expected estimated size");
```
nrabinowitz, line 715: Updated these assertions to expect a reasonable range rather than a specific value

isaacbrodsky: Is 5 the actual max that could be contained? I guess it doesn't really matter as long as it's equal or above the actual number that could be contained.

src/apps/testapps/testPolygonToCellsExperimental.c
```diff
@@ -398,6 +437,22 @@ SUITE(polygonToCells) {
         free(hexagons);
     }
 
+    TEST(polygonToCellsContainsPolygon_OverlappingBBox) {
```
isaacbrodsky, line 440:
```suggestion
    TEST(polygonToCellsContainsPolygon_overlappingBBox) {
```
nitpicking

src/apps/testapps/testPolygonToCellsExperimental.c
```diff
@@ -226,6 +234,21 @@ SUITE(polygonToCells) {
         free(hexagons);
     }
 
+    TEST(polygonToCells_OverlappingBBox) {
```
isaacbrodsky, line 237:
```suggestion
    TEST(polygonToCells_overlappingBBox) {
```
nitpicking

src/h3lib/lib/polyfill.c
```diff
@@ -448,20 +450,31 @@ void iterStepPolygonCompact(IterCellsPolygonCompact *iter) {
...
+                LatLng firstVertex = iter->_polygon->geoloop.verts[0];
+
+                // We have to check whether the point is in the expected range
+                // first, because out-of-bounds values will yield false
+                // positives with latLngToCell
+                if (bboxContains(&VALID_RANGE_BBOX, &firstVertex)) {
+                    H3Index polygonCell;
+                    H3Error polygonCellErr = H3_EXPORT(latLngToCell)(
+                        &(iter->_polygon->geoloop.verts[0]), cellRes,
+                        &polygonCell);
```
isaacbrodsky, line 467: this is only an optimization so it does not matter if the first vertex would actually overlap the cell or if other parts of the polygon overlap the cell?

isaacbrodsky (approved):

src/h3lib/lib/polyfill.c
nrabinowitz: This check is here because our other logic didn't catch polygons that were entirely contained by the cell (a situation that only applies to `CONTAINMENT_OVERLAPPING`). This catches that case, and additionally catches other cases that would be caught by the more expensive line-intersection check. So it's not an optimization - it's required for the cell-contains-polygon case - but it might cheaply catch other cases as well.

Does that answer the question?

src/apps/testapps/testPolygonToCellsExperimental.c
nrabinowitz: I think the max I saw in tests was 4, but 5 seemed reasonable 🤷

src/h3lib/lib/polyfill.c
isaacbrodsky: Might it be the case that a user passes in a polygon that is entirely contained by a cell, but the first vertex is expressed as over the antimeridian (> M_PI)? In that case, wouldn't it be necessary to check the later vertexes since one of them might be validly in range? Or that the user would expect that single out-of-range vertex to result in latLngToCell?

nrabinowitz: I'm comfortable requiring input for `polygonToCells` to be within valid coordinate range if you want reliable output. Because we use Cartesian logic for most of the checks here, coordinates > M_PI are likely to fail in a number of places. We should definitely guard against bad input causing segfaults or insecure memory access, but I don't feel like we need to try to produce good results for bad input. FWIW I'm pretty sure our previous algo wouldn't properly fill a polygon with out-of-bounds coords either.

isaacbrodsky: I am ok with this if it not worse than the previous algorithm.

PR #818, opened 2024-02-22
Implement the cellToBoundary command for the h3 binary

dfellis:
Just this one command at first as I remember how we had structured the arg parsing API. Tested for correctness manually, I will see if/how testing this binary was set up shortly.

### DIFF
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Uber Technologies, Inc.
+ * Copyright 2021, 2024 Uber Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -53,7 +53,10 @@ bool cellToLatLngCmd(int argc, char *argv[]) {
         return helpArg.found;
     }
     LatLng ll;
-    H3_EXPORT(cellToLatLng)(cell, &ll);
+    int err = H3_EXPORT(cellToLatLng)(cell, &ll);
+    if (err) {
+        return false;
+    }
     printf("%.10lf, %.10lf\n", H3_EXPORT(radsToDegs)(ll.lat),
            H3_EXPORT(radsToDegs)(ll.lng));
     return true;
@@ -110,6 +113,38 @@ bool latLngToCellCmd(int argc, char *argv[]) {
     return true;
 }
 
+bool cellToBoundaryCmd(int argc, char* argv[]) {
+    Arg cellToBoundaryArg = {
+        .names = {"cellToBoundary"},
+        .required = true,
+        .helpText = "Convert an H3 cell to an array of latitude/longitude coordinates defining its boundary",
+    };
+    Arg helpArg = ARG_HELP;
+    DEFINE_CELL_ARG(cell, cellArg);
+    Arg *args[] = {&cellToBoundaryArg, &helpArg, &cellArg};
+    if (parseArgs(argc, argv, 3, args, &helpArg,
+                  "Convert an H3 cell to an array of latitude/longitude coordinates defining its boundary")) {
+        return helpArg.found;
+    }
+    CellBoundary cb;
+    int err = H3_EXPORT(cellToBoundary)(cell, &cb);
+    if (err) {
+        return false;
+    }
+    // TODO: Is it better to use brackets or parentheses for the output?
+    printf("[\n");
+    for (int i = 0; i < cb.numVerts; i++) {
+        LatLng *ll = &cb.verts[i];
+        printf(
+            "  [%.10lf, %.10lf],\n",
+            H3_EXPORT(radsToDegs)(ll->lat),
+            H3_EXPORT(radsToDegs)(ll->lng)
+        );
+    }
+    printf("]\n");
+    return true;
+}
+
 bool generalHelp(int argc, char *argv[]) {
     Arg helpArg = ARG_HELP;
     Arg cellToLatLngArg = {
@@ -121,12 +156,16 @@ bool generalHelp(int argc, char *argv[]) {
         .helpText =
             "Convert degrees latitude/longitude coordinate to an H3 cell.",
     };
-    Arg *args[] = {&helpArg, &cellToLatLngArg, &latLngToCellArg};
+    Arg cellToBoundaryArg = {
+        .names = {"cellToBoundary"},
+        .helpText = "Convert an H3 cell to an array of latitude/longitude coordinates defining its boundary",
+    };
+    Arg *args[] = {&helpArg, &cellToLatLngArg, &latLngToCellArg, &cellToBoundaryArg};
     const char *helpText =
         "Please use one of the subcommands listed to perform an H3 "
         "calculation. Use h3 <SUBCOMMAND> --help for details on the usage of "
         "any subcommand.";
-    return parseArgs(argc, argv, 3, args, &helpArg, helpText);
+    return parseArgs(argc, argv, 4, args, &helpArg, helpText);
 }
 
 int main(int argc, char *argv[]) {
@@ -138,6 +177,9 @@ int main(int argc, char *argv[]) {
         return 0;
     }
     if (has("latLngToCell", 1, argv) && latLngToCellCmd(argc, argv)) {
+        return 0;
+    }
+    if (has("cellToBoundary", 1, argv) && cellToBoundaryCmd(argc, argv)) {
         return 0;
     }
     if (generalHelp(argc, argv)) {
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -164,7 +164,7 @@ bool generalHelp(int argc, char *argv[]) {
     };
     Arg *args[] = {&helpArg, &cellToLatLngArg, &latLngToCellArg,
                    &cellToBoundaryArg};
-    #define ARGLEN 4
+#define ARGLEN 4
 
     const char *helpText =
         "Please use one of the subcommands listed to perform an H3 "
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -59,8 +59,8 @@ bool cellToLatLngCmd(int argc, char *argv[]) {
     }
     // Using WKT formatting for the output. TODO: Add support for JSON
     // formatting
-    printf("POINT(%.10lf %.10lf)\n", H3_EXPORT(radsToDegs)(ll.lat),
-           H3_EXPORT(radsToDegs)(ll.lng));
+    printf("POINT(%.10lf %.10lf)\n", H3_EXPORT(radsToDegs)(ll.lng),
+           H3_EXPORT(radsToDegs)(ll.lat));
     return true;
 }
 
@@ -142,12 +142,12 @@ bool cellToBoundaryCmd(int argc, char *argv[]) {
     printf("POLYGON((");
     for (int i = 0; i < cb.numVerts; i++) {
         LatLng *ll = &cb.verts[i];
-        printf("%.10lf %.10lf, ", H3_EXPORT(radsToDegs)(ll->lat),
-               H3_EXPORT(radsToDegs)(ll->lng));
+        printf("%.10lf %.10lf, ", H3_EXPORT(radsToDegs)(ll->lng),
+               H3_EXPORT(radsToDegs)(ll->lat));
     }
     // WKT has the first and last points match, so re-print the first one
-    printf("%.10lf %.10lf))\n", H3_EXPORT(radsToDegs)(cb.verts[0].lat),
-           H3_EXPORT(radsToDegs)(cb.verts[0].lng));
+    printf("%.10lf %.10lf))\n", H3_EXPORT(radsToDegs)(cb.verts[0].lng),
+           H3_EXPORT(radsToDegs)(cb.verts[0].lat));
     return true;
 }
 
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -164,11 +164,13 @@ bool generalHelp(int argc, char *argv[]) {
     };
     Arg *args[] = {&helpArg, &cellToLatLngArg, &latLngToCellArg,
                    &cellToBoundaryArg};
+    #define ARGLEN 4
+
     const char *helpText =
         "Please use one of the subcommands listed to perform an H3 "
         "calculation. Use h3 <SUBCOMMAND> --help for details on the usage of "
         "any subcommand.";
-    return parseArgs(argc, argv, 4, args, &helpArg, helpText);
+    return parseArgs(argc, argv, ARGLEN, args, &helpArg, helpText);
 }
 
 int main(int argc, char *argv[]) {
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -113,17 +113,20 @@ bool latLngToCellCmd(int argc, char *argv[]) {
     return true;
 }
 
-bool cellToBoundaryCmd(int argc, char* argv[]) {
+bool cellToBoundaryCmd(int argc, char *argv[]) {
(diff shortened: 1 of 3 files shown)


dfellis:
So... I don't see any tests for the `h3` binary at all. Will like to discuss with you guys how we want to test it.

src/apps/filters/h3.c
```diff
@@ -53,7 +53,10 @@ bool cellToLatLngCmd(int argc, char *argv[]) {
         return helpArg.found;
     }
     LatLng ll;
-    H3_EXPORT(cellToLatLng)(cell, &ll);
+    int err = H3_EXPORT(cellToLatLng)(cell, &ll);
```
isaacbrodsky, line 56:
```suggestion
    H3Error err = H3_EXPORT(cellToLatLng)(cell, &ll);
```

src/apps/filters/h3.c
```diff
@@ -110,6 +113,38 @@ bool latLngToCellCmd(int argc, char *argv[]) {
...
+        return false;
+    }
+    // TODO: Is it better to use brackets or parentheses for the output?
+    printf("[\n");
+    for (int i = 0; i < cb.numVerts; i++) {
+        LatLng *ll = &cb.verts[i];
+        printf("  [%.10lf, %.10lf],\n", H3_EXPORT(radsToDegs)(ll->lat),
+               H3_EXPORT(radsToDegs)(ll->lng));
+    }
+    printf("]\n");
```
dfellis, line 144: This output is *not quite* JSON-ish, at the moment. Do we want it to conform to JSON? (If so, we should change the other outputs, too.) Or do we want to focus on human readability? (If so, should this be switched to parentheses between coordinates and no outer parens/brackets?)

src/apps/filters/h3.c
```diff
@@ -110,6 +113,38 @@ bool latLngToCellCmd(int argc, char *argv[]) {
...
+    return true;
```
isaacbrodsky, line 145: it occurs to me we may want to have different forms of the output, and particularly we may want to support WKT encoding

src/apps/filters/h3.c
```diff
@@ -110,6 +113,38 @@ bool latLngToCellCmd(int argc, char *argv[]) {
...
+    Arg helpArg = ARG_HELP;
+    DEFINE_CELL_ARG(cell, cellArg);
+    Arg *args[] = {&cellToBoundaryArg, &helpArg, &cellArg};
+    if (parseArgs(argc, argv, 3, args, &helpArg,
+                  "Convert an H3 cell to an array of latitude/longitude "
+                  "coordinates defining its boundary")) {
+        return helpArg.found;
+    }
+    CellBoundary cb;
+    int err = H3_EXPORT(cellToBoundary)(cell, &cb);
```
isaacbrodsky, line 133:
```suggestion
    H3Error err = H3_EXPORT(cellToBoundary)(cell, &cb);
```

src/apps/filters/h3.c
```diff
@@ -121,12 +156,19 @@ bool generalHelp(int argc, char *argv[]) {
...
+            "defining its boundary",
+    };
+    Arg *args[] = {&helpArg, &cellToLatLngArg, &latLngToCellArg,
+                   &cellToBoundaryArg};
     const char *helpText =
         "Please use one of the subcommands listed to perform an H3 "
         "calculation. Use h3 <SUBCOMMAND> --help for details on the usage of "
         "any subcommand.";
-    return parseArgs(argc, argv, 3, args, &helpArg, helpText);
+    return parseArgs(argc, argv, 4, args, &helpArg, helpText);
```
isaacbrodsky, line 171: 4 is the number of arguments? Maybe add a named constant or comment here to remind us what this is.

dfellis, line 171: The duplication of effort with this help function and needing to increment this little integer on each new verb is kinda annoying. It'd be nice if we could generate this function with a macro or something.

isaacbrodsky (approved):

src/apps/filters/h3.c
dfellis: Yes, it's basically the length of the `args` array, and it would be nice to not need to provide it, but I'll at least make it a CONSTANT to be clearer, since I missed needing to change this at first.

dfellis: @isaacbrodsky any thoughts about potentially generating the entire `generalHelp` function with a macro (that would presumably be fed by other macros the other sub commands use)? Would it be too magical?

src/apps/filters/h3.c
isaacbrodsky: I would suggest offering WKT and GeoJSON as output formats. WKT is I think close enough to this kind of "human readable" form.

src/apps/filters/h3.c
isaacbrodsky: I'm not sure exactly how you would write such a macro, since you'd need to append to a list, but in principle that sounds OK to me, since the macro magic would be limited to this CLI application.

sahrk: > needing to increment this little integer on each new verb is kinda annoying. It'd be nice if we could generate this function with a macro or something.

I think that you could avoid the manual incrementing of `numArgs` by using `sizeof(args) / sizeof(Arg *)` instead, which could be defined as a macro.

src/apps/filters/h3.c
dfellis: Getting back to this again with some spare time. So I am planning on just making it all WKT in this PR and make JSON formatting a follow-up PR. (Or I could reverse that ordering if you prefer.)

I've checked a few WKT definitions, but [they don't specify lat, lng ordering; they just call it X and Y](https://www.ibm.com/docs/en/i/7.4?topic=formats-well-known-text-wkt-format) perhaps to avoid the lat, lng vs lng, lat holy war? But not sure which ordering we want to use here. I'll stick with lat, lng for now and if there's any reason to swap, let me know.

dfellis:
> So... I don't see any tests for the `h3` binary at all. Will like to discuss with you guys how we want to test it.

I finally found the tests near the bottom of the `CMakeTests.cmake` file, and then I updated them with the new WKT formatting and added the new test. No need to bring in more complicated machinery for these tests.

src/apps/filters/h3.c
isaacbrodsky: I believe it needs to be longitude, latitude, which corresponds to X, Y. I tried visualizing the output on e.g. https://wktmap.com/ and it does not look right if it is not longitude, latitude, so we would not be compatible with existing tooling.

isaacbrodsky (changes requested):
please change lat, lng ordering for compatibility with existing WKT implementations

src/apps/filters/h3.c
```diff
@@ -110,6 +116,41 @@ bool latLngToCellCmd(int argc, char *argv[]) {
...
+    // formatting
+    printf("POLYGON((");
+    for (int i = 0; i < cb.numVerts; i++) {
+        LatLng *ll = &cb.verts[i];
+        printf("%.10lf %.10lf, ", H3_EXPORT(radsToDegs)(ll->lng),
+               H3_EXPORT(radsToDegs)(ll->lat));
+    }
+    // WKT has the first and last points match, so re-print the first one
+    printf("%.10lf %.10lf))\n", H3_EXPORT(radsToDegs)(cb.verts[0].lng),
+           H3_EXPORT(radsToDegs)(cb.verts[0].lat));
```
isaacbrodsky, line 150: nit if this function gets reused, that it depends on numVerts > 0.

isaacbrodsky (approved):

src/apps/filters/h3.c
dfellis: Is it possible for it to be zero and not produce an H3Error value that short-circuits it, though?

src/apps/filters/h3.c
```diff
@@ -48,14 +48,19 @@ bool cellToLatLngCmd(int argc, char *argv[]) {
...
     LatLng ll;
-    H3_EXPORT(cellToLatLng)(cell, &ll);
-    printf("%.10lf, %.10lf\n", H3_EXPORT(radsToDegs)(ll.lat),
-           H3_EXPORT(radsToDegs)(ll.lng));
+    H3Error err = H3_EXPORT(cellToLatLng)(cell, &ll);
+    if (err) {
+        return false;
+    }
+    // Using WKT formatting for the output. TODO: Add support for JSON
+    // formatting
```
nrabinowitz, line 61: I'm curious about plans for output and input formats here. I assume for output we'd want a `-f` or `--format` param similar to `ogr2ogr`. It would be easy to go overboard here, but I think at a minimum:
- `csv`: `37.5012466151,-122.5003039349`
- `wkt`: `POINT(-122.5003039349 37.5012466151)`
- `geojson`: `{"type": "Point", "coordinates": [-122.5003039349 37.5012466151]}`

We might also support `json-array` (`[-122.5003039349 37.5012466151]`) or `json-object` (`{"lat":-122.5003039349, "lng": 37.5012466151}`) and maybe an `--ordering` param specifying `latlng` or `lnglat`?

The parallel question is what inputs we support - accepting these formats as inputs is nicely symmetrical, but more painful than `--lat=x --lng=y` or bare arguments.

Not saying we _should_ do any of this, but it might be nice to know where we plan to draw the line. I think a good yardstick should be how easy it is to pipe data from a tool like `ogr2ogr` to and from these CLI commands.

src/apps/filters/h3.c
```diff
@@ -48,14 +48,19 @@ bool cellToLatLngCmd(int argc, char *argv[]) {
...
                   "Convert an H3 cell to a latitude/longitude coordinate")) {
         return helpArg.found;
     }
     LatLng ll;
-    H3_EXPORT(cellToLatLng)(cell, &ll);
-    printf("%.10lf, %.10lf\n", H3_EXPORT(radsToDegs)(ll.lat),
-           H3_EXPORT(radsToDegs)(ll.lng));
+    H3Error err = H3_EXPORT(cellToLatLng)(cell, &ll);
+    if (err) {
+        return false;
```
nrabinowitz, line 58: For my info, why return `true`/`false` from these functions? Wouldn't we just want to return the error code, so callers would get a `0` exit code on success and a meaningful non-zero code on failure?

src/apps/filters/h3.c
```diff
@@ -102,6 +107,7 @@ bool latLngToCellCmd(int argc, char *argv[]) {
     H3Index c;
     H3Error e = H3_EXPORT(latLngToCell)(&ll, res, &c);
 
+    // TODO: Add support for JSON formatting
```
nrabinowitz, line 110: Does JSON formatting mean `"8928342e20fffff"` in this case? Also, does `h3Println` make piping harder than `h3Print`?

src/apps/filters/h3.c
```diff
@@ -110,6 +116,41 @@ bool latLngToCellCmd(int argc, char *argv[]) {
     return true;
 }
 
+bool cellToBoundaryCmd(int argc, char *argv[]) {
+    Arg cellToBoundaryArg = {
+        .names = {"cellToBoundary"},
+        .required = true,
+        .helpText =
+            "Convert an H3 cell to an array of latitude/longitude coordinates "
+            "defining its boundary",
```
nrabinowitz, line 125: Does this description still make sense if we're outputting WKT?

src/apps/filters/h3.c
```diff
@@ -48,14 +48,19 @@ bool cellToLatLngCmd(int argc, char *argv[]) {
     Arg helpArg = ARG_HELP;
     DEFINE_CELL_ARG(cell, cellArg);
     Arg *args[] = {&cellToLatLngArg, &helpArg, &cellArg};
-    if (parseArgs(argc, argv, 3, args, &helpArg,
+    if (parseArgs(argc, argv, sizeof(args) / sizeof(Arg *), args, &helpArg,
                   "Convert an H3 cell to a latitude/longitude coordinate")) {
```
nrabinowitz, line 52: Nit: Update description w/r/t output format?

src/apps/filters/h3.c
```diff
@@ -110,6 +116,41 @@ bool latLngToCellCmd(int argc, char *argv[]) {
...
+        .helpText =
+            "Convert an H3 cell to an array of latitude/longitude coordinates "
+            "defining its boundary",
+    };
+    Arg helpArg = ARG_HELP;
+    DEFINE_CELL_ARG(cell, cellArg);
+    Arg *args[] = {&cellToBoundaryArg, &helpArg, &cellArg};
+    if (parseArgs(argc, argv, sizeof(args) / sizeof(Arg *), args, &helpArg,
+                  "Convert an H3 cell to an array of latitude/longitude "
+                  "coordinates defining its boundary")) {
```
nrabinowitz, line 132: Nit: Is it possible to reuse `cellToBoundaryArg.helpText` here instead of repeating it?

src/apps/filters/h3.c
```diff
@@ -121,12 +162,21 @@ bool generalHelp(int argc, char *argv[]) {
         .helpText =
             "Convert degrees latitude/longitude coordinate to an H3 cell.",
     };
-    Arg *args[] = {&helpArg, &cellToLatLngArg, &latLngToCellArg};
+    Arg cellToBoundaryArg = {
+        .names = {"cellToBoundary"},
+        .helpText =
+            "Convert an H3 cell to an array of latitude/longitude coordinates "
+            "defining its boundary",
+    };
```
nrabinowitz, line 170: Nit: Is it possible to define this a top-level `static const` so that we can reuse it instead of repeating it?

src/apps/filters/h3.c
```diff
@@ -140,6 +190,9 @@ int main(int argc, char *argv[]) {
     if (has("latLngToCell", 1, argv) && latLngToCellCmd(argc, argv)) {
         return 0;
     }
+    if (has("cellToBoundary", 1, argv) && cellToBoundaryCmd(argc, argv)) {
+        return 0;
```
nrabinowitz, line 194: As above, I would have gone with `return cellToBoundaryCmd(argc, argv)` and have each command return `H3_ERROR`

nrabinowitz (approved):
This looks good - most of my comments are non-blocking questions about input/output and some of the existing structure of these commands

src/apps/filters/h3.c
isaacbrodsky: As far as I know it should always have vertices if the return code is success

src/apps/filters/h3.c
dfellis: So, I *think* we started work on this binary before `H3Error` was a thing and it was doing `h3IsValid` checks on the inputs/outputs to determine if it should exit with an error or not.

Then the conversion of that was done this way because it was the most straightforward. Bubbling up the `H3Error` *does* make sense to me. If you don't mind, I'll keep that refactor for when I add in the next sub-command.

src/apps/filters/h3.c
dfellis: That or `{"cell": "8928342e20fffff"}`. I'm not sure which would be a better idea.

src/apps/filters/h3.c
dfellis: Kinda? But I'll change it now to be more precise.

src/apps/filters/h3.c
dfellis: I really want to DRY this stuff up, but I think I need to rework or add some macros to do so in a way that maintains its legibility. I left it here for now because I found it easier to read the code this way, but I will do a bigger refactor in a follow-up PR that's focused just on DRYing the code so we can debate that without holding up progress on implementing a sub-command.

src/apps/filters/h3.c
dfellis: See my prior comment. Leaving DRY-ing it to another PR focused solely on that so we can better debate the merits/long-term maintenance.

src/apps/filters/h3.c
dfellis: I agree with you, but would like that to also be a separate PR.

Issue #406, opened 2020-11-04
Cannot  install in /usr/lib64 with cmake3/make


mrdvt92:
Fedora packaging guidelines place .so files /usr/lib64 but the h3 cmake/make do not appear to support LIB_SUFFIX=64.

This is what I have tried but nothing works to put .so file in /usr/lib64.

```
cmake3 -DCMAKE_INSTALL_PREFIX=/usr -DENABLE_DOCS=YES -DBUILD_SHARED_LIBS=true -DCMAKE_INSTALL_LIBDIR=/usr/lib64 -DLIBRARY_OUTPUT_PATH=/usr/lib64 ..
make %{?_smp_mflags}
make docs
make kml
make install DESTDIR=%{buildroot} LIB_SUFFIX=64 LIB_INSTALL_DIR=/usr/lib64
```
I have this kludge in my spec file to get around the issue but we really need a supported cmake/make option.

```
mv %{buildroot}/usr/lib %{buildroot}/usr/lib64
```

I forked to https://github.com/mrdvt92/h3 and the spec file builds these RPMs.

```
h3-3.7.1-4.el7.mrdvt92.src.rpm
h3-3.7.1-4.el7.mrdvt92.x86_64.rpm
h3-devel-3.7.1-4.el7.mrdvt92.x86_64.rpm
```

The Makefile does not have a `make dist` target so you will have to build a tar ball like this.

```bash
git clone https://github.com/mrdvt92/h3.git h3-3.7.1
tar cvfz h3-3.7.1.tar.gz h3-3.7.1/
rpmbuild -ta h3-3.7.1.tar.gz
```

isaacbrodsky:
I think this is because around line 47 in CMakeLists.txt the `LIBRARY_OUTPUT_PATH` variable is not set to be cached, so each time CMake will take the value in CMakeLists over the one you provide. If it's changed to `set(LIBRARY_OUTPUT_PATH lib CACHE STRING "Output library directory")` it might help. It may need some more changes around line 684 to use LIBRARY_OUTPUT_PATH there too.

mrdvt92:
This one line change worked for me. I do not know cmake and so I cannot state how portable this is or if it's even the correct way to handle this issue.

```
$ git diff CMakeLists.txt
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f6d32f8..1cdab52 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -681,7 +681,7 @@ install(
[...]

PR #819, opened 2024-02-26
Use CMAKE_INSTALL_LIBDIR

kou:
We always use "lib" for library directory but Debian/Ubuntu use "lib/x86_64-linux-gnu" and RHEL use "lib64" on 64bit.

CMake provides CMAKE_INSTALL_LIBDIR that abstracts system library directory. We can use it by including GNUInstallDirs. See also: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html


kou:
This will fix #406.

isaacbrodsky (approved):

PR #826, opened 2024-03-29
DRY the CLI

dfellis:
Based on the discussion with @nrabinowitz in the prior CLI PR. This PR removes duplicate help text and argument types, and turns the arg length trick (and all of the "show subcommand-specific help text" code) into a macro, deleting about twice as much text as it adds.

PS. DRY the CLI does *not* rhyme.

### DIFF
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -35,7 +35,7 @@
 #include "h3Index.h"
 #include "utility.h"
 
-#define PARSE()                                                             \
+#define PARSE_SUBCOMMAND(argc, argv, args)                                  \
     if (parseArgs(argc, argv, sizeof(args) / sizeof(Arg *), args, &helpArg, \
                   args[0]->helpText)) {                                     \
         return E_SUCCESS;                                                   \
@@ -62,7 +62,7 @@ Arg cellToBoundaryArg = {
 H3Error cellToLatLngCmd(int argc, char *argv[]) {
     DEFINE_CELL_ARG(cell, cellArg);
     Arg *args[] = {&cellToLatLngArg, &helpArg, &cellArg};
-    PARSE();
+    PARSE_SUBCOMMAND(argc, argv, args);
     LatLng ll;
     H3Error err = H3_EXPORT(cellToLatLng)(cell, &ll);
     if (err) {
@@ -100,7 +100,7 @@ H3Error latLngToCellCmd(int argc, char *argv[]) {
                   .helpText = "Longitude in degrees."};
 
     Arg *args[] = {&latLngToCellArg, &helpArg, &resArg, &latArg, &lngArg};
-    PARSE();
+    PARSE_SUBCOMMAND(argc, argv, args);
     LatLng ll = {.lat = H3_EXPORT(degsToRads)(lat),
                  .lng = H3_EXPORT(degsToRads)(lng)};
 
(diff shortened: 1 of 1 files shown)


src/apps/filters/h3.c
```diff
@@ -35,47 +35,51 @@
 #include "h3Index.h"
 #include "utility.h"
 
+#define PARSE()                                                             \
+    if (parseArgs(argc, argv, sizeof(args) / sizeof(Arg *), args, &helpArg, \
+                  args[0]->helpText)) {                                     \
+        return E_SUCCESS;                                                   \
+    }
```
nrabinowitz, line 42: Nit: Do we have a `.h` header file for this? Should we? If we did, should this macro go there?

src/apps/filters/h3.c
```diff
@@ -35,47 +35,51 @@
...
+    .helpText = "Convert an H3 cell to a WKT POINT coordinate",
+};
+Arg latLngToCellArg = {
+    .names = {"latLngToCell"},
+    .helpText = "Convert degrees latitude/longitude coordinate to an H3 cell.",
+};
+Arg cellToBoundaryArg = {
+    .names = {"cellToBoundary"},
+    .helpText = "Convert an H3 cell to a WKT POLYGON defining its boundary",
+};
```
nrabinowitz, line 60: Total nit: I'd put these next to the functions that implement them, but 🤷

src/apps/filters/h3.c
```diff
@@ -35,47 +35,51 @@
...
-    if (parseArgs(argc, argv, sizeof(args) / sizeof(Arg *), args, &helpArg,
-                  "Convert an H3 cell to a WKT POINT coordinate")) {
-        return helpArg.found;
-    }
+    PARSE();
     LatLng ll;
     H3Error err = H3_EXPORT(cellToLatLng)(cell, &ll);
     if (err) {
-        return false;
+        return err;
```
nrabinowitz, line 69: 👍

nrabinowitz (approved):
Looks great, thanks for the cleanup!

src/apps/filters/h3.c
dfellis: We kinda do, but it's shared across many of these binaries while this is only useful here.

On top of that, if I put it in there, I can't make the assumption about the naming of the various variables in use nor that the first argument in the array is declaring the sub-command name with the help text attached, so much of what makes it "magic" here would go away. It would just be a `parseArgs` function that does the `sizeof(args) / sizeof(Arg *)` trick for you, and not much else.

src/apps/filters/h3.c
dfellis: I waffled back and forth on that. As long as the `generalHelp` and `main` functions are the last in the file, that should work just fine, but it might be less obvious *why* they aren't defined within their function body, so I thought grouping them together as a collection of global variables would at least get people to pause before immediately wanting to refactor it and accidentally break the `generalHelp` function.

I was also thinking of turning the sub-command function declaration into a macro that makes these global variables at the same time it makes the sub-command function, but I don't know how "magic" we want the syntax in this file to be, so I didn't go that route.

src/apps/filters/h3.c
```diff
@@ -35,47 +35,51 @@
 #include "h3Index.h"
 #include "utility.h"
 
+#define PARSE()                                                             \
+    if (parseArgs(argc, argv, sizeof(args) / sizeof(Arg *), args, &helpArg, \
+                  args[0]->helpText)) {                                     \
```
isaacbrodsky, line 40: I'd prefer to pass in argc, argv, etc. as macro arguments to make it clear those variables are getting used when the macro is called.

I'd also suggest adding a comment here that explicitly goes over some preconditions of this macro (there should be at least one arg, the subcommand arg should be first, etc)

isaacbrodsky (approved):

ajfriend (approved):

isaacbrodsky (approved):

PR #836, opened 2024-05-03
Redesign the website homepage

isaacbrodsky:
Here are some work in progress screenshots:

Screenshot 2024-05-03 at 3 08 29 PM
Screenshot 2024-05-03 at 3 08 20 PM

Goals are for the h3geo.org home page to be more informative and useful.

- [ ] Paste in one or more hexagon IDs and see where they are.
- [ ] Paste in coordinates and see the hexagon ID
- [ ] "Where am I" in H3 (using the user location, if permission granted)
- [ ] More informative text - what is it, what can it be used for, what are facts someone needs to know about it

I adapted the code from [deck.gl](https://github.com/visgl/deck.gl/tree/master/website). There is still a lot of extra abstraction that can be cleaned up and removed here. We will also need to substitute appropriate basemap keys for this website.


ajfriend:
How easy would it be to let users click on the map to select and deselect hexes? And maybe in a way that would sync with the text box of given hexes?

isaacbrodsky:
> How easy would it be to let users click on the map to select and deselect hexes? And maybe in a way that would sync with the text box of given hexes?

The text box and the map can certainly be synchronized to each other. Right now we need to remove a lot of deck.gl's example boilerplate to make that state code easier to follow.

Selecting and deselecting hexes is possible but depends heavily on the exact logic for the selection interaction. Maybe we could use the approach @nrabinowitz mentioned to figure out what resolution of cell to select for a click?

isaacbrodsky:
@ajfriend can you look into the license/snyk check and let me know what's going on? I do not have access to that report.

website/src/components/example/doc-item-component.jsx
nrabinowitz, line 1: What's going on with the `/example` code? Doesn't seem to be H3-related.

website/src/examples/home-demo.js
```diff
@@ -0,0 +1,135 @@
...
+    getHexagon: (d) => d.hex,
+    extruded: false,
+    filled: false,
+    stroked: true,
+    getLineColor: [0, 0, 0],
+    getLineWidth: 3,
+    lineWidthMinPixels: 3,
+    highPrecision: true,
+  })] : [
+    new H3HexagonLayer({
```
nrabinowitz, line 48: Are you intending to implement support for finer resolutions (viewport-only)? Or is the idea to just show the first 3 resolutions? If it's the latter, this might actually be simpler and more performant just with Mapbox/Maplibre styles, rather than Deck...

ajfriend:
I got added to Synk, here's what it is complaining about:
Screenshot 2024-05-06 at 9 11 53 AM

I'm asking if we can have external accounts also added to Synk.

ajfriend:
Oh, its "snyk". Not "synk". :/

dfellis:
Now that it's building, here's some updated screenshots. :)

Screenshot from 2024-05-20 18-38-30
Screenshot from 2024-05-20 18-38-51

website/src/components/explorer/details.jsx
```diff
@@ -0,0 +1,187 @@
...
+  cellToLatLng,
+  h3IndexToSplitLong,
+  cellArea,
+  edgeLength,
+  originToDirectedEdges,
+  UNITS,
+} from "h3-js";
+
+function h3IndexToDigits(h) {
+  const THREE_BITS = 0x1 | 0x2 | 0x4;
```
dfellis, line 22: Super minor nit: Since JS is interpreted, these operations would be executed on each function call. Maybe just replace this with:

```suggestion
  const THREE_BITS = 7;
```

ajfriend:
Amazing! This is super exciting, and the screenshots look great!

I'll try building locally to test things out in the next few days.

Quick thoughts just based off of the screenshots: Of the information you're displaying in each cell, I'd actually consider the following to be fairly "advanced" H3 content that the average user probably wouldn't know what to make of:

- pentagon
- base cell
- icosa faces
- indexing digits

For a maximally-welcoming **front-page** display, I wonder if we'd want to limit it down to just the basics of what might be the most relevant to a new user:

- cell id
- resolution
- center (and should we specify lat/lng order?)

Maybe a separate view could provide more of the advanced cell stats, or just reordering or deprioritizing the advanced stuff? Also, all this info might be fine, this is just my first impression :)

isaacbrodsky:
Some notes from reviewing this live:
* It would be nice to just select a resolution for the user automatically
* It would be nice on mobile to show where the user is. We probably need to shrink the map & disable touch interactions so it doesn't mess up scrolling the rest of the page.
* It would be great to have more eyes on the text that's shown under the map.
* TBD if the details stay on the home page or which details are shown -- maybe we can rank what we want and choose a couple of them.

dfellis:
I have to note that I can't really test this as it currently is with the switch to mapbox, as it needs a mapbox token to load the tiles, and the dev console reports that it's not present.

Secondarily, I'm not sure if it's related to the tiles failing to load or not, but it can never resolve my location.

Screenshot from 2024-06-26 15-09-16

Somewhat unrelated, it seems the snyk integration has broken entirely (or they've blocked me from viewing it, as the details button even for the "Public Github" fails after I log in).

isaacbrodsky:
That shouldn't be related to the Mapbox failure. I believe it should run with a free Mapbox account but yes it is a limitation of using Mapbox.

ajfriend (approved):

website/src/components/explorer/details.jsx
```diff
@@ -0,0 +1,209 @@
...
+    );
+    const meanEdgeLength = (
+      edgeLengths.reduce((prev, curr) =>
+        prev !== undefined ? prev + curr : curr,
+      ) / edgeLengths.length
+    ).toPrecision(floatPrecision);
+    const parent = res !== 0 && cellToParent(hex, res - 1);
+    const children = res !== 15 && cellToChildren(hex, res + 1);
+    const neighbors = new Set(gridDisk(hex, 1));
+    neighbors.delete(hex);
```
nrabinowitz, line 140: Nit: Maybe `gridDisk(hex, 1).filter(n => n !== hex)` would be simpler, and then you don't need to spread it below

website/src/components/explorer/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+  str = str.replaceAll("}", " ");
+
+  return str;
+}
+
+function doSplitUserInput(userInput) {
+  if (userInput) {
+    // TODO: support WKT, GeoJSON inputs here too
+    const unwrapAnyArray = fullyUnwrap(userInput);
+    const split = unwrapAnyArray.split(/\s/).filter((str) => str !== "");
```
nrabinowitz, line 48: Nit: 
```suggestion
    const split = unwrapAnyArray.split(/\s+/);
```

website/src/components/explorer/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+    for (let i = 0; i < split.length; i++) {
+      const currentInput = fullyTrim(split[i]);
+      const nextInput = fullyTrim(split[i + 1]);
+
+      if (isValidCell(currentInput)) {
+        result.push(currentInput);
+      } else if (
+        i < split.length - 1 &&
+        Number.isFinite(Number.parseFloat(currentInput)) &&
+        Number.isFinite(Number.parseFloat(nextInput))
```
nrabinowitz, line 60: Oh huh, so pasting in lat/lng gives you the list of all cells? That's cool

website/src/components/explorer/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    return result;
+  }
+
+  return [];
+}
+
+function zoomToResolution(zoom) {
+  return Math.max(Math.min(zoom / 1.5, 15), 0);
```
nrabinowitz, line 82: Hacky but efficient 👍

website/src/components/explorer/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+}
+
+export default function HomeExporer({ children }) {
```
nrabinowitz, line 85:
```suggestion
export default function HomeExplorer({ children }) {
```

website/src/components/explorer/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+      return [...resAsSet][0];
+    } else {
+      return undefined;
+    }
+  }, [splitUserInput]);
+
+  const objectOnClick = useCallback(
+    ({ hex }) => {
+      const asSet = new Set(splitUserInput);
+      if (!asSet.delete(hex)) {
```
nrabinowitz, line 107: TIL, that's a nice pattern

website/src/components/explorer/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+          />
+          {splitUserInput.length ? (
+            <SelectedHexDetails
+              splitUserInput={splitUserInput}
+              setUserInput={setUserInput}
+              showNavigation={false}
+              showDetails={true}
+            />
+          ) : (
+            <></>
```
nrabinowitz, line 171: Nit: Prefer `null` rather than an empty fragment

website/src/components/explorer/map.jsx
```diff
@@ -0,0 +1,223 @@
...
+        for (const coord of coords) {
+          if (coord[0] < minX) {
+            minX = coord[0];
+          } else if (coord[0] > maxX) {
+            maxX = coord[0];
+          }
+          if (coord[1] < minY) {
+            minY = coord[1];
+          } else if (coord[1] > maxY) {
+            maxY = coord[1];
```
nrabinowitz, line 72: This approach might not work as expected for cells crossing the antimeridian I think. You might want to adjust each coord with 
```
function fixTransmeridianCoord(coord) {
  const lng = coord[0];
  coord[0] = lng < 0 ? lng + 360 : lng;
}
```

website/src/components/explorer/map.jsx
```diff
@@ -0,0 +1,223 @@
...
+          longitude,
+          zoom,
+          transitionInterpolator: new FlyToInterpolator(),
+          transitionDuration: 1600,
+        });
+      }
+    }
+  }, [userInput, userValidHex, deckLoaded]);
+
+  const layers = userValidHex
```
nrabinowitz, line 104: Hm. So we don't get any background grid when the user has selected some hexagons? Maybe that's correct, but sometimes it could be helpful

website/src/components/explorer/map.jsx
```diff
@@ -0,0 +1,223 @@
...
+          dashJustified: true,
+          dashGapPickable: true,
+          extensions: [new PathStyleExtension({ dash: true })],
+        }),
+      ];
+
+  const getTooltip = useCallback(({ object }) => {
+    if (object && object.hex) {
+      return {
+        html: `<tt>${object.hex}</tt>`,
```
nrabinowitz, line 174: Maybe
```suggestion
        html: `<code>${object.hex}</code>`,
```
? `<tt>` is deprecated

website/src/components/explorer/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+          )}
+        </BannerContainer>
+        <WhereAmIButton setUserInput={setUserInput} />
```
nrabinowitz, line 174: Cute, but actually useful?

website/src/pages/index.js
```diff
@@ -1,92 +1,78 @@
...
-                {features.map((props, idx) => (
-                  <Feature key={idx} {...props} />
-                ))}
-              </div>
-            </div>
-          </section>
-        )}
-      </main>
+    <Layout title="Home" description="H3">
+      <BrowserOnly>
```
nrabinowitz, line 46: TIL

website/src/pages/index.js
```diff
@@ -1,92 +1,78 @@
...
+        {() => {
+          const HomeExplorer = require("../components").HomeExplorer;
+          return <HomeExplorer />;
```
nrabinowitz, line 49: For my info, why is the dynamic import needed here? Do the 3rd-party libs access browser properties on import?

website/src/pages/index.js
```diff
@@ -1,92 +1,78 @@
...
+        }}
+      </BrowserOnly>
+      <div style={{ position: "relative" }}>
+        <TextContainer>
+          <h2>H3 indexes points and shapes into a hexagonal grid.</h2>
```
nrabinowitz, line 54: Nit: I'd say something about geospatial up front, e.g. `H3 is is a system for indexing geospatial data to a global hexagon-based grid` or similar

nrabinowitz (approved):
Lots of small comments, but looks really good!

website/src/components/explorer/details.jsx
```diff
@@ -0,0 +1,209 @@
...
+    const floatPrecision = res / 3 + 7;
+
+    const baseCell = getBaseCellNumber(hex);
+    const pent = isPentagon(hex);
+    const faces = getIcosahedronFaces(hex).join(", ");
+    const coords = cellToLatLng(hex)
+      .map((n) => n.toPrecision(floatPrecision))
+      .join(", ");
+    const digits =
+      res === 0 ? "(none)" : h3IndexToDigits(hex).slice(0, res).join("");
```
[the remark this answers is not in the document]
isaacbrodsky: I found that some of the detail lines got a bit log so I did not want to extend them

website/src/pages/index.js
isaacbrodsky: My mistake, the dynamic import does not seem to avtually be needed.

website/src/pages/index.js
isaacbrodsky: Let's iterate on this in Google Docs or Slack?

website/src/pages/index.js
isaacbrodsky: Actually DocuSaurus server side rendering will crash without it.

PR #837, opened 2024-05-04
Implement all of the inspection subcommands for the h3 cli program

dfellis:
I refactored the macros a bit more to reduce (but not completely eliminate, unfortunately) the redundancy in the code.

So, similar to what @nrabinowitz mentioned in the last PR, the definition of the `Arg` values for the subcommands is baked into the definition of the subcommands themselves. I was unable to find any way to generate an array of function pointers to these subcommands at the same time the functions were being defined, because you can't inline define them in C and I couldn't figure out any macro trickery to fake it.

Anyways, this should hopefully make binding the rest of the H3 functions much faster (which I think I demonstrated by binding [all of the inspection functions](https://h3geo.org/docs/api/inspection) excepting `maxFaceCount` since it's not really meant to be used directly).


dfellis:
This error doesn't make any sense to me: https://github.com/uber/h3/actions/runs/8947967184/job/24580540663?pr=837#step:4:27

It's saying that `H3Index` is an `unsigned long long` but when I try to make the suggested change locally, my compiler is telling me that `H3Index` is an `unsigned long` like I originally wrote.

grim7reaper:
> It's saying that `H3Index` is an `unsigned long long` but when I try to make the suggested change locally, my compiler is telling me that `H3Index` is an `unsigned long` like I originally wrote.

Could be because the system/toolchain used in GitHub CI map uint64_t to`unsigned long long` and your system map it onto `unsigned long`.
IIRC, that's why the C99 standard library provides `PRIu64` format specifier: to be portable.
Code should probably be:
```c
printf("%"PRIu64, c);
```
That will resolve to `lu` or `llu` according to what the system expects.

src/apps/filters/h3.c
```diff
@@ -41,25 +43,66 @@
...
-};
-Arg cellToBoundaryArg = {
-    .names = {"cellToBoundary"},
-    .helpText = "Convert an H3 cell to a WKT POLYGON defining its boundary",
-};
 
-H3Error cellToLatLngCmd(int argc, char *argv[]) {
+/// Indexing subcommands
+
+SUBCOMMAND(cellToLatLng, "Convert an H3Cell to a WKT POINT coordinate") {
```
ajfriend, line 105: Maybe it is too late to change at this point, but it seems unintuitive that the function ends with `LatLng` but it returns a WKT point, which is in Lng/Lat order.

Is it too late to call this function `cellToWKTPoint`? Or to change the output to be a lat/lng pair?

src/apps/filters/h3.c
```diff
@@ -75,7 +118,8 @@ H3Error cellToLatLngCmd(int argc, char *argv[]) {
     return E_SUCCESS;
 }
 
-H3Error latLngToCellCmd(int argc, char *argv[]) {
+SUBCOMMAND(latLngToCell,
+           "Convert degrees latitude/longitude coordinate to an H3 cell") {
```
ajfriend, line 122: I haven't looked at the CLI closely before, so apologies if this is a dumb question: I'm guessing there's no easy way in the CLI to get these functions to be inverses of each other (`latLngToCell` and `cellToLatLng`)?

ajfriend (approved):
Just some late-to-the-game questions, but other than that, this looks good to me!

src/apps/filters/h3.c
dfellis: @isaacbrodsky and I discussed this in the last PR. We're going to try and come up with a few "standard" input and output formats where applicable and have as much overlap between subcommands as possible.

But for this first pass of just getting an implementation actually done, since this is just happening when I have some spare time, it's whichever we think is the "most useful" standard, and Isaac asked for this to output WKT. I originally had it as a bare `lat, lng`.

So, I wouldn't rename them, as I'd rather the names match the h3geo.org documentation. Having it output in the ordering you mention will come. And I wouldn't mind if others wanted to chip in. ;)

src/apps/filters/h3.c
dfellis: I think a WKT input does make sense, as well, but it just hasn't been done, yet. I'm thinking a `-w / --wkt` arg would probably fit.

dfellis: I think the reason why we didn't do that at first is because it could be confusing what to do if the WKT text is not just a `POINT`. Should it automatically compute cells for each point provided if it's passed a `LINESTRING`? Making decisions on the "right" behavior in situations like that was basically punted for later.

src/apps/filters/h3.c
```diff
@@ -140,35 +185,177 @@ H3Error cellToBoundaryCmd(int argc, char *argv[]) {
...
+
+SUBCOMMAND(isResClassIII,
+           "Checks if the provided H3 index has a Class III orientation") {
+    DEFINE_CELL_ARG(cell, cellArg);
+    Arg *args[] = {&isResClassIIIArg, &helpArg, &cellArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    bool isClassIII = H3_EXPORT(isResClassIII)(cell);
+    printf("%s", isClassIII ? "true" : "false");
+    return E_SUCCESS;
+}
```
isaacbrodsky, line 283: it's surprising that getResolution does a lot of extra checks but isResClassIII doesn't

src/apps/filters/h3.c
```diff
@@ -140,35 +185,177 @@ H3Error cellToBoundaryCmd(int argc, char *argv[]) {
...
+    Arg *args[] = {&getResolutionArg, &helpArg, &cellArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    // TODO: Should there be a general `isValidIndex`?
+    H3Error cellErr = H3_EXPORT(isValidCell)(cell);
+    H3Error edgeErr = H3_EXPORT(isValidDirectedEdge)(cell);
+    H3Error vertErr = H3_EXPORT(isValidVertex)(cell);
+    if (cellErr && edgeErr && vertErr) {
+        // Not exactly sure how to handle this
+        return cellErr | edgeErr | vertErr;
+    }
```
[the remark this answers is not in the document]
dfellis: No, because if it's a valid edge but not a valid vertex, then we still want to process the data because the index is an edge index and not a vertex index. So only if it fails *all* index types are we sure it's not a valid index.

src/apps/filters/h3.c
```diff
@@ -41,25 +43,66 @@
...
+struct Subcommand {
+    char *name;
+    Arg *arg;
+    H3Error (*subcommand)(int, char **);
+};
+
+#define SUBCOMMANDS_INDEX                        \
+    H3Error generalHelp(int argc, char *argv[]); \
+    struct Subcommand subcommands[] = {
+#define SUBCMD(s) {.name = #s, .arg = &s##Arg, .subcommand = &s##Cmd},
```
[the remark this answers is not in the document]
dfellis: I am totally open to alternate naming here.

src/apps/filters/h3.c
dfellis: Yeah, I will update that.

src/apps/filters/h3.c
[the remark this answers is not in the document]
isaacbrodsky: I see, in that case I think the `return cellErr || edgeErr || vertErr;` is still right.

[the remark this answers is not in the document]
dfellis: Well, that would always short-circuit to just `cellErr` in that case, since they'd all be non-zero, so why not just `return cellErr` in that case?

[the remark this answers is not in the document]
isaacbrodsky: Ok, fair enough. As long as a single value is returned I think it's ok.

src/apps/filters/h3.c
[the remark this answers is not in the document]
isaacbrodsky: `SUBCOMMAND` to `DEFINE_SUBCOMMAND`, `SUBCMD` to `SUBCOMMAND_INDEX`?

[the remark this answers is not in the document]
dfellis: I'm alright with the former, but I'd rather not on the latter, if that's okay.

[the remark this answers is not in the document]
isaacbrodsky: That sounds good, the second one I was stretching for how to add a word to it

CMakeTests.cmake
```diff
@@ -231,6 +231,15 @@ add_h3_test(testCellToBBoxExhaustive src/apps/testapps/testCellToBBoxExhaustive.
...
 add_h3_cli_test(testCliCellToBoundary "cellToBoundary -c 8928342e20fffff" "POLYGON((-122.4990471431 37.4997389893, -122.4979805011 37.5014245698, -122.4992373065 37.5029321860, -122.5015607527 37.5027541980, -122.5026273256 37.5010686174, -122.5013705214 37.4995610248, -122.4990471431 37.4997389893))")
+add_h3_cli_test(testCliGetResolution "getResolution -c 85283473fffffff" "5")
+add_h3_cli_test(testCliGetBaseCellNumber "getBaseCellNumber -c 85283473fffffff" "20")
+add_h3_cli_test(testCliStringToInt "stringToInt -c 85283473fffffff" "599686042433355775")
+add_h3_cli_test(testCliIntToString "intToString -c 599686042433355775" "85283473fffffff")
+add_h3_cli_test(testCliIsValidCell "isValidCell -c 85283473fffffff" "true")
+add_h3_cli_test(testCliIsNotValidCell "isValidCell -c 85283473ffff" "false")
+add_h3_cli_test(testCliIsResClassIII "isResClassIII -c 85283473fffffff" "true")
+add_h3_cli_test(testCliIsPentagon "isPentagon -c 85283473fffffff" "false")
+add_h3_cli_test(testCliGetIcosahedronFaces "getIcosahedronFaces -c 81743ffffffffff" "3, 8, 13, 9, 4")
```
nrabinowitz, line 242: I'm sure we talked about this before, but it seems weird to me to inline the actual test logic into the CMake file, particularly alongside other CMake commands doing more... make-y stuff. If this is really the best way to do it, can we separate out these commands into a different `CMakeCLITests.cmake` file?

src/apps/filters/h3.c
nrabinowitz: I remember the previous discussion, but I guess I'd really like a `--wkt` or `--format=wkt` flag here, just because otherwise we have an API we can't change in a backwards-compatible way...

src/apps/filters/h3.c
```diff
@@ -117,7 +161,8 @@ H3Error latLngToCellCmd(int argc, char *argv[]) {
     return e;
 }
 
-H3Error cellToBoundaryCmd(int argc, char *argv[]) {
+SUBCOMMAND(cellToBoundary,
+           "Convert an H3 cell to a WKT POLYGON defining its boundary") {
```
nrabinowitz, line 165: Same, maybe require a flag even for the default output format to give us more room to maneuver in the future? Plus it makes the output format really explicit...

src/apps/filters/h3.c
```diff
@@ -140,35 +185,195 @@ H3Error cellToBoundaryCmd(int argc, char *argv[]) {
...
+           "Extracts the base cell number (0 - 121) from the H3 cell") {
+    DEFINE_CELL_ARG(cell, cellArg);
+    Arg *args[] = {&getBaseCellNumberArg, &helpArg, &cellArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    // TODO: Should there be a general `isValidIndex`?
+    H3Error cellErr = H3_EXPORT(isValidCell)(cell);
+    H3Error edgeErr = H3_EXPORT(isValidDirectedEdge)(cell);
+    H3Error vertErr = H3_EXPORT(isValidVertex)(cell);
+    if (cellErr && edgeErr && vertErr) {
+        return cellErr;
```
nrabinowitz, line 219: I don't object, but for my info, is the principle here that we should always validate CLI input because we assume that correctness is more important than perf in this context?

nrabinowitz (approved):

src/apps/filters/h3.c
dfellis: That was my reasoning on this. You have a whole process spawn and arg parse for each call, so the validation logic should be a rounding error here.

CMakeTests.cmake
dfellis: I don't remember who introduced CLI testing in this way, actually. I am pretty sure that I *didn't*, but I might be wrong about it.

In any case, I was following the existing convention and there was no objection before on this, but I agree it feels weird.

Do you mind if I take a stab at restructuring these tests during the next PR?

nrabinowitz:
Sure, sgtm

isaacbrodsky (approved):

PR #845, opened 2024-06-04
Fix cellToChildPos and childPosToCell docs

dfellis:
I noticed while going through the H3 documentation working on the cli that the live examples for the `cellToChildPos` and `childPosToCell` don't work because the website is using an out of date version of h3-js. This small PR fixes that (confirmed locally).

### DIFF
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -1546,7 +1546,7 @@
     "@docusaurus/theme-search-algolia" "2.0.1"
     "@docusaurus/types" "2.0.1"
 
-"@docusaurus/react-loadable@5.5.2", "react-loadable@npm:@docusaurus/react-loadable@5.5.2":
+"@docusaurus/react-loadable@5.5.2":
   version "5.5.2"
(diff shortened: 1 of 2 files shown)


dfellis:
What's currently on the website:

Screenshot from 2024-06-04 17-27-40

After the fix:

Screenshot from 2024-06-04 17-30-24

isaacbrodsky (approved):

website/yarn.lock
```diff
@@ -6263,6 +6263,14 @@ react-loadable-ssr-addon-v5-slorber@^1.0.1:
   dependencies:
     "@babel/runtime" "^7.10.3"
 
+"react-loadable@npm:@docusaurus/react-loadable@5.5.2":
```
nrabinowitz, line 6266: Seems a little odd that the lockfile would change here (different yarn version maybe?) but NBD

nrabinowitz (approved):
Thanks for the fix!

PR #846, opened 2024-06-14
H3 CLI Hierarchical Subcommands

dfellis:
This adds all of the hierarchical subcommands. Everything was easy until I reached `compactCells` and `uncompactCells`.

I am *not* happy with the way the flexible parser ended up in a soup of looping constructs, but I couldn't figure out any way that wasn't worse in some ways (eg, using `goto` and labels). I also couldn't figure out how handle a completely undelimited input file without copying the characters to a temporary string since you can have more than 15 hex characters in a 64-bit integer, so it would want to slurp up follow-on characters. We *can* simplify a bit if you think an undelimited deluge of hex characters is not a reasonable input format we should handle.

I also didn't DRY the huge similarities between the two functions. I am willing to do that, but only after we agree on how the string parsing logic should look like.


dfellis:
I think [*this*](https://github.com/uber/h3/actions/runs/9517829423/job/26237292550?pr=846#step:3:2) may be why the tests are failing. It's setting up a build directory that is not the CWD, while the files are using relative paths. I need to see how to get the directory they're actually being put in vs where `make test` is getting invoked.

dfellis:
Hmm... There was another error that I didn't notice before. https://github.com/uber/h3/actions/runs/9521371842/job/26248637576?pr=846#step:7:301

Let me see if I can figure that out.

dfellis:
So that last failure was legit, and was a copy-paste error when I set up arg parsing in one of the new subcommands.

I have no idea what black magic the GNU libc is doing under the hood to get that casting to work "correctly" on Linux.

src/apps/filters/h3.c
```diff
@@ -727,6 +727,557 @@ SUBCOMMAND(localIjToCell,
...
+    DEFINE_CELL_ARG(cell, cellArg);
+    int res = 0;
+    Arg resArg = {.names = {"-r", "--resolution"},
+                  .required = true,
+                  .scanFormat = "%d",
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, 1-15 inclusive, excluding 0 as it can "
+                      "never be a child"};
```
isaacbrodsky, line 774: could cellToChildren be called with res 0 if the input cell is res 0? It would only return the input cell.

src/apps/filters/h3.c
```diff
@@ -727,6 +727,557 @@ SUBCOMMAND(localIjToCell,
...
+    // path just ends at that point. As we have no idea how many cells we're
+    // going to load, we allocate enough for 128 cells, but if that's not
+    // enough, we also have to re-allocate double the number of cells, copy them
+    // over, and free the old buffer of cells. Doing this manually since we want
+    // to keep the build process for H3 simple and C's stdlib is pretty bare.
+    H3Index *cells = calloc(128, sizeof(H3Index));
+    FILE *fp = 0;
+    int cellStrsOffset = 0;
+    int cellsOffset = 0;
+    int cellsLen = 128;
```
isaacbrodsky, line 977: `size_t`?

tests/inputfiles/compact_test3.txt
```diff
@@ -0,0 +1 @@
+85283473fffffff85283447fffffff8528347bfffffff85283463fffffff85283477fffffff8528340ffffffff8528340bfffffff85283457fffffff85283443fffffff8528344ffffffff852836b7fffffff8528346bfffffff8528346ffffffff85283467fffffff8528342bfffffff8528343bfffffff85283407fffffff85283403fffffff8528341bfffffff
```
isaacbrodsky, line 1: Do we really want to support this kind of fixed length record format? This seems like the worst of human readability from binary files (hard to know if the records are the right size, as opposed to e.g. JSON or CSV where you can eyeball and see where the record separators are) and parsing performance of text files. (can't just directly reference this as a buffer of `(u)int64`, must parse to integers)

dfellis: I simplified it a bit (and moved it to a singular function), but I couldn't reduce the complexity as much as I wanted to because `sscanf` returns the *number* of matched variables, not the number of matched characters in the string, so I can't advance the string in the obvious way.

I am half-tempted to rewrite it without any `sscanf` call and build the integer manually.

src/apps/filters/h3.c
```diff
@@ -727,6 +727,468 @@ SUBCOMMAND(localIjToCell,
...
+    }
+    H3Error err = H3_EXPORT(childPosToCell)(pos, cell, res, &child);
+    if (err) {
+        return err;
+    }
+    h3Println(child);
+    return E_SUCCESS;
+}
+
+H3Index *readCellsFromFile(FILE *fp, char *buffer, int *totalCells) {
```
isaacbrodsky, line 932:
```suggestion
H3Index *readCellsFromFile(FILE *fp, char *buffer, size_t *totalCells) {
```
otherwise may run into an issue after a few gigabytes

isaacbrodsky (approved):

ajfriend (approved):

dfellis:
Just rebased to get the new tests run on this branch. Since I have two approvals if these new tests also pass I'll merge this.

PR #888, opened 2024-07-14
Update hierarchy.mdx

SabNK:
changed method name to the current version 4.1.1


isaacbrodsky (approved):
Unless I'm forgetting something, we should be able to name this function `cellToParent` in H3-Java.

SabNK:
> Unless I'm forgetting something, we should be able to name this function `cellToParent` in H3-Java.

This sounds reasonable, do you want me to claim deprecated method and make a new pr?

PR #897, opened 2024-08-11
Website: Enable zoom in/out on mobile

isaacbrodsky:
Two fixes in this PR:

* Try to enable zoom in/out on mobile so the map is more useful there.
* When zooming in on user specified hexagons, set their size in terms of pixels so very small hexagons (e.g. resolution 13+) do not seem to have very thick outlines
* 
Screenshot 2024-08-11 at 1 45 14 PM


website/src/components/explorer/map.jsx
```diff
@@ -205,11 +219,14 @@ export function ExplorerMap({
...
       initialViewState={currentInitialViewState}
-      controller={true}
       getTooltip={getTooltip}
       getCursor={getCursor}
       onClick={onClick}
       onLoad={() => setDeckLoaded(true)}
+      controller={{
+        dragPan: windowWidth && windowWidth >= MOBILE_CUTOFF_WINDOW_WIDTH,
+      }}
+      touchAction="pan-y"
```
dfellis, line 229: What do these two properties alter, specifically?

dfellis (approved):

website/src/components/explorer/map.jsx
isaacbrodsky: They're suggested by the deck.gl docs: https://deck.gl/docs/api-reference/core/deck#touchaction

* Disable drag to pan when the window width is too low. Single finger touch interactions should not trigger moving the map.
* I think what touchAction does is allow that touch interaction to go to the page rather than get consumed by deck.gl.

ajfriend (approved):

PR #912, opened 2024-09-26
Compute constants at start of program

heshpdx:
Instead of using `#define`d constants, we can use `static const` variables that lets the system compute what the values would be. This puts to bed any debate around accuracy and rounding and if we added enough significant figures into the definition. What is one-third anyway? Just let the system decide...
```
(gdb) p (double) 1/3
$1 = 0.33333333333333331
```

Ideally we would want to use `constexpr` instead of `const`, because `constexpr` is evaluated at compile time. But `constexpr` is only available in C++. `const` is evaluated at the beginning of runtime. What I don't know is how often does a program start up in the field? If it is very frequent then maybe this is not a good idea.

### DIFF
--- a/src/h3lib/include/constants.h
+++ b/src/h3lib/include/constants.h
@@ -19,58 +19,59 @@
 
 #ifndef CONSTANTS_H
 #define CONSTANTS_H
+#include <math.h>
 
 #ifndef M_PI
 /** pi */
-#define M_PI 3.14159265358979323846
+static const double M_PI = 3.14159265358979323846;
 #endif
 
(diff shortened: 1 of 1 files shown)


dfellis (changes requested):
I don't understand the point of this. Switching to computing on-the-fly doesn't improve the maintainability of the code in any way and has the negative of needing to recompute these values at runtime.

We already have comments on what computation derived the constants, so I don't think it's a big problem that they're "magic" numbers.

heshpdx:
Understood! I will close this PR.

PR #914, opened 2024-09-26
Test for compacting all res 1 cells

ajfriend:
- `compactCells` currently fails when trying to compact all res 1 cells.
- note that, interestingly, it also fails if you try to compact just the first 41 res 1 cells.

### DIFF
--- a/src/apps/applib/include/args.h
+++ b/src/apps/applib/include/args.h
@@ -90,7 +90,7 @@ int _parseArgsList(int argc, char *argv[], int numArgs, Arg *args[],
 // common arguments
 
 #define ARG_HELP \
-    { .names = {"-h", "--help"}, .helpText = "Show this help message." }
+    {.names = {"-h", "--help"}, .helpText = "Show this help message."}
 #define DEFINE_INDEX_ARG(varName, argName) \
     H3Index varName = 0;                   \
     Arg argName = {                        \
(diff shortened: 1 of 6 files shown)


ajfriend:
I'm not sure why it did all this formatting... Any ideas?

isaacbrodsky:
> I'm not sure why it did all this formatting... Any ideas?

Probably different version of clang-format being used?

ajfriend:
> Probably different version of clang-format being used?

Do we have a standard version I should be using?

isaacbrodsky:
> > Probably different version of clang-format being used?
> 
> Do we have a standard version I should be using?

Yes, it is important to use clang-format-14 (until we update the formatter version again) as they change defaults between versions.

ajfriend:
> Yes, it is important to use clang-format-14 (until we update the formatter version again) as they change defaults between versions.

OK. For my future reference, I had to run

```
brew install llvm@14
ln -s "$(brew --prefix llvm@14)/bin/clang-format" /opt/homebrew/bin/clang-format
```

which feels a bit janky.

ajfriend:
I can also make it fail on a different (disjoint from the first example) set of cells with something like

```c
int64_t numUncompacted = 44;
t_assertSuccess(
    H3_EXPORT(compactCells)(&cells1[80], out, numUncompacted));
```

isaacbrodsky:
The problem is around line https://github.com/uber/h3/blob/b355c9d825428b8f08636ffa7250422f52c2f6e0/src/h3lib/lib/h3Index.c#L490 , the `parentRes` is -1 but we try to take `cellToParent` of `currIndex` at `parentRes` anyways.

PR #916, opened 2024-09-26
Test `cellsToLinkedMultiPolygon` for children of pentagon

ajfriend:
Adds a test that catches a current bug when computing the `cellsToLinkedMultiPolygon` of a set of children of a pentagon cell.

### DIFF
--- a/src/apps/testapps/testCellsToLinkedMultiPolygon.c
+++ b/src/apps/testapps/testCellsToLinkedMultiPolygon.c
@@ -72,6 +72,24 @@ SUITE(cellsToLinkedMultiPolygon) {
         H3_EXPORT(destroyLinkedMultiPolygon)(&polygon);
     }
 
+    TEST(pentagonChldren) {
+        // children of pentagon 0x80ebfffffffffff
+        H3Index kids[] = {0x81ea3ffffffffff, 0x81eabffffffffff,
+                          0x81eafffffffffff, 0x81eb3ffffffffff,
+                          0x81eb7ffffffffff, 0x81ebbffffffffff};
+        int numCells = ARRAY_SIZE(set);
+
+        LinkedGeoPolygon polygon;
+        t_assertSuccess(
+            H3_EXPORT(cellsToLinkedMultiPolygon)(kids, numCells, &polygon));
+
+        // Since these are the children of a cell, we exepect a single loop with
+        // no holes.
+        t_assert(countLinkedLoops(&polygon) == 2, "1 loop added to polygon");
+
+        H3_EXPORT(destroyLinkedMultiPolygon)(&polygon);
+    }
+
     // TODO: This test asserts incorrect behavior - we should be creating
     // multiple polygons, each with their own single loop. Update when the
     // algorithm is corrected.


src/apps/testapps/testCellsToLinkedMultiPolygon.c
```diff
@@ -72,6 +72,24 @@ SUITE(cellsToLinkedMultiPolygon) {
         H3_EXPORT(destroyLinkedMultiPolygon)(&polygon);
     }
 
+    TEST(pentagonChldren) {
```
isaacbrodsky, line 75:
```suggestion
    TEST(pentagonChildren) {
```

src/apps/testapps/testCellsToLinkedMultiPolygon.c
```diff
@@ -72,6 +72,24 @@ SUITE(cellsToLinkedMultiPolygon) {
...
+                          0x81eafffffffffff, 0x81eb3ffffffffff,
+                          0x81eb7ffffffffff, 0x81ebbffffffffff};
+        int numCells = ARRAY_SIZE(kids);
+
+        LinkedGeoPolygon polygon;
+        t_assertSuccess(
+            H3_EXPORT(cellsToLinkedMultiPolygon)(kids, numCells, &polygon));
+
+        // Since these are the children of a cell, we exepect a single loop with
+        // no holes.
```
isaacbrodsky, line 87:
```suggestion
        // Since these are the children of a cell, we expect a single loop with
        // no holes.
```

src/apps/testapps/testCellsToLinkedMultiPolygon.c
```diff
@@ -72,6 +72,24 @@ SUITE(cellsToLinkedMultiPolygon) {
         H3_EXPORT(destroyLinkedMultiPolygon)(&polygon);
     }
 
+    TEST(pentagonChildren) {
+        // children of pentagon 0x80ebfffffffffff
+        H3Index kids[] = {0x81ea3ffffffffff, 0x81eabffffffffff,
+                          0x81eafffffffffff, 0x81eb3ffffffffff,
+                          0x81eb7ffffffffff, 0x81ebbffffffffff};
+        int numCells = ARRAY_SIZE(kids);
```
isaacbrodsky, line 80: does this happen for any set of res 1 children of a res 0 pentagon?

ajfriend: At res 2, exactly two pentagons of the 12 do not work: `8009fffffffffff` and `80c3fffffffffff`

heshpdx:
In case it helps, I ran `make test` with this branch on two aarch64 Ampere machines running linux, and everything passed including `testCellsToLinkedMultiPolygon_test95`. I tried on Altra and AmpereOne.

isaacbrodsky:
I suspect this has to do with https://github.com/uber/h3/blob/b355c9d825428b8f08636ffa7250422f52c2f6e0/src/h3lib/lib/vertexGraph.c#L76

For example, changing it from
```
    return (uint32_t)fmod(fabs((vertex->lat + vertex->lng) * pow(10, 15 - res)),
                          numBuckets);
```

to
```
    return (uint32_t)fmod(fabs((vertex->lat + vertex->lng) * pow(10, 16 - res)),
                          numBuckets);
```

seems to resolve this particular error case. I think we had considered using vertex indexes for this algorithm before, but hadn't done so. If I recall right the performance wasn't much different (wasn't better anyways). I think @nrabinowitz did those tests so I'll let him chime in. Do we still have that on a branch we could resurrect?

nrabinowitz:
> I think we had considered using vertex indexes for this algorithm before, but hadn't done so. If I recall right the performance wasn't much different (wasn't better anyways). I think @nrabinowitz did those tests so I'll let him chime in. Do we still have that on a branch we could resurrect?

I'm... not sure, I'll need to check. It is possible my WIP branch (now several years old) went away with an old work laptop :/.

nrabinowitz:
> > I think we had considered using vertex indexes for this algorithm before, but hadn't done so. If I recall right the performance wasn't much different (wasn't better anyways). I think @nrabinowitz did those tests so I'll let him chime in. Do we still have that on a branch we could resurrect?
> 
> I'm... not sure, I'll need to check. It is possible my WIP branch (now several years old) went away with an old work laptop :/.

Does not seem to be in my remote GH branches. I think it's unlikely I can access the code any more, sorry about that.

grim7reaper:
Not sure if it could help to implement it in C, but the h3o implementation is based on Vertex indexes.

IIRC, it wasn't too complicated to adapt the algorithm, the only tricky part being the "distortions" vertices that needs to be added and cannot be represented by the vertex indexes, which only encode topological vertices. I got sth working at the end of the day, but not very satisfied of the approach and there may be room for improvements.

Issue #917, opened 2024-09-27
`cellsToLinkedMultiPolygon` bug on Apple Silicon


ajfriend:
We have a bug computing `cellsToLinkedMultiPolygon` for the children of a pentagon which only seems to show up on Apple Silicon builds. I'm guessing this is a floating point issue?

Demonstrated in #916 .

ajfriend:
Only on Apple Silicon, but not on x86: https://github.com/uber/h3/actions/runs/11062819778/job/30737940554?pr=916

<img width="313" alt="image" src="https://github.com/user-attachments/assets/b1124746-6e41-41a2-996a-baabf354bc57">

(If I'm interpreting this correctly: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

dylanpulver:
Your guess was right — it is floating point, and specifically it is **FMA contraction**, which is why it split by architecture rather than by OS.

Measured on an Apple M5 Max (arm64), at `1d5346bb` (the commit before the fix), for the six children of `80ebfffffffffff`. Two cells compute the same shared vertex and disagree:

```
81ea3ffffffffff vert5 lat = -1.1775462077297509378
81eafffffffffff vert6 lat = -1.1775462077297504937
delta                     = 4.44e-16 rad  ->  2.8 nanometres on the ground
geoAlmostEqual            = TRUE  (threshold 1.745e-11 rad, ~0.11 mm)
```

`geoAlmostEqual` is happy — the tolerance is about 40 million times larger than the error. The problem is that it never gets consulted, because `_hashVertex` picks the bucket first and it is an **exact-value** hash sitting in front of an epsilon comparison:

```c
(uint32_t)fmod(fabs((lat + lng) * pow(10, 15 - res)), numBuckets)
```

At res 1 that multiplies by 1e14, so the 4.44e-16 becomes 0.0625, and that straddles an integer boundary:

```
(lat+lng)*1e14   -417723903726126        vs  -417723903726125.9375
fmod(|.|, 6)      0                      vs   5.9375
bucket            0                      vs   5
```

`findNodeForEdge` searches the wrong bucket, the reverse edge is never found, so the shared edge is not cancelled and the ring is left open — hence the second loop.

The architecture split is FMA. Same commit, same machine, only the flag changes:

```
default (-ffp-contract=on, arm64 emits fmadd)  ->  2 loops   (bug)
[...]

PR #918, opened 2024-09-27
Test on x86 and apple silicon

ajfriend:
We've seen a bug (https://github.com/uber/h3/issues/917) that appears on Apple Silicon but not x86, so adding to the build matrix to test across both.

Also, I shorten the job name so we can see the OS version in the github actions web UI.

Note, I don't think there's a way to guarantee architecture, but it seems like it is usually correlated with macos version:
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/test-macos.yml
+++ b/.github/workflows/test-macos.yml
@@ -8,11 +8,12 @@ on:
 
 jobs:
   tests:
-    name: Test Compile ${{ matrix.build_type }}
-    runs-on: macos-latest
+    name: ${{ matrix.build_type }} on ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


dfellis (approved):

ajfriend:
image

If everyone agrees, I can make the same name change on the other tests.

isaacbrodsky (approved):

isaacbrodsky:
I've changed the branch protection rules to not require the previous test names and will update them to require the new ones right after merging

PR #919, opened 2024-09-29
Possible fix for #914 compactCells failing when passed all res 1 cells

isaacbrodsky:
Incorporates #914

### DIFF
--- a/src/apps/testapps/testCompactCells.c
+++ b/src/apps/testapps/testCompactCells.c
@@ -386,9 +386,7 @@ SUITE(compactCells) {
                          0x7,
                          0x400000000};
         H3Index output[43] = {0};
-        t_assert(H3_EXPORT(compactCells)(bad, output, numHex) == E_RES_DOMAIN,
-                 "compactCells returns E_RES_DOMAIN on bad input (parent "
-                 "error #2)");
+        t_assertSuccess(H3_EXPORT(compactCells)(bad, output, numHex));
     }
 
     TEST(uncompactCells_wrongRes) {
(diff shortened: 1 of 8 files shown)


nrabinowitz (approved):
Fix looks correct to me

dfellis (approved):

src/apps/testapps/testCompactCells.c
```diff
@@ -91,6 +92,44 @@ SUITE(compactCells) {
...
+            H3_EXPORT(uncompactCells)(cells0, numRes0, cells1, numRes1, 1));
+
+        // Fails at compactCells.
+        // However:
+        //   Passes if numUncompacted <= 40
+        //   Fails  if numUncompacted >= 41.
+        int64_t numUncompacted = numRes1;
+        t_assertSuccess(H3_EXPORT(compactCells)(cells1, out, numUncompacted));
+
+        // TODO: check that output matches cells0
```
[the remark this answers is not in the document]
isaacbrodsky: Sure, I'll add that too

ajfriend (approved):

src/apps/testapps/testCompactCells.c
```diff
@@ -91,6 +92,85 @@ SUITE(compactCells) {
...
+
+                t_assertSuccess(H3_EXPORT(compactCells)(&cells1[offset], out,
+                                                        numUncompacted));
+            }
+        }
+
+        free(cells0);
+        free(cells1);
+        free(out);
+    }
```
ajfriend, line 172: Actually, I'm not positive that we need this test. I had code like this previously just to try to generate a minimal example of a failure. I don't think it hurts to add this test, but I don't think there's any reason to think this provides much additional value beyond compacting all the res 1 cells. Happy to go either way with it, based on your judgement @isaacbrodsky .

isaacbrodsky: It's not that strong a test. More convincing might be enumerating all possible sets of res 1 cells, but it might be useful for just ensuring all of these cases are not inadvertently failing.

PR #921, opened 2024-09-30
Docstring updates

isaacbrodsky:
This updates the doc strings in a few places to be consistent with the actual parameters. I also changed Doxygen settings to improve that a little, e.g. to include the public API header.

Based on #919, #920.


isaacbrodsky:
This should be ready for review now.

src/h3lib/lib/directedEdge.c
```diff
@@ -127,7 +127,7 @@ H3Error H3_EXPORT(areNeighborCells)(H3Index origin, H3Index destination,
  * destination
  * @param origin The origin H3 hexagon index
  * @param destination The destination H3 hexagon index
- * @return The directed edge H3Index, or H3_NULL on failure.
+ * @param out Output: The directed edge H3Index.
```
dfellis, line 130: Should we include the `@return` still for the `H3Error` path?

dfellis: Eg, just copy this from further below:

```suggestion
 * @param out Output: The directed edge H3Index.
 * @returns E_SUCCESS on success, or another value on error
```

isaacbrodsky: I was most focused on ensuring the docstring wasn't incorrect, rather than completing it. That seems like a good addition.

ajfriend (approved):

PR #923, opened 2024-10-07
H3 CLI Region Subcommands

dfellis:
Still as a draft because I've only hand tested some simple cases, and I know there's an issue with reading large polygon definitions from a file, but I wanted to get what I have up sooner rather than later so you can start taking a look at it.

### DIFF
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -1189,6 +1189,375 @@ SUBCOMMAND(uncompactCells,
     return E_SUCCESS;
 }
 
+/// Region subcommands
+
+H3Error polygonStringToGeoPolygon(FILE *fp, char *polygonString,
+                                  GeoPolygon *out) {
+    // There are two kinds of valid input, an array of arrays of lat, lng values
+    // defining a singular polygon loop, or an array of array of arrays of lat,
+    // lng values defining a polygon and zero or more holes. Which kind of input
+    // this is can be determined by the `[` depth reached before the first
+    // floating point number is found. This means either 2 or 3 `[`s at the
+    // beginning are valid, and nothing more than that.
+    int8_t maxDepth = 0;
+    int8_t curDepth = 0;
+    int64_t numVerts = 6;
+    int64_t curVert = 0;
+    int64_t curLoop = 0;
+    LatLng *verts = calloc(numVerts, sizeof(LatLng));
+    int64_t strPos = 0;
+    while (polygonString[strPos] != 0) {
+        // Load more of the file if we've hit our buffer limit
+        if (strPos == 1500 && fp != 0) {
+            int res = fread(polygonString, 0, 1500, fp);
+            strPos = 0;
+            // If we didn't get any data from the file, we're done.
+            if (res == 0) {
+                break;
+            }
+        }
+        // Try to match `[` first
+        if (polygonString[strPos] == '[') {
+            curDepth++;
+            if (curDepth > maxDepth) {
+                maxDepth = curDepth;
+            }
+            strPos++;
+            continue;
+        }
+        // Similar matching for `]`
+        if (polygonString[strPos] == ']') {
+            curDepth--;
+            if (curDepth < 0) {
+                break;
+            }
+            strPos++;
+            // We may need to set upa  new geoloop at this point. If the
+            // curDepth <= maxDepth - 2 then we increment the curLoop and get a
+            // new one set up.
+            if (curDepth <= maxDepth - 2) {
+                if (curLoop == 0) {
+                    out->geoloop.verts = verts;
+                    out->geoloop.numVerts = curVert;
+                } else {
+                    GeoLoop *holes = calloc(++(out->numHoles), sizeof(GeoLoop));
+                    for (int i = 0; i < out->numHoles - 1; i++) {
+                        holes[i].numVerts = out->holes[i].numVerts;
+                        holes[i].verts = out->holes[i].verts;
+                    }
+                    free(out->holes);
+                    out->holes = holes;
+                }
+                curLoop++;
+                curVert = 0;
+                numVerts = 6;
+                verts = calloc(numVerts, sizeof(LatLng));
+            }
+            continue;
+        }
+        // Try to grab a floating point value followed by optional whitespace, a
+        // comma, and more optional whitespace, and a second floating point
+        // value, then store the lat, lng pair
+        double lat, lng;
+        int count;
+        int res = sscanf(polygonString + strPos, "%lf%*[, \n]%lf%n", &lat, &lng,
+                         &count);
+        if (count > 0 && res != 0) {
+            verts[curVert].lat = H3_EXPORT(degsToRads)(lat);
+            verts[curVert].lng = H3_EXPORT(degsToRads)(lng);
+            curVert++;
+            // Create a new vert buffer, copy the old buffer, and free it, if
+            // necessary
+            if (curVert == numVerts) {
+                LatLng *newVerts = calloc(numVerts * 2, sizeof(LatLng));
+                for (int i = 0; i < numVerts; i++) {
+                    newVerts[i].lat = verts[i].lat;
+                    newVerts[i].lng = verts[i].lng;
+                }
+                free(verts);
+                verts = newVerts;
+                numVerts *= 2;
+            }
+            strPos += count;
+            continue;
+        }
+        // Check for whitespace and skip it if we reach this point.
+        if (polygonString[strPos] == ',' || polygonString[strPos] == ' ' ||
+            polygonString[strPos] == '\n') {
+            strPos++;
+            continue;
+        } else {
+            // TODO: We've encountered an unexpected character. This may mean we
+            // need to read more of the file and put the remainder of this
+            // buffer as part of the next buffer. Or it could mean that we
+            // should just skip forward a character and try again.
+            strPos++;
+        }
+    }
+    return E_SUCCESS;
+}
+
+SUBCOMMAND(
+    polygonToCells,
+    "Converts a polygon (array of lat, lng points, or array of arrays of lat, "
+    "lng points) into a set of covering cells at the specified resolution") {
+    char filename[1024] = {0};  // More than Windows, lol
+    Arg filenameArg = {
+        .names = {"-f", "--file"},
+        .scanFormat = "%1023c",
+        .valueName = "FILENAME",
+        .value = &filename,
+        .helpText =
+            "The file to load the cells from. Use -- to read from stdin."};
+    char polygonStr[1501] = {0};  // Up to 100 cells with zero padding
+    Arg polygonStrArg = {
+        .names = {"-p", "--polygon"},
+        .scanFormat = "%1500c",
+        .valueName = "POLYGON",
+        .value = &polygonStr,
+        .helpText = "The polygon to convert. Up to 1500 characters."};
+    int res = 0;
+    Arg resArg = {.names = {"-r", "--resolution"},
+                  .required = true,
+                  .scanFormat = "%d",
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, 0-15 inclusive, that the compacted set "
+                      "should be uncompacted to. Must be greater than or equal "
+                      "to the highest resolution within the compacted set."};
+    Arg *args[] = {&polygonToCellsArg, &helpArg, &filenameArg, &polygonStrArg,
+                   &resArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (!filenameArg.found && !polygonStrArg.found) {
+        fprintf(stderr,
+                "You must provide either a file to read from or a polygon "
+                "to cover to use polygonToCells");
+        exit(1);
+    }
+    FILE *fp = 0;
+    bool isStdin = false;
+    if (filenameArg.found) {
+        if (strcmp(filename, "--") == 0) {
+            fp = stdin;
+            isStdin = true;
+        } else {
+            fp = fopen(filename, "r");
+        }
+        if (fp == 0) {
+            fprintf(stderr, "The specified file does not exist.");
+            exit(1);
+        }
+        // Do the initial population of data from the file
+        if (fread(polygonStr, 1, 1500, fp) == 0) {
+            fprintf(stderr, "The specified file is empty.");
+            exit(1);
+        }
+    }
+    GeoPolygon polygon = {0};
+    H3Error err = polygonStringToGeoPolygon(fp, polygonStr, &polygon);
+    if (err != E_SUCCESS) {
+        if (!isStdin) {
+            fclose(fp);
+        }
+        return err;
+    }
+    int64_t cellsSize = 0;
+    err = H3_EXPORT(maxPolygonToCellsSize)(&polygon, res, 0, &cellsSize);
+    if (err != E_SUCCESS) {
+        return err;
+    }
+    H3Index *cells = calloc(cellsSize, sizeof(H3Index));
+    err = H3_EXPORT(polygonToCells)(&polygon, res, 0, cells);
+    if (err != E_SUCCESS) {
+        free(cells);
+        return err;
+    }
+    // We can print out the cells
+    for (int i = 0; i < cellsSize; i++) {
+        if (cells[i] == 0) {
+            continue;
+        }
+        h3Println(cells[i]);
+    }
+    free(cells);
+    return E_SUCCESS;
+}
+
+SUBCOMMAND(maxPolygonToCellsSize,
+           "Returns the maximum number of cells that could be needed to cover "
+           "the polygon. Will always be more than actually necessary") {
+    char filename[1024] = {0};  // More than Windows, lol
+    Arg filenameArg = {
+        .names = {"-f", "--file"},
+        .scanFormat = "%1023c",
+        .valueName = "FILENAME",
+        .value = &filename,
+        .helpText =
+            "The file to load the cells from. Use -- to read from stdin."};
+    char polygonStr[1501] = {0};  // Up to 100 cells with zero padding
+    Arg polygonStrArg = {
+        .names = {"-p", "--polygon"},
+        .scanFormat = "%1500c",
+        .valueName = "POLYGON",
+        .value = &polygonStr,
+        .helpText = "The polygon to convert. Up to 1500 characters."};
+    int res = 0;
+    Arg resArg = {.names = {"-r", "--resolution"},
+                  .required = true,
+                  .scanFormat = "%d",
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, 0-15 inclusive, that the compacted set "
+                      "should be uncompacted to. Must be greater than or equal "
+                      "to the highest resolution within the compacted set."};
+    Arg *args[] = {&maxPolygonToCellsSizeArg, &helpArg, &filenameArg,
+                   &polygonStrArg, &resArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (!filenameArg.found && !polygonStrArg.found) {
+        fprintf(stderr,
+                "You must provide either a file to read from or a polygon "
+                "to cover to use polygonToCells");
+        exit(1);
+    }
+    FILE *fp = 0;
+    bool isStdin = false;
+    if (filenameArg.found) {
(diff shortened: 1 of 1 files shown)


src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,375 @@ SUBCOMMAND(uncompactCells,
...
+    "Converts a polygon (array of lat, lng points, or array of arrays of lat, "
+    "lng points) into a set of covering cells at the specified resolution") {
+    char filename[1024] = {0};  // More than Windows, lol
+    Arg filenameArg = {
+        .names = {"-f", "--file"},
+        .scanFormat = "%1023c",
+        .valueName = "FILENAME",
+        .value = &filename,
+        .helpText =
+            "The file to load the cells from. Use -- to read from stdin."};
```
isaacbrodsky, line 1311:
```suggestion
            "The file to load the polygon from. Use -- to read from stdin."};
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,375 @@ SUBCOMMAND(uncompactCells,
...
+           "Returns the maximum number of cells that could be needed to cover "
+           "the polygon. Will always be more than actually necessary") {
+    char filename[1024] = {0};  // More than Windows, lol
+    Arg filenameArg = {
+        .names = {"-f", "--file"},
+        .scanFormat = "%1023c",
+        .valueName = "FILENAME",
+        .value = &filename,
+        .helpText =
+            "The file to load the cells from. Use -- to read from stdin."};
```
isaacbrodsky, line 1397:
```suggestion
            "The file to load the polygon from. Use -- to read from stdin."};
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+            continue;
+        }
+        // Similar matching for `]`
+        if (polygonString[strPos] == ']') {
+            curDepth--;
+            if (curDepth < 0) {
+                break;
+            }
+            strPos++;
+            // We may need to set upa  new geoloop at this point. If the
```
isaacbrodsky, line 1235: "up a"

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+            int res = fread(polygonString, 1, 1500, fp);
+            strPos = 0;
+            // If we didn't get any data from the file, we're done.
+            if (res == 0) {
+                break;
+            }
+        }
+        // Try to match `[` first
+        if (polygonString[strPos] == '[') {
+            curDepth++;
```
isaacbrodsky, line 1221: If you had more than a couple hundred [ in a row, signed integer overflow would occur. How concerned do we need to be about that?

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+                verts = newVerts;
+                numVerts *= 2;
+            }
+            strPos += count;
+            curDepth--;
+            continue;
+        }
+        // Check for whitespace and skip it if we reach this point.
+        if (polygonString[strPos] == ',' || polygonString[strPos] == ' ' ||
+            polygonString[strPos] == '\n') {
```
isaacbrodsky, line 1293: Tab \t?

src/apps/filters/h3.c
dfellis:
```suggestion
            // We may need to set up a new geoloop at this point. If the
```

dfellis:
I just ran [this failure](https://github.com/uber/h3/actions/runs/11227958246/job/31211057915) locally (set up CMAKE with the same configuration) and I can't reproduce it. @isaacbrodsky any idea what's going on here?

src/apps/filters/h3.c
dfellis:
```suggestion
            polygonString[strPos] == '\n' || polygonString[strPos] == '\t') {
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+                verts = newVerts;
+                numVerts *= 2;
+            }
+            strPos += count;
+            curDepth--;
+            continue;
+        }
+        // Check for whitespace and skip it if we reach this point.
+        if (polygonString[strPos] == ',' || polygonString[strPos] == ' ' ||
+            polygonString[strPos] == '\n' || polygonString[strPos] == '\t') {
```
isaacbrodsky, line 1293: or `\r`?

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+    int res = 0;
+    Arg resArg = {.names = {"-r", "--resolution"},
+                  .required = true,
+                  .scanFormat = "%d",
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, 0-15 inclusive, that the compacted set "
+                      "should be uncompacted to. Must be greater than or equal "
+                      "to the highest resolution within the compacted set."};
```
isaacbrodsky, line 1376:
```suggestion
                      "Resolution, 0-15 inclusive, that the polygon "
                      "should be converted to."};
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
```
isaacbrodsky, line 1467:
```suggestion
                      "Resolution, 0-15 inclusive, that the polygon "
                      "should be converted to."};
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+    if (err != E_SUCCESS) {
+        return err;
+    }
+    printf("%" PRId64 "\n", cellsSize);
+    return E_SUCCESS;
+}
+
+SUBCOMMAND(cellsToMultiPolygon,
+           "Returns a polygon (array of lat, lng points, or array of arrays of "
+           "lat, lng points) for a set of cells") {
```
isaacbrodsky, line 1520:
```suggestion
           "Returns a polygon (array of arrays of "
           "lat, lng points) for a set of cells") {
```
the type doesn't change, so let's remove the "or"

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+        .value = &filename,
+        .helpText =
+            "The file to load the cells from. Use -- to read from stdin."};
+    char cellStrs[1501] = {0};  // Up to 100 cells with zero padding
+    Arg cellStrsArg = {.names = {"-c", "--cells"},
+                       .scanFormat = "%1500c",
+                       .valueName = "CELLS",
+                       .value = &cellStrs,
+                       .helpText =
+                           "The cells to compact. Up to 100 cells if provided "
```
isaacbrodsky, line 1535:
```suggestion
                           "The cells to convert. Up to 100 cells if provided "
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+                           "as hexadecimals with zero padding."};
+    Arg *args[] = {&cellsToMultiPolygonArg, &helpArg, &filenameArg,
+                   &cellStrsArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (!filenameArg.found && !cellStrsArg.found) {
+        fprintf(stderr,
+                "You must provide either a file to read from or a set of cells "
+                "to compact to use compactCells");
```
isaacbrodsky, line 1543:
```suggestion
                "You must provide either a file to read from or a set of cells "
                "to convert to use cellsToMultiPolygon");
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+                      "Resolution, 0-15 inclusive, that the compacted set "
+                      "should be uncompacted to. Must be greater than or equal "
+                      "to the highest resolution within the compacted set."};
+    Arg *args[] = {&maxPolygonToCellsSizeArg, &helpArg, &filenameArg,
+                   &polygonStrArg, &resArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (!filenameArg.found && !polygonStrArg.found) {
+        fprintf(stderr,
+                "You must provide either a file to read from or a polygon "
+                "to cover to use polygonToCells");
```
isaacbrodsky, line 1474:
```suggestion
                "You must provide either a file to read from or a polygon "
                "to cover to use maxPolygonToCellsSize");
```

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+        }
+        h3Println(cells[i]);
+    }
+    free(cells);
+    return E_SUCCESS;
+}
+
+SUBCOMMAND(maxPolygonToCellsSize,
+           "Returns the maximum number of cells that could be needed to cover "
+           "the polygon. Will always be more than actually necessary") {
```
isaacbrodsky, line 1442:
```suggestion
           "the polygon. Will always be equal or more than actually necessary") {
```

src/apps/filters/h3.c
isaacbrodsky: We could support that up to a reasonable limit. Perhaps before incrementing, check if overflow would occur (equals to INT8_MAX) and if so return an error. I haven't checked what would happen if signed overflow occurred but I presume it's not great.

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, 0-15 inclusive, that the compacted set "
+                      "should be uncompacted to. Must be greater than or equal "
+                      "to the highest resolution within the compacted set."};
+    Arg *args[] = {&polygonToCellsArg, &helpArg, &filenameArg, &polygonStrArg,
+                   &resArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (!filenameArg.found && !polygonStrArg.found) {
```
isaacbrodsky, line 1380:
```suggestion
    if (filenameArg.found == polygonStrArg.found) {
```

both shouldn't be provided, right?

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+                  .valueName = "res",
+                  .value = &res,
+                  .helpText =
+                      "Resolution, 0-15 inclusive, that the compacted set "
+                      "should be uncompacted to. Must be greater than or equal "
+                      "to the highest resolution within the compacted set."};
+    Arg *args[] = {&maxPolygonToCellsSizeArg, &helpArg, &filenameArg,
+                   &polygonStrArg, &resArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (!filenameArg.found && !polygonStrArg.found) {
```
isaacbrodsky, line 1471:
```suggestion
    if (filenameArg.found == polygonStrArg.found) {
```
as above

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,433 @@ SUBCOMMAND(uncompactCells,
...
+                       .scanFormat = "%1500c",
+                       .valueName = "CELLS",
+                       .value = &cellStrs,
+                       .helpText =
+                           "The cells to compact. Up to 100 cells if provided "
+                           "as hexadecimals with zero padding."};
+    Arg *args[] = {&cellsToMultiPolygonArg, &helpArg, &filenameArg,
+                   &cellStrsArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (!filenameArg.found && !cellStrsArg.found) {
```
isaacbrodsky, line 1540:
```suggestion
    if (filenameArg.found == polygonStrArg.found) {
```
as above

src/apps/filters/h3.c
dfellis: No, the `filenameArg` is you provide a file containing the polygon, the `polygonStrArg` is if you provide the polygon directly in the command.

src/apps/filters/h3.c
dfellis: Same. This is right as-is.

src/apps/filters/h3.c
dfellis: Ugh, Windows... Yeah, I'll tackle that.

dfellis:
```suggestion
        if (polygonString[strPos] == ',' || polygonString[strPos] == ' ' ||
            polygonString[strPos] == '\n' || polygonString[strPos] == '\t' ||
            polygonString[strPos] == '\r') {
```

src/apps/filters/h3.c
dfellis: Same. This change would break things.

src/apps/filters/h3.c
dfellis: Then I should probably just stop it at 4 and error at that, since even that depth is enough to be invalid.

dfellis: There. Let me know if the latest change is satisfactory @isaacbrodsky

src/apps/filters/h3.c
isaacbrodsky: But if you provide both then we overwrite polygonStr on line 1404, so one of them will be disregarded.

My thought is,
* If both are false (and thus equal), one of them should have been provided, and thus error.
* If both are true (and thus equal), one of them will be disregarded, and thus error.  This is the case my change would address.
* If they have different found values, one of them was found and one of them was not found, so we have exactly one input specifying the polygon.

dfellis: Ahh, I got it, I agree. Just a sec.

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+                       .scanFormat = "%1500c",
+                       .valueName = "CELLS",
+                       .value = &cellStrs,
+                       .helpText =
+                           "The cells to convert. Up to 100 cells if provided "
+                           "as hexadecimals with zero padding."};
+    Arg *args[] = {&cellsToMultiPolygonArg, &helpArg, &filenameArg,
+                   &cellStrsArg};
+    PARSE_SUBCOMMAND(argc, argv, args);
+    if (filenameArg.found == polygonStrArg.found) {
```
isaacbrodsky, line 1544:
```suggestion
    if (filenameArg.found == cellStrsArg.found) {
```
my bad on the suggestion sorry

isaacbrodsky (approved):

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+    err = H3_EXPORT(maxPolygonToCellsSize)(&polygon, res, 0, &cellsSize);
+    for (int i = 0; i < polygon.numHoles; i++) {
+        free(polygon.holes[i].verts);
+    }
+    free(polygon.holes);
+    free(polygon.geoloop.verts);
+    if (err != E_SUCCESS) {
+        return err;
+    }
+    printf("%" PRId64 "\n", cellsSize);
```
nrabinowitz, line 1518: I'm a little confused by the format here - isn't that what we use for `h3Print`, and doesn't that mean that the output will be in hexadecimal instead of base 10 digits? The tests look fine, so I must be wrong, but in that case I don't think I know what this format actually does

nrabinowitz (approved):
Left some comments, but overall this looks great!

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+    int8_t maxDepth = 0;
+    int8_t curDepth = 0;
+    int64_t numVerts = 6;
+    int64_t curVert = 0;
+    int64_t curLoop = 0;
+    LatLng *verts = calloc(numVerts, sizeof(LatLng));
+    int strPos = 0;
+    while (polygonString[strPos] != 0) {
+        // Load more of the file if we've hit our buffer limit
+        if (strPos >= 1500 && fp != 0) {
```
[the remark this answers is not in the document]
dfellis: Sure.

src/apps/filters/h3.c
isaacbrodsky: You're thinking of PRIx64, this is PRId64, the d indicates decimal

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+        }
+        // Try to match `[` first
+        if (polygonString[strPos] == '[') {
+            curDepth++;
+            if (curDepth > maxDepth) {
+                maxDepth = curDepth;
+            }
+            if (curDepth > 4) {
+                // This is beyond the depth for a valid input, so we abort early
+                return E_FAILED;
```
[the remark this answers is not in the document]
dfellis: I didn't want to pollute the error enum with CLI errors, which is why I just used the unspecified error enum here. I don't really want to rewrite the error handling logic in the CLI to support two sets of enum types since H3Error reporting is baked into the macro-generated return types of all of the CLI functions.

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+            strPos++;
+            // We may need to set up a new geoloop at this point. If the
+            // curDepth <= maxDepth - 2 then we increment the curLoop and get a
+            // new one set up.
+            if (curDepth <= maxDepth - 2) {
+                if (curLoop == 0) {
+                    out->geoloop.verts = verts;
+                    out->geoloop.numVerts = curVert;
+                } else {
+                    GeoLoop *holes = calloc(out->numHoles + 1, sizeof(GeoLoop));
```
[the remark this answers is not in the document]
dfellis: Sure.

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+                    for (int i = 0; i < out->numHoles; i++) {
+                        holes[i].numVerts = out->holes[i].numVerts;
+                        holes[i].verts = out->holes[i].verts;
+                    }
```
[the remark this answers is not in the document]
dfellis: Yes. We're also copying the *pointers* not the values in the pointers, so it's a similar cost to scaling up a JS array beyond its initial capacity (not exactly the same since that tends to double the new capacity, while I'm scaling for each new hole, but yeah, the number of holes per polygon is not expected to be large enough for that to matter).

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+            curDepth--;
+            continue;
+        }
+        // Check for whitespace and skip it if we reach this point.
+        if (polygonString[strPos] == ',' || polygonString[strPos] == ' ' ||
+            polygonString[strPos] == '\n' || polygonString[strPos] == '\t' ||
+            polygonString[strPos] == '\r') {
+            strPos++;
+            continue;
+        } else {
```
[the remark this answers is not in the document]
dfellis: That's true. It wasn't there originally, but since it's there we don't need the else branch.

src/apps/filters/h3.c
```diff
@@ -1189,6 +1189,437 @@ SUBCOMMAND(uncompactCells,
...
+            if (strPos != 0 && fp != 0) {
```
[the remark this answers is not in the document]
dfellis: When you specify the string being parsed as a command line argument rather than passing the filename to read from. (This let me DRY up the parsing logic instead of having one for CLI arg and another for files.)

src/apps/filters/h3.c
dfellis: This is `PRId64` not `PRIx64`, so it prints the decimal version of the 64-bit integer.

src/apps/filters/h3.c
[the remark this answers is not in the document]
nrabinowitz: Got it, thanks!

src/apps/filters/h3.c
nrabinowitz: Ah, gotcha, thanks all

PR #933, opened 2024-10-17
Update the cli output to be controllable, when possible, and always produce JSON-compliant output by default

dfellis:
I will update the website docs tomorrow with these changes, but putting this up for review now.


src/apps/filters/h3.c
```diff
@@ -113,11 +113,22 @@ bool has(char *subcommand, int level, char *argv[]) {
 
 Arg helpArg = ARG_HELP;
 
+#define DEFINE_FORMAT_ARG(desc)                   \
+    char format[8] = {0};                         \
+    Arg formatArg = {.names = {"-f", "--format"}, \
+                     .scanFormat = "%s",          \
```
isaacbrodsky, line 119: This should have the format buffer length here. `%7c` or `%8c`?
```suggestion
                     .scanFormat = "%8c",          \
```

isaacbrodsky (approved):

ajfriend (approved):

src/apps/filters/h3.c
dfellis: I don't think that's the right solution. It would *require* there to be 8 characters in order to read them with that form, while `%s` reads up to the next whitespace character.

I expect that would either fail to parse at all, or would accidentally grab characters from the next argument, depending on how it's laid out in memory.

But [reading the docs](https://www.w3schools.com/c/ref_stdio_scanf.php), it looks like `%7s` would read up to seven characters (leaving one zero for string termination), to avoid a buffer overrun issue. I'm going to try that.

isaacbrodsky: That looks right to me

PR #936, opened 2024-10-20
Run cmake-format on all CMake files

isaacbrodsky:
We could choose to enforce this in CI, but not sure the format is sufficiently better to do that & require everyone run it

Here is the command used:
```
cmake-format CMakeLists.txt CMakeTests.cmake cmake/Config.cmake.in cmake/TestWrapValgrind.cmake tests/cli/*.txt -i
```


dfellis:
This PR misses all of the cmake text in [`/tests/cli`](https://github.com/uber/h3/tree/5d5ad33901e42550aa3dd310c96592b5298a5dbb/tests/cli). Perhaps we should rename them to `*.cmake` and modify the macro embedding them to use that new extension instead of `*.txt`?

dfellis (approved):

ajfriend (approved):

PR #938, opened 2024-10-22
Update docs to match h3-py v4.x

ajfriend:
Update the cells <> polygon functions in the h3geo.org docs to match h3-py v4.x.

Update: Also rework the h3geo.org docs so that language-agnostic function descriptions are given first, and language-specific details are given below the language tabs.

This also fixes some bugs in the docs that I found along the way.

### DIFF
--- a/website/docs/api/directededge.mdx
+++ b/website/docs/api/directededge.mdx
@@ -13,6 +13,8 @@ edge from one cell to a neighboring cell.
 
 ## areNeighborCells
 
+Determines whether or not the provided H3 cells are neighbors.
+
 <Tabs
   groupId="language"
   defaultValue="c"
@@ -29,6 +31,8 @@ edge from one cell to a neighboring cell.
 ```c
 H3Error areNeighborCells(H3Index origin, H3Index destination, int *out);
 ```
+
+`out` will be 1 if the indexes are neighbors, 0 otherwise.
 
 </TabItem>
 <TabItem value="java">
@@ -83,11 +87,11 @@ true
 </TabItem>
 </Tabs>
 
-Returns whether or not the provided H3 cell indexes are neighbors.
-
-`out` will be 1 if the indexes are neighbors, 0 otherwise.
 
 ## cellsToDirectedEdge
+
+Returns a directed edge H3 index based on the provided origin and
+destination.
 
 <Tabs
   groupId="language"
@@ -105,6 +109,8 @@ Returns whether or not the provided H3 cell indexes are neighbors.
 ```c
 H3Error cellsToDirectedEdge(H3Index origin, H3Index destination, H3Index *out);
 ```
+
+Returns 0 on success.
 
 </TabItem>
 <TabItem value="java">
@@ -159,12 +165,9 @@ $ h3 cellsToDirectedEdge -o 85283473fffffff -d 85283477fffffff
 </TabItem>
 </Tabs>
 
-Returns a unidirectional edge H3 index based on the provided origin and
-destination.
-
-Returns 0 on success.
-
 ## isValidDirectedEdge
+
+Determines if the provided H3Index is a valid unidirectional edge index.
 
 <Tabs
   groupId="language"
@@ -182,6 +185,8 @@ Returns 0 on success.
 ```c
 int isValidDirectedEdge(H3Index edge);
 ```
+
+Returns 1 if it is a unidirectional edge H3Index, otherwise 0.
 
 </TabItem>
 <TabItem value="java">
@@ -234,11 +239,9 @@ true
 </TabItem>
 </Tabs>
 
-Determines if the provided H3Index is a valid unidirectional edge index.
-
-Returns 1 if it is a unidirectional edge H3Index, otherwise 0.
-
 ## getDirectedEdgeOrigin
+
+Returns the origin hexagon from the directed edge H3Index.
 
 <Tabs
   groupId="language"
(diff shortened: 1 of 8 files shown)


ajfriend:
One issue I see in the docs, however, is that the text below is shown for all the languages, but may not apply to all equally:

> polygonToCells takes a given GeoJSON-like data structure and preallocated, zeroed memory, and fills it with the hexagons that are contained by the GeoJSON-like data structure.
> 
> Containment is determined by the cells' centroids. A partitioning using the GeoJSON-like data structure, where polygons cover an area without overlap, will result in a partitioning in the H3 grid, where cells cover the same area without overlap.

The Python function does not take a GeoJSON-like data structure, for example

isaacbrodsky (approved):

isaacbrodsky:
> One issue I see in the docs, however, is that the text below is shown for all the languages, but may not apply to all equally:
> 
> > polygonToCells takes a given GeoJSON-like data structure and preallocated, zeroed memory, and fills it with the hexagons that are contained by the GeoJSON-like data structure.
> > Containment is determined by the cells' centroids. A partitioning using the GeoJSON-like data structure, where polygons cover an area without overlap, will result in a partitioning in the H3 grid, where cells cover the same area without overlap.
> 
> The Python function does not take a GeoJSON-like data structure, for example.
> 
> Would it make sense to have some of the text be language-specific?

Maybe it should be written more generally, and the parts specific to the C api placed under the appropriate tab. Preallocated memory does not apply to any of the bindings.

website/docs/api/regions.mdx
```diff
@@ -60,7 +60,7 @@ function example() {
 <TabItem value="python">
 
 ```py
-h3.polygon_to_cells(polygons, res, geo_json_conformant=False)
+h3.h3shape_to_cells(h3shape, res)
```
dfellis, line 63: `polygon_to_cells` does exist in the new version of `h3-py`, though. Wouldn't that make more sense to put here?

ajfriend: I had wanted to direct folks to `h3shape_to_cells`, since `polygon_to_cells` is a weird duck in Python, but maybe that's what the `h3-py` docs are for :)

I can have it be `polygon_to_cells` here.

ajfriend:
> Maybe it should be written more generally, and the parts specific to the C api placed under the appropriate tab. Preallocated memory does not apply to any of the bindings.

I took a stab as an example. Let me know what you think!

website/docs/api/regions.mdx
```diff
@@ -12,6 +12,15 @@ These functions convert H3 indexes to and from polygonal areas.
...
 ## polygonToCells
 
+Each binding's version of `polygonToCells` takes as input a GeoJSON-like data
+structure describing a polygon (i.e., an outer ring and optional holes) and
+a target cell resolution.
+It produces a collection of cells that are contained within the polygon.
+
+Containment is determined by centroids of the cells, so that a partitioning
+of polygons (covering an area without overlaps) will result in
+a partitioning of H3 cells.
```
dfellis, line 22: This looks good, but we usually have the explanatory text after the code example throughout the docs, so maybe this should be moved back down?

website/docs/api/regions.mdx
dfellis: I think your compromise solution of having both API calls mentioned is fine here.

website/docs/api/regions.mdx
ajfriend: I had considered that. But one benefit of having it this way is that it provides clear separation between the text that changes when you click a tab, and the text that doesn't.

dfellis: Well... then we probably need to expand the scope of this PR to re-organize all of the explanatory text along those lines (unchanging explanation above, per-language explanation below) :/

ajfriend: Yeah, I can do that. Does everyone like the new organization I have above?

isaacbrodsky (approved):

ajfriend:
I don't think we explain the difference between "centroid" and "center" in the docs, but we should. The old description here makes a distinction: https://github.com/uber/h3/pull/938/commits/e3a48add4f4e0adbec81593119451812f3670586

ajfriend:
This is ready for review. I made a bunch of changes, so I'll re-request.

website/docs/api/directededge.mdx
```diff
@@ -525,12 +546,12 @@ $ h3 originToDirectedEdges -c 85283473fffffff
 </TabItem>
 </Tabs>
 
-Provides all of the directed edges from the current H3Index. `edges` must be of length 6,
-and the number of directed edges placed in the array may be less than 6. If this is the case,
-one of the members of the array will be `0`.
```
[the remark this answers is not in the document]
ajfriend: That's how I have it. I think you're looking at the deleted text here.

website/docs/api/inspection.mdx
```diff
@@ -532,14 +573,10 @@ $ h3 getIcosahedronFaces -c 85283473fffffff
 </TabItem>
 </Tabs>
 
-Find all icosahedron faces intersected by a given H3 index and places them in the array `out`. `out` must be at least length of `maxFaceCount(h)`.
-
-Faces are represented as integers from 0-19, inclusive. The array is sparse, and empty (no intersection) array values are represented by -1.
```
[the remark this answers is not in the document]
ajfriend: That's how I have it. I think you're looking at the removed text here.

website/docs/api/directededge.mdx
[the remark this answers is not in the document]
dfellis: Oh, you're right on this one. I missed it.

dfellis (approved):

isaacbrodsky (approved):

PR #948, opened 2024-12-03
Update overview.md

aaronjohnson:
Fixed a broken link to an external PDF.

Perhaps there is a more canonical location.

### DIFF
--- a/website/docs/core-library/overview.md
+++ b/website/docs/core-library/overview.md
@@ -5,7 +5,7 @@ sidebar_label: Overview
 slug: /core-library/overview
 ---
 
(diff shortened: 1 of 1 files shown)


ajfriend:
I think, technically, we should be pointing to the publisher's website here: https://www.tandfonline.com/doi/abs/10.1559/152304003100011090

Unfortunately, the PDF is paywalled at that link, but that's the proper way to handle this. But hopefully it isn't too hard for an enterprising person to find legitimate access to the PDF via other sites (including Dr. Sahr's website). I just don't think we can link it directly here.

sahrk:
> I think, technically, we should be pointing to the publisher's website

Agreed. The paper is owned by the publisher, so the only legal link to it is the publisher's website. I note that the paper is available on [LibGen](https://libgen.is/).

PR #955, opened 2024-12-16
Increase timeout threshold for fuzzer CI run

isaacbrodsky:
This is to avoid getting failures about the fuzzer timing out, which seems to happen since some cases may be slower than before. This is a bit inexact since perhaps there is some bad behavior with larger numbers of cells.

e.g. https://github.com/uber/h3/actions/runs/12289773577/job/34295819216?pr=953

### DIFF
--- a/.github/workflows/test-fuzzer.yml
+++ b/.github/workflows/test-fuzzer.yml
@@ -69,5 +69,5 @@ jobs:
           cd build
           for fuzzer in bin/fuzzer*; do
             echo $fuzzer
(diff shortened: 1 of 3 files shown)


isaacbrodsky:
The original idea of reducing the fuzzer max size parameter didn't really work, so let's try increasing the timeout. OSS-Fuzz for comparison allows >= 1 hour before a timeout from what I see: https://github.com/search?q=repo%3Agoogle%2Foss-fuzz%20timeout&type=code

ajfriend (approved):

PR #968, opened 2025-02-17
new fast isValidCell

ajfriend:
Taking over from https://github.com/uber/h3/pull/496.

New `isValidCell` should be faster in all cases. Relies on bit manipulation instead of loops.

In the last correctness check, we use compiler intrinsics, so we'll want to verify that that approach works on all platforms. We can always fall back to a loop, and that would still be faster (see the benchmarks plot below). However, we need to ensure that we have the right test to evaluate which approaches are available on each platform. I'm not currently sure how to do that in an exhaustive way.


ajfriend:
OK, current approach breaks some rules:

```
/home/runner/work/h3/h3/src/h3lib/lib/h3Index.c: In function ‘_isValidCell_const’:
/home/runner/work/h3/h3/src/h3lib/lib/h3Index.c:297:14: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
```

ajfriend:
Speedup results on my M3 mac are promising. This is for `8_14_null_100`.

- old is the original algorithm
- final is the final algorithm using intrinsics
- fallback is the final algorithm, but will the fallback for when intrinsics are not availab.e

image

Speedups, comparing the fastest run for each name and algo:

```
┌───────────────┬─────────┬─────────┬─────────┐
│     name      │  algo1  │  algo2  │ speedup │
│    varchar    │ varchar │ varchar │ double  │
├───────────────┼─────────┼─────────┼─────────┤
│ 2_8           │ old     │ final   │     6.4 │
│ 8_14          │ old     │ final   │     9.1 │
│ 8_14_null_10  │ old     │ final   │     7.8 │
│ 8_14_null_100 │ old     │ final   │     8.8 │
│ 8_14_null_2   │ old     │ final   │     5.0 │
└───────────────┴─────────┴─────────┴─────────┘
```

ajfriend:
Latest benchmarks of final algo:

```
┌───────────────┬──────────────┬───────┐
│     name      │ microseconds │ runs  │
│    varchar    │    double    │ int64 │
├───────────────┼──────────────┼───────┤
│ 2_8           │      145.976 │   500 │
│ 8_14          │      145.646 │   500 │
│ 8_14_null_10  │      157.995 │   500 │
│ 8_14_null_100 │      149.786 │   500 │
│ 8_14_null_2   │      145.231 │   500 │
└───────────────┴──────────────┴───────┘
```

ajfriend:
TODO: I think we test 32 bit windows in CI. Should we do the same for mac and linux?

isaacbrodsky:
> TODO: We test 32 bit windows in CI. Should we do the same for mac and linux?

I don't think there is an equivalent Mac 32-bit.
Linux we can test a x86 or ARM 32-bit configuration

isaacbrodsky (approved):

src/h3lib/lib/h3Index.c
```diff
@@ -119,50 +119,183 @@ H3Error H3_EXPORT(h3ToString)(H3Index h, char *str, size_t sz) {
...
+
+We can check that (in the lower 45 = 15*3 bits) the position of the
+first 1 bit isn't divisible by 3.
+*/
+static inline bool _hasDeletedSubsequence(H3Index h, int base_cell) {
+    static const bool isBaseCellPentagonArr[128] = {
+        [4] = 1,  [14] = 1, [24] = 1, [38] = 1, [49] = 1,  [58] = 1,
+        [63] = 1, [72] = 1, [83] = 1, [97] = 1, [107] = 1, [117] = 1};
+
+    if (isBaseCellPentagonArr[base_cell]) {
```
isaacbrodsky, line 247: Is there a reason not to use (perhaps an inlined version of) `_isBaseCellPentagon`?

ajfriend: Good point. I'll run some benchmarks and see if there's any difference.

ajfriend:
> > TODO: We test 32 bit windows in CI. Should we do the same for mac and linux?
> 
> I don't think there is an equivalent Mac 32-bit. Linux we can test a x86 or ARM 32-bit configuration

Think that's a blocker for this PR, or can I create a separate task/PR to add that to CI?

isaacbrodsky:
> > > TODO: We test 32 bit windows in CI. Should we do the same for mac and linux?
> > 
> > 
> > I don't think there is an equivalent Mac 32-bit. Linux we can test a x86 or ARM 32-bit configuration
> 
> Think that's a blocker for this PR, or can I create a separate task/PR to add that to CI? @isaacbrodsky

I don't think it's a blocker, but we could certainly merge it in first.

My sense is Linux x86/i*86 is much less common today and probably not an architecture we need to be too concerned with testing.

I'll open a PR to include Linux arm64. I don't think Github provides Linux arm32 runners, at least as standard: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

src/h3lib/lib/h3Index.c
ajfriend: Definitely slower with the naive use of `_isBaseCellPentagon`. Test setup in the latest commit:

image

I'll figure out the right `extern` and `inline` incantation and try that next. 

Another alternative might be to just expose the `BaseCellData` directly.

More generally for the H3 lib as a whole, I'm not sure if there's a benefit to using a struct of arrays vs an array of structs.

ajfriend:
Issues trying to inline:

image

image

dfellis:
I'm not super familiar with inline functions in C, but [going by this StackOverflow answer](https://stackoverflow.com/a/34140421) you need `inline int ...` in the header, and `extern int ...` in the implementation.

ajfriend:
> I'm not super familiar with inline functions in C, but [going by this StackOverflow answer](https://stackoverflow.com/a/34140421) you need `inline int ...` in the header, and `extern int ...` in the implementation.

Nice. That works for me locally, but with some errors. Let's see now it does in CI.

src/h3lib/lib/h3Index.c
ajfriend: I think doing this will be a little involved, so I'm proposing doing it later :)

https://github.com/uber/h3/issues/984

dfellis (approved):

ajfriend:
Note the follow-up: https://github.com/uber/h3/issues/984

PR #974, opened 2025-02-25
ci: include ubuntu arm test, clean up check names

isaacbrodsky:
Adds Linux arm64 test. I added it only for GCC to avoid exploding the test matrix (since clang has more options that are tested). It would be a valid configuration to also test, so open to suggestions. Maybe we throw it in?

Since the Github UI already shows which file the test comes from, including "Compile" or "Test" seem superfluous.

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2.4.0
+      - uses: actions/checkout@v4
 
       - uses: actions/setup-node@v2.4.1
         with:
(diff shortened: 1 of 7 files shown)


dfellis (approved):

ajfriend (approved):

ajfriend:
> I added it only for GCC to avoid exploding the test matrix (since clang has more options that are tested). It would be a valid configuration to also test, so open to suggestions. Maybe we throw it in?

How much longer does it take to run? Maybe try in a separate PR? We could also run it just for a subset of options on clang.

We could also split these up to run a faster subset on draft PRs, like we do in h3-py.

ajfriend:
Also, I see a bunch of warnings in the Actions output. Maybe time to update the checkout (and other) actions, since that looks like where its coming from:

image

PR #975, opened 2025-02-25
ci: test on clang/arm64

isaacbrodsky:


ajfriend:
I can't immediately tell what's failing. Maybe try `fail-fast: false` to get the complete list of what passes and what doesn't?

dfellis:
> I can't immediately tell why things are failing. Maybe try `fail-fast: false` to get the complete list of what passes and what doesn't?

https://github.com/uber/h3/actions/runs/13526744257/job/37799141287#step:8:570

Every single test fails immediately.

We should see if we can get it to dump the logs on failure, but I suspect it is accidentally trying to run x86-64 binaries in an arm64 environment.

ajfriend:
> Every single test fails immediately.

Only one check fails, and the rest in that group are cancelled early

dfellis:
> 
> > Every single test fails immediately.
> 
> Only one check fails, and the rest in that group are cancelled early 
> 
> 
> 

Not what I meant. If you open the link, you will see that every test is executed *for that run* and they all fail. That implies that it is not generating binaries that work for the platform.

isaacbrodsky:
I'll enable the print-output-on-failure option in CMake so we can find out. I wonder if some of these sanitizers are not implemented/available on arm64.

dfellis:
Definitely better logging here. https://github.com/uber/h3/actions/runs/13529596356/job/37808380267#step:8:10522

Do these compile arg recommendations make sense for this test?

ajfriend:
@dfellis, ah, we were talking about different things. I mentioned `fail-fast: false` as a way to determine which checks are failing (as in what combinations of options/sanitizers aren't working).

That still might be helpful here, even if we only leave it on temporarily for this PR.

ajfriend:
From what I can tell, `-fsanitize=memory -fno-sanitize-recover=memory` just isn't supported on ARM. Let's just drop it for ARM?

ajfriend:
So maybe drop this test for now to get things passing? And we can create a follow up issue to update clang and add msan.

ajfriend (approved):

dfellis (approved):

PR #981, opened 2025-03-06
CMake: Fix wrong MSVC related conditions

### LINKED ISSUE #980: Windows: Wrong condition for `ENABLE_WARNINGS`
#831 added `ENABLE_WARNINGS` but it has a wrong condition for choosing suitable compiler flag. We can use `/W2` with MSVC but `/W2` is used for MinGW too.

See also: https://github.com/uber/h3/pull/831/files#r1983232019

Reported by @giordano . Thanks!

kou:
Fix #980

MSVC and GCC/Clang use different style for command line options. We should use "MSVC" not "WIN32" for detecting MSVC. If we use "WIN32", mingw-w64 also uses MSVC style command line options because "WIN32" is true for both of MSVC and mingw-w64.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,14 +101,14 @@ set(H3_COMPILE_FLAGS "")
 set(H3_LINK_FLAGS "")
 option(ENABLE_WARNINGS "Enables compiler warnings" ON)
 if(ENABLE_WARNINGS)
-    if(WIN32)
+    if(MSVC)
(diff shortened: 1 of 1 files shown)


kou:
I noticed that we have similar problems not only for `ENABLE_WARNINGS` but also for `WARNINGS_AS_ERRORS` and debug related options.

dfellis (approved):

isaacbrodsky (approved):

ajfriend (approved):

PR #1005, opened 2025-05-05
website: Docs for DuckDB SQL

isaacbrodsky:


website/docs/api/directededge.mdx
```diff
@@ -23,6 +23,7 @@ Determines whether or not the provided H3 cells are neighbors.
     {label: 'Java', value: 'java'},
     {label: 'JavaScript (Live)', value: 'javascript'},
     {label: 'Python', value: 'python'},
+    {label: 'SQL', value: 'sql'},
```
dfellis, line 26: Should this say "SQL" or "DuckDB"? Because there is [h3-pg](https://github.com/zachasme/h3-pg), as well.

isaacbrodsky: I was actually hoping that Postgres, DuckDB, presto, snowflake etc could all use the same function names and so on. Taking a quick look at  Postgres extension I suppose that is not exactly how it works right now, which is unfortunate.

zachasme: I'm open to changing function names for the sake of consistency. Generally, I renamed functions from `camelCase` to `snake_case`, while adding a `h3_` prefix (except for functions that already begun with `h3_`).

Differences appear to be double `h3_h3_` prefix and `latlng` vs `lat_lng`. Are there more I'm missing?

isaacbrodsky: Those were the main differences I noticed. In DuckDB I didn't add an extension type, so there is an additional h3_latlng_to_cell_string for indexing directly to the string form instead of uint64.

zachasme: Great, I'll add the new function names in the next release, and soft-deprecate the old ones. :+1:  Just out of curiosity, is there a reason all the other SQL bindings landed on the same naming convention?

dfellis: Well, [`h3-presto`](https://github.com/foursquare/h3-presto) was [mostly written by @isaacbrodsky](https://github.com/foursquare/h3-presto/commits/main/) ;)

Snowflake's H3 support is independent, however, [but also only a subset of functionality](https://docs.snowflake.com/en/sql-reference/functions-geospatial). So we may still need to include special logic in the documentation to point out functions that don't exist in certain SQL dialects.

tbiraj8 (approved):

isaacbrodsky:
For the time being I've renamed the tab to DuckDB to be clear which SQL dialect is being described, I may open a new PR for a combined SQL tab

dfellis (approved):

ajfriend (approved):

zachasme:
@isaacbrodsky I would love to be able to direct users to the SQL docs on h3geo.org. What are your thoughts on aligning the docs on SQL bindings?

I'm working on [the new function names](https://github.com/zachasme/h3-pg/pull/177) in the PostgreSQL bindings. For now I only had to rename `lat_lng` to `latlng`. There is still a few differences around inputs (string and integer representations in DuckDB and a custom type in Postgres) and outputs (array return value plus separate _lat/_lng suffixed variants in DuckDB, and (non-postgis) Point return values in Postgres).

I could make a PR for Postgres-specific docs, but I feel we are so close to API parity that we should try and go for a combined SQL tab. Would you be interested in a PR for a combined tab with a few notes about differences where necessary?

PR #1011, opened 2025-06-04
feat: Implement gridRing

justinhwang:
This commit introduces a new public API function, `gridRing`, to
retrieve H3 cells located at an exact grid distance `k` from an origin
cell, forming a "hollow" ring.

The implementation mirrors implementation of `gridDiskDistances`
wherein we first attempt to use the unsafe version. If the unsafe
version fails, we fallback to `_gridRingInternal`.

`_gridRingInternal` utilizes `_gridDiskDistancesInternal` to get all
cells up to distance `k`, then filters the output to only include those
at exactly `k`.


justinhwang:
Not quite sure why were getting these errors
```
/home/runner/work/h3/h3/src/apps/testapps/testGridRing.c:32:35: error: call to undeclared function 'gridRing'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   32 |         t_assertSuccess(H3_EXPORT(gridRing)(sfHex, 0, k0));
      |    
```
when the test file has:
```
#include "h3api.h"
```

also my first time writing in C so any help is appreciated!

isaacbrodsky:
> Not quite sure why were getting these errors
> 
> ```
> /home/runner/work/h3/h3/src/apps/testapps/testGridRing.c:32:35: error: call to undeclared function 'gridRing'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
>    32 |         t_assertSuccess(H3_EXPORT(gridRing)(sfHex, 0, k0));
>       |    
> ```
> 
> when the test file has:
> 
> ```
> #include "h3api.h"
> ```
> 
> also my first time writing in C so any help is appreciated!

The issue is because the declaration needs to be added to `h3api.h.in` (which gets processed into `h3api.h`), otherwise the compiler isn't sure what that function looks like.

isaacbrodsky:
Thanks for this contribution! I will take a look in more detail.

If you're comfortable, could you also add this function to `fuzzerGridDisk` so we get the full fuzzer coverage? (We can also add that in a follow up)

justinhwang:
> The issue is because the declaration needs to be added to `h3api.h.in` (which gets processed into `h3api.h`), otherwise the compiler isn't sure what that function looks like.

Ah I didn't notice that `src/h3lib/include/h3api.h` was local and in `.gitignore`, thanks!

> Thanks for this contribution! I will take a look in more detail.
> 
> If you're comfortable, could you also add this function to `fuzzerGridDisk` so we get the full fuzzer coverage? (We can also add that in a follow up)

Would prefer if someone could help me do that in a followup, thank you :)

src/apps/testapps/testH3Memory.c
```diff
@@ -128,6 +128,38 @@ SUITE(h3Memory) {
         free(gridDiskOutput);
     }
 
+    TEST(gridRing) {
+        const int k = 2;
+        H3Index *gridRingOutput = calloc(6 * k, sizeof(H3Index));
```
dfellis: My only comment is that it would be best if we also add a `maxGridRingSize` function to perform this calculation for users to inform the amount of memory to allocate. Other than that this looks good to me, thank you!

justinhwang: Yeah that's fair - it wasn't entirely clear to me whether the max ring size is necessarily `k == 0 ? 1 : 6 * k` when we encounter a pentagon when I was writing the code, but it would seem that we usually have fewer than `6 * k` elements (as far as the tests have indicated to me). Could you confirm that `6 * k` is the max?

dfellis: `6 * k` is the maximum, when the origin of the ring is a hexagon and all rings before the current ring do not contain a pentagon. When there is a pentagon then the ring becomes distorted, but always in a way where there are slightly fewer cells to allocate for, so the `max` function can ignore that situation.

justinhwang: ack yeah totally makes sense, didn't want to cause any mem errors but I'll add in the `maxGridRingSize` func!

justinhwang: Added!

src/apps/filters/h3.c
```diff
@@ -642,14 +642,18 @@ SUBCOMMAND(gridRing,
...
+        exit(1);
+    }
     H3Index *out = calloc(len, sizeof(H3Index));
     if (out == NULL) {
         fprintf(stderr, "Failed to allocate memory for the output H3 indexes");
         exit(1);
     }
-    H3Error err = H3_EXPORT(gridRingUnsafe)(cell, k, out);
+    err = H3_EXPORT(gridRingUnsafe)(cell, k, out);
     if (err) {
```
justinhwang, line 657: Just now noticing this implementation here @dfellis 😆 we've done it basically the same way

isaacbrodsky (approved):

PR #1020, opened 2025-06-23
vertex.c: mark an unreachable case as NEVER

isaacbrodsky:
From reviewing the coverage report and code, this should never be possible


nrabinowitz (approved):
LGTM, I believe you're correct here

dfellis (approved):

ajfriend (approved):

PR #1023, opened 2025-06-29
H3index always unsigned

isaacbrodsky:
Cover a few more additional lines / branches, either via tests or by NEVER/ALWAYS type macros.

### DIFF
--- a/src/h3lib/lib/h3Index.c
+++ b/src/h3lib/lib/h3Index.c
@@ -61,7 +61,9 @@ static char *H3ErrorDescriptions[] = {
  * @return The string describing the H3Error
  */
 const char *H3_EXPORT(describeH3Error)(H3Error err) {
-    if (err >= 0 && err <= 15) {  // TODO: Better way to bounds check here?
+    // err is always non-negative because it is an unsigned integer
+    // TODO: Better way to bounds check here?
+    if (ALWAYS(err >= 0) && err <= 15) {
         return H3ErrorDescriptions[err];
     } else {
         return "Invalid error code";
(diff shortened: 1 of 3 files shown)


src/h3lib/lib/h3Index.c
```diff
@@ -61,7 +61,9 @@ static char *H3ErrorDescriptions[] = {
  * @return The string describing the H3Error
  */
 const char *H3_EXPORT(describeH3Error)(H3Error err) {
-    if (err >= 0 && err <= 15) {  // TODO: Better way to bounds check here?
+    // err is always non-negative because it is an unsigned integer
+    // TODO: Better way to bounds check here?
+    if (ALWAYS(err >= 0) && err <= 15) {
```
dfellis: Do we need the negative number bounds check at all, then? It's literally impossible to assign a negative number to an `H3Error`, so I would consider that bounds check a silly mistake and drop it entirely.

isaacbrodsky: That's fair -- we may as well just remove it.

src/h3lib/lib/vertex.c
```diff
@@ -71,11 +71,13 @@ static H3Error vertexRotations(H3Index cell, int *out) {
...
         int p = 0;
-        for (; ALWAYS(p < NUM_PENTAGONS); p++) {
+        // Don't use a for loop here, for coverage reasons.
+        while (ALWAYS(p < NUM_PENTAGONS)) {
             if (pentagonDirectionFaces[p].baseCell == baseCell) {
                 dirFaces = pentagonDirectionFaces[p];
                 break;
             }
+            p++;
         }
```
isaacbrodsky, line 81: This may be specific to my compiler / lcov version, but should be inconsequential

Issue #1022, opened 2025-06-28
Expose a `getIndexDigit()` function in public API


ajfriend:
We currently have a few functions in the public API to get components of an H3 cell: `getResolution` and `getBaseCellNumber`. However, we are missing a function to retrieve the index digits of a cell. Internal to the library, there are macros for all three:

|        macro         |      API function     |
|----------------------|-----------------------|
| `H3_GET_RESOLUTION`  | `getResolution()`     |
| `H3_GET_BASE_CELL`   | `getBaseCellNumber()` |
| `H3_GET_INDEX_DIGIT` | *PROPOSED*          |

What do folks think about adding a function to the public API that would correspond to the `H3_GET_INDEX_DIGIT` macro? It would be useful for users who want to inspect the structure of a given cell.

It might have a signature like:

```c
int getIndexDigit(H3Index h, int res);
```

Naming might alternatively be `getCellDigit`.

Compare to the existing signatures:

```c
int getBaseCellNumber(H3Index h);
int getResolution(H3Index h);
```

ajfriend:
Also, I've had plenty of occasions where I've wanted to create a cell from the base cell, res, and digit information, so how about a function like

```c
H3Error createCell(int res, int baseCellNumber, int* digits, H3Index *out);
```

ajfriend:
Alternatively, we might define a struct to make things a bit more clear. Something like

```c
typedef struct {
    int res;
    int baseCellNumber;
    int digits[15];
} CellComponents;
```

with functions like

```c
H3Error createCell(CellComponents *comp, H3Index *out);
H3Error getCellComponents(H3Index h, CellComponents *out);
```

isaacbrodsky:
This seems like the old `H3IndexFat` structure that defined a cell as a struct rather than a 64 bit integer. 

What is the use case for retrieving a cell's full set of components, or specifying a cell by its components? Base cells and digits are fairly opaque to most users so I would be surprised if cells can be made by an application. Is the idea that cell components are being split up and stored in different parts of an application or dataset?

This also seems to connect with the idea of storing additional data in the unused `7` indexing digits. Perhaps functions that specifically support that would be helpful too. Edit: I will see if I can pull out some code for this.

ajfriend:
I've wanted something like this whenever I try prototyping an algo quickly in Python that deals with the H3 internals. I typically just code up a version with bit strings in Python, but it would be nice (and less error-prone) if I could just call a library function.

dfellis:
[...]

Issue #277, opened 2019-08-30
Exposing index inspection/manipulation macros/functions


zachasme:
I'm implementing (SP-)GiST index support for our PostgreSQL bindings and I find myself copying over macros for inspecting and manipulating index digits and resolution. I need these because I'm encoding each index digit as a branch in a tree, and reconstructing H3 indexes as the path from root to leaf node. 

Have you considered exposing some of the tools for low-level manipulation of H3 indexes as part of the public API or maybe in a separate header file?

dfellis:
Is there a problem with the `h3-pg` project depending on the `h3Index.h` file? I think everything you need is defined in that header file.

zachasme:
Yes those are exactly the things I need. The only problem was that I couldn't figure out a way to install `h3Index.h` alongside `h3api.h`, and I would like to support linking against H3 built as a shared library at some point.

isaacbrodsky:
To install `h3Index.h`, you can modify `CMakeLists.txt` before generating and `make install`ing:

Change (around line 609):
```
install(
    FILES "${CMAKE_CURRENT_BINARY_DIR}/src/h3lib/include/h3api.h"
    DESTINATION "${include_install_dir}/h3"
)
```

To:
```
install(
    FILES "${CMAKE_CURRENT_BINARY_DIR}/src/h3lib/include/h3api.h" "${CMAKE_CURRENT_SOURCE_DIR}/src/h3lib/include/h3Index.h"
    DESTINATION "${include_install_dir}/h3"
)
```

I don't think that linking to H3 as a shared library should have an impact on using the macros or using internal H3 functions.

dfellis:
Should that be a default, then @isaacbrodsky ?

zachasme:
Unless it is on by default I cannot assume existence of `h3Index.h` even if `h3` is already installed on the system.

For example on Arch Linux `h3` might have been installed manually or using [the aur package](https://aur.archlinux.org/packages/h3-git/) but I will have to instruct users to rebuild it using the changed `CMakeLists`.

I fully understand if it doesn't make sense for you to expose that file by default in it's current form.

isaacbrodsky:
Only `h3api.h` was installed because that's the public API. I think it would be fine to add the internal headers with clarity that they're internal. The drawback of this is that different library versions may break your build. Perhaps we could initially add something like:

```
# (First, define internal_header_files)
install(
    FILES ${internal_header_files}
    DESTINATION "${include_install_dir}/h3/internal"
)
[...]

PR #1037, opened 2025-08-16
Create cell from components

ajfriend:
Follow up from https://github.com/uber/h3/issues/1022: provides a function to create an H3 cell from (resolution, base cell number, and digit) components. 

Relevant to https://github.com/uber/h3/issues/277

### DIFF
--- a/src/h3lib/lib/h3Index.c
+++ b/src/h3lib/lib/h3Index.c
@@ -38,6 +38,11 @@ static const bool isBaseCellPentagonArr[128] = {
     [4] = 1,  [14] = 1, [24] = 1, [38] = 1, [49] = 1,  [58] = 1,
     [63] = 1, [72] = 1, [83] = 1, [97] = 1, [107] = 1, [117] = 1};
 
+// TODO: https://github.com/uber/h3/issues/984
+static const bool isBaseCellPentagonArr[128] = {
+    [4] = 1,  [14] = 1, [24] = 1, [38] = 1, [49] = 1,  [58] = 1,
+    [63] = 1, [72] = 1, [83] = 1, [97] = 1, [107] = 1, [117] = 1};
+
 /** @var H3ErrorDescriptions
  *  @brief An array of strings describing each of the H3ErrorCodes enum values
  */


ajfriend:
Open questions:

- Better function name?
- How much validation do we want to do? I'm doing a few basic range checks on res, base cell, and digits, but do we also want to call `isValidCell` to ensure the output is valid? The downside would be the additional overhead of the check, and I expect that most users of this function will be creating cells from component data that they already know would produce a valid cell.
    - One potential use case for this function would be to create **invalid** cells (like deleted subsequences) for us to use in tests elsewhere, so we'd want this function (or maybe an internal-only version) to be able to create such cells. Thus, one option would be to restrict the domains of the inputs for this function to only those that are representable in the H3Index bit representation, even if it results in an invalid cell.
- Errors: we have `E_RES_DOMAIN`, but the most specific domain error for base cell or digit being out of range is `E_DOMAIN`. Maybe we should create additional, more specific errors for these two cases.

ajfriend:
cc: @zachasme

zachasme:
This would definitely cover my use-case! :+1:

src/h3lib/lib/h3Index.c
```diff
@@ -112,6 +112,38 @@ H3Error H3_EXPORT(getIndexDigit)(H3Index h, int res, int *out) {
...
+            return E_DOMAIN;
+        }
+        H3_SET_INDEX_DIGIT(h, r, d);
+    }
+
+    // Optional: isValidCells is a more expensive test. do we want to run it
+    // every time?
+    // if (!isValidCell(h)) {
+    //     return E_CELL_INVALID;
+    // }
```
isaacbrodsky: My preference would be to allow users to specify invalid indexes if desired, as that could be used with getIndexDigit

isaacbrodsky: If you want to only generate valid cells, I think the only missing check is for the pentagon deleted subequence, so the full call to isValidCell might not be needed

ajfriend: Yeah, I like the idea of this function only returning valid cells. I've reworked the code towards that.

ajfriend:
Update:

- The function should only be able to create **valid** H3 cells that pass `isValidCell()`. All other malformed inputs should raise an error.
- Add a few new `H3Error` codes for more specific error information: `E_BASE_CELL_DOMAIN`, `E_DIGIT_DOMAIN`, `E_DELETED_DIGIT`

ajfriend:
Close in favor of #1063

PR #1042, opened 2025-09-01
Fix valid index bug in getIndexDigit

ajfriend:
The following logic in the `getIndexDigit` CLI doesn't correctly identify invalid cells/vertexes/edges:

https://github.com/uber/h3/blob/491eeadeeb3f3689413257f0087dc9330f5fe6ff/src/apps/filters/h3.c#L312-L317

This is because the `isValid*` functions return a boolean, not an `H3Error` code.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,19 @@
+init: purge
+    mkdir build
+
+build:
+    cd build; cmake -DCMAKE_BUILD_TYPE=Release ..; make
+
+purge:
+    rm -rf build
+
+test: build
+    # ./build/bin/testCreateCell
+    # just test-fast
+    ./build/bin/h3 getIndexDigit -c 5 -r 15
+
+test-fast: build
+    cd build; make test-fast
+
+test-slow: build
+    cd build; make test


ajfriend:
Alternatively, if we want to do this kind of validation, should we add it to the function directly? The downside would be the overhead of the check.

However, to me, it feels a little weird to have the extra validation in the CLI but not on the function itself. Or, does it make sense to do the extra validation for the CLI? Would it make sense to have the function and CLI be aligned in terms of how much validation they do?

justfile
```diff
@@ -0,0 +1,23 @@
+init: purge
+    mkdir build
+
+build:
+    cd build; cmake -DCMAKE_BUILD_TYPE=Release ..; make
+
+purge:
+    rm -rf build
+
+test: build
```
isaacbrodsky: Is this file intended to be checked in? The test commands seem specific to this change. If not, maybe gitignore it?

src/apps/filters/h3.c
```diff
@@ -309,11 +309,11 @@ SUBCOMMAND(getIndexDigit,
     Arg *args[] = {&getIndexDigitArg, &helpArg, &cellArg, &digitArg};
     PARSE_SUBCOMMAND(argc, argv, args);
     // TODO: Should there be a general `isValidIndex`?
-    H3Error cellErr = H3_EXPORT(isValidCell)(cell);
-    H3Error edgeErr = H3_EXPORT(isValidDirectedEdge)(cell);
-    H3Error vertErr = H3_EXPORT(isValidVertex)(cell);
-    if (cellErr && edgeErr && vertErr) {
-        return cellErr;
+    bool is_valid = H3_EXPORT(isValidCell)(cell) ||
```
isaacbrodsky: Camel case?

justfile
ajfriend: I'll remove it before landing.

I keep it in the git history because I often find myself referring to these files in old PRs after I haven't touched the repo in a while and need to remember how to do things. I find it is a good historical reference of what commands I was actually running at the time.

I also get a lot of value while working on a PR from seeing the diffs of this file over time.

Ideas on better ways to organize these somewhat personalized tooling files?

src/apps/filters/h3.c
ajfriend: done.

justfile
ajfriend: Also, I can ask questions like this: @isaacbrodsky, is this how you typically run tests in the repo? Do you tend to run `ctest` directly or commands like `make test-fast`? Do you know of a way to run just the CLI tests, or just one CLI test? Currently, `ctest` runs both the C and CLI tests.

:)

isaacbrodsky (approved):
Please remove `justfile` before merging

justfile
isaacbrodsky: I usually run `make test` or `make test-fast`. When debugging in more detail I would switch to `ctest` since that provides more direct control and feedback about verbosity, print-on-error, etc. When debugging a specific failure I would then go to running the test binary directly, since I may need to run it under valgrind or lldb/gdb or something.

For the CLI tests I think I mainly run them through `make test`. For running a specific CLI test I would probably copy out the command from the applicable CMake file.

I don't have a recommendation on the specific commands ran. I feel this would be personal unless you document more about what the commands are and why someone would want to run them. For that I would either gitignore or keep in another branch or repo (as with dotfile repos). Unfortunately that loses the association with which uber/h3 commit they go. You could then try to use submodules to take care of that but my experience is most people find submodules challenging to work with.

ajfriend: Thanks. For the `justfile`, I'll probably just stick to my plan of keeping it in the git history while developing, removing it before merging, and then squash merge at the end to keep the history clean.

src/apps/filters/h3.c
```diff
@@ -309,11 +309,11 @@ SUBCOMMAND(getIndexDigit,
     Arg *args[] = {&getIndexDigitArg, &helpArg, &cellArg, &digitArg};
     PARSE_SUBCOMMAND(argc, argv, args);
     // TODO: Should there be a general `isValidIndex`?
-    H3Error cellErr = H3_EXPORT(isValidCell)(cell);
-    H3Error edgeErr = H3_EXPORT(isValidDirectedEdge)(cell);
-    H3Error vertErr = H3_EXPORT(isValidVertex)(cell);
-    if (cellErr && edgeErr && vertErr) {
-        return cellErr;
+    bool isValid = H3_EXPORT(isValidCell)(cell) ||
```
dfellis, line 312: I see how I made this mistake. In most of the code the error condition is the "truthy" value, but here success is `true`, though I named these variables as if they were `H3Error`s. That feels like a point against [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation).

dfellis (approved):

ajfriend:
Link to the `justfile` for future reference:  

- https://github.com/uber/h3/blob/0243a2c0c503f2b7ce0790f3e3cb279dea5f6345/justfile
- https://gist.github.com/ajfriend/1b8f07bacb046a4ef0fcea51d5b99530

justfile
justinhwang: Could add it to .gitignore and then keep it in your local?

ajfriend: I like to keep it in the git history to see how it changes as I work on the diff, and also to be able to reference it later when I come back in a few months and I forget how to build everything. Removing at the last commit and then squashing (to remove it from the repo history) seems like a good compromise, but always open to other ideas!

src/apps/filters/h3.c
```diff
@@ -309,11 +309,11 @@ SUBCOMMAND(getIndexDigit,
     Arg *args[] = {&getIndexDigitArg, &helpArg, &cellArg, &digitArg};
```
LucaDiba, line 309: @ajfriend should lines 265-268, 285-288, 391-394, and 420-423 be updated as well?

justinhwang: Yeah, fixed in #1055

ajfriend: Ah, good catch @LucaDiba! And thanks for the fix @justinhwang!

Issue #1053, opened 2025-10-06
Hierarchical inconsistency between res 0→1 and res 1→2→3→4→5


benguild:
Hi all, please let me know if this is not the correct interpretation or if it's actually not broken and I'm missing something, but it seems like any situation assuming "latLngToCell(lat, lng, N)" is always the parent of "latLngToCell(lat, lng, N+1)" can break in certain cases, as far as I can tell. If this is verifiable, it seems like kind of a showstopper. (…unless it's intended behavior, which does not seem correct from my current understanding of how this should work)

## Demo

```c
#include <h3/h3api.h>
#include <stdio.h>
#include <stdbool.h>

void testPoint(const char* name, double lat, double lng) {
    LatLng coord = {.lat = lat, .lng = lng};

    H3Index res0, res1, res2;
    latLngToCell(&coord, 0, &res0);
    latLngToCell(&coord, 1, &res1);
    latLngToCell(&coord, 2, &res2);

    printf("Testing: %s (%.6f, %.6f)\\n", name, lat, lng);
    printf("  LatLngToCell results:\\n");
    printf("    Res 0: %llx\\n", (unsigned long long)res0);
    printf("    Res 1: %llx\\n", (unsigned long long)res1);
    printf("    Res 2: %llx\\n", (unsigned long long)res2);

    H3Index res1Parent, res2Parent;
    cellToParent(res1, 0, &res1Parent);
    cellToParent(res2, 1, &res2Parent);

    bool res0Match = (res1Parent == res0);
    bool res1Match = (res2Parent == res1);

    printf("  Hierarchy check:\\n");
    printf("    Res 1 parent == Res 0 LatLngToCell: %s", res0Match ? "true" : "false");
    if (res0Match) {
        printf(" ✅\\n");
    } else {
        printf(" ❌ (parent is %llx)\\n", (unsigned long long)res1Parent);
    }

    printf("    Res 2 parent == Res 1 LatLngToCell: %s", res1Match ? "true" : "false");
    if (res1Match) {
        printf(" ✅\\n");
    } else {
        printf(" ❌ (parent is %llx)\\n", (unsigned long long)res2Parent);
    }

    int64_t childrenSize;
    cellToChildrenSize(res0, 1, &childrenSize);
    H3Index* children = malloc(childrenSize * sizeof(H3Index));
    cellToChildren(res0, 1, children);

    bool found = false;
    for (int64_t i = 0; i < childrenSize; i++) {
        if (children[i] == res1) {
[...]

Issue #1030, opened 2025-07-25
Potential inconsistent traversal of cell hierarchy


iw-MikeS:
I have been using h3 to assign positions to geographies following an approach of allocation of resolution 0,3,6,9,12,15 to positions set directly using latlng_to_cell, I calculated zones using 

```
poly = h3.LatLngPoly(outer)

cells = h3.h3shape_to_cells_experimental(poly, res=resolution, contain='overlap')

x = h3.compact_cells(cells)
```

After compaction I am getting the children of 1,2,4,5,7,8,10,11,13,14 and storing all the x3 cells

When I have then looked up positions in zones I am finding some positions are not matching a cell in the shape. 

In trying to debug this I wrote the following and noticed an inconsistency in the way traversal sets the cell. 

```
%pip install h3

import h3

lat = 51.64604
lon = 7.09634

x = h3.latlng_to_cell(lat, lon, 9)

print(x + ' Original res of 9')

print ('------------------------------------------------')
x8 = h3.cell_to_parent(x)
print(x8 + ' res 8 as parent')
xx8 = h3.latlng_to_cell(lat, lon, 8)
print(xx8 + ' GET RES 8')
print ('------------------------------------------------')
x7 = h3.cell_to_parent(x8)
print(x7 + ' res 7 as parent')
xx7 = h3.latlng_to_cell(lat, lon, 7)
print(xx7 + ' GET RES 7')
print ('------------------------------------------------')
x6 = h3.cell_to_parent(x7)
print(x6 + ' res 6 as parent')
xx6 = h3.latlng_to_cell(lat, lon, 6)
print(xx6 + ' GET RES 6')
print ('------------------------------------------------')
x5 = h3.cell_to_parent(x6)
print(x5 + ' res 5 as parent')
xx5 = h3.latlng_to_cell(lat, lon, 5)
print(xx5 + ' GET RES 5')
print ('------------------------------------------------')
x4 = h3.cell_to_parent(x5)
print(x4 + ' res 4 as parent')
xx4 = h3.latlng_to_cell(lat, lon, 4)
print(xx4 + ' GET RES 4')
print ('------------------------------------------------')
x3 = h3.cell_to_parent(x4)
print(x3 + ' res 3 as parent')
xx3 = h3.latlng_to_cell(lat, lon, 3)
print(xx3 + ' GET RES 3 INCONSISTENT')
print ('------------------------------------------------')
x2 = h3.cell_to_parent(x3)
[...]

Issue #973, opened 2025-02-24
cellToParent result is inaccurate


jnissitta2204:
I am working on a usecase, where I have to do find out parent H3 of a cell for the optimization of point-in-polygon problem.
I used the cellToParent() function to get the parent H3 cell-id, and the returned parent H3 cell is not covering/overlapping the child H3 cell-id.

Here, is the simulation of the problem:

```
extern "C" {
  #include "h3/h3api.h"
}
#include <iostream>

int main(){
    //Consider this point
    double longitude = -73.99158477783203;
    double latitude = 40.76464080810547;
    LatLng a_point;
    H3Index h3_point;
    H3Index parent_H3_Cell;
    char point_as_h3_str[20];
    char parent_h3_string[20];

    //Converting them to radians
    a_point.lng = degsToRads(longitude);
    a_point.lat = degsToRads(latitude);

    //Given Point at resolution 7
    std::cout << "The given point is : " << std::endl;
    std::cout << "Longitude: " << longitude << std::endl;
    std::cout << "Latitude: " << latitude << std::endl;

    // H3 cell-id of the point at resolution 7
    H3Error h3_res = latLngToCell(&a_point, 7, &h3_point);
    h3ToString(h3_point, point_as_h3_str, 20);

    //compute parent
    cellToParent(h3_point, 4, &parent_H3_Cell);
    h3ToString(parent_H3_Cell, parent_h3_string, 20);

    
    std::cout << "The h3 index of point at resolution 7 : " << std::string(point_as_h3_str) << std::endl;
    std::cout << "The parent h3 index of point at resolution 4 " << std::string(parent_h3_string) << std::endl;
}
```

And, the h3 cells visualized:

<img width="738" alt="Image" src="https://github.com/user-attachments/assets/889c4e61-e26a-42bb-8908-1a98ead2a355" />

H3 Version : 4.1.0

Is this behavior expected?

isaacbrodsky:
This is an expected property of the H3 grid. Child (finer resolution) cells are not necessarily geographically contained by their parent (coarser) cell. 

Some resources about this: https://observablehq.com/@nrabinowitz/h3-hierarchical-non-containment?collection=@nrabinowitz/h3 and https://h3geo.org/docs/highlights/indexing

jnissitta2204:
I am using H3 for as a preliminary check for a point-in-polygon problem.
In the above mentioned case, the h3 check will indicate that the point does not lie in polygon (false negative).
[...]

PR #1054, opened 2025-10-06
Updating "README.md" to avoid confusion

benguild:
#1053 #1030 #973 etc.

### DIFF
--- a/website/docs/README.md
+++ b/website/docs/README.md
@@ -13,7 +13,7 @@ The H3 Core Library implements the H3 grid system. It includes functions for con
 
 ## Highlights
 
-* H3 is a [hierarchical geospatial index](https://github.com/docs/highlights/indexing), which provides [exact logical containment but only approximate geometric containment](https://observablehq.com/@nrabinowitz/h3-hierarchical-non-containment) across the cell hierarchy. That is, **points within children may have different parent cells,** because child cells are not always entirely geometrically contained by the same parent.
+* H3 is a [hierarchical geospatial index](/docs/highlights/indexing), which provides [exact logical containment but only approximate geometric containment](https://observablehq.com/@nrabinowitz/h3-hierarchical-non-containment) across the cell hierarchy. That is, **points within children may have different parent cells,** because child cells are not always entirely geometrically contained by the same parent.
 * H3 was developed to address the [challenges of Uber's data science needs](/docs/highlights/aggregation).
 * H3 can be used to [join disparate data sets](/docs/highlights/joining).
 * In addition to the benefits of the hexagonal grid shape, H3 includes features for [modeling flow](/docs/highlights/flowmodel).


ajfriend:
Thanks for this! I'm trying to think through the best way to represent this in the docs. What about adding a note about this under the hierarchical bullet in the Highlights section, and then linking to the docs for more details, including Nicks Observable page. I'm also wondering if the `/docs/highlights/indexing` page might be better named as something like `/docs/highlights/hierarchical_containment`. What do folks think? @isaacbrodsky @dfellis @nrabinowitz 

Maybe something like:


## Highlights

* H3 is a [hierarchical geospatial index](/docs/highlights/indexing), providing **exact logical** containment and **approximate geometric** containment across the cell hierarchy.

benguild:
> * H3 is a [hierarchical geospatial index](/docs/highlights/indexing), providing **exact logical** containment and **approximate geometric** containment across the cell hierarchy.

That's not clear enough, honestly. You could say that but it's probably best to reinforce it with an even simpler explanation of the limitations afterward… in that the children are not an exact fit to the parent, and vice versa, etc.

As noted in my comment https://github.com/uber/h3/issues/1053#issuecomment-3372729513, this should really be introduced as a limitation and clarification as early in the discovery process as possible.

ajfriend:
I agree this should be as obvious as quickly as possible. I'm just trying to craft a good, short warning up front that would fit well in an readme page, that we can then expand on immediately in the linked page. And we may want to also update the linked page.

How about:

## Highlights

* H3 is a [hierarchical geospatial index](/docs/highlights/indexing), providing **exact logical** containment across the cell hierarchy, and **approximate geometric** containment (e.g., children are not entirely contained in their parents).

isaacbrodsky:
> I agree this should be as obvious as quickly as possible. I'm just trying to craft a good, short warning up front that would fit well in an readme page, that we can then expand on immediately in the linked page. And we may want to also update the linked page.
> 
> How about:
> ## Highlights
> 
>     * H3 is a [hierarchical geospatial index](/docs/highlights/indexing), providing **exact logical** containment across the cell hierarchy, and **approximate geometric** containment (e.g., children are not entirely contained in their parents).

This looks good but I think linking "approximate geometric" to Nick's Observable may be helpful. That phrase may not be immediately obvious what it means, and having a graphic example for it could be very informative.

We definitely also want to improve the docs pages :)

benguild:
> * H3 is a [hierarchical geospatial index](/docs/highlights/indexing), providing **exact logical** containment across the cell hierarchy, and **approximate geometric** containment (e.g., children are not entirely contained in their parents).

How about something like:
- H3 is a [hierarchical geospatial index](/docs/highlights/indexing), which provides **exact logical** containment but **approximate geometric** containment across the cell hierarchy. **⚠️ Children are not entirely contained by their parent cells, and points within them may in fact map to adjacent parents instead.**

ajfriend:
Based off the last few comments, how about something like this for the intro/readme page:

- H3 is a [hierarchical geospatial index](/docs/highlights/indexing), which provides [**exact logical** containment but only **approximate geometric** containment](https://observablehq.com/@nrabinowitz/h3-hierarchical-non-containment) across the cell hierarchy. **That is, child cells are not always geometrically contained by their parent cells, so points within them may map to other (adjacent) parent cells.**

That being said, I still think @benguild is correct in that we should document this better (and improve upon the meager attempt above), as it seems to be a common gotcha for new users coming to the library. I'd be in favor of a new **top-level** docs page on "hierarchical containment" that makes it vividly clear to new users what the potential problems might be, while also explaining how to avoid or work with them. Alternatively, we can rename/rework the https://h3geo.org/docs/highlights/indexing page to make this issue more obvious.

What do folks think? I can create a task for it and give it a stab.

isaacbrodsky:
> Based off the last few comments, how about something like this for the intro/readme page:
> 
>     * H3 is a [hierarchical geospatial index](/docs/highlights/indexing), which provides [**exact logical** containment but only **approximate geometric** containment](https://observablehq.com/@nrabinowitz/h3-hierarchical-non-containment) across the cell hierarchy. **That is, child cells are not always geometrically contained by their parent cells, so points within them may map to other (adjacent) parent cells.**
> 

This text seems overall good to me. I think the word "map" may be a little unclear (what does "map to other ... cells" mean?) Maybe "be logically contained by other (adjacent) parent cells"?

> 
> That being said, I still think @benguild is correct in that we should document this better (and improve upon the meager attempt above), as it seems to be a common gotcha for new users coming to the library. I'd be in favor of a new **top-level** docs page on "hierarchical containment" that makes it vividly clear to new users what the potential problems might be, while also explaining how to avoid or work with them. Alternatively, we can rename/rework the https://h3geo.org/docs/highlights/indexing page to make this issue more obvious.
> 
> What do folks think? I can create a task for it and give it a stab.

I would love to see more and improved docs pages to guide new users through H3! I find we frequently direct people to @nrabinowitz 's great notebook visualizations to explain things, so it would be great to have all of that in one place.

ajfriend:
Good point on "map". How about:

- H3 is a [hierarchical geospatial index](https://github.com/docs/highlights/indexing), which provides [exact logical containment but only approximate geometric containment](https://observablehq.com/@nrabinowitz/h3-hierarchical-non-containment) across the cell hierarchy. That is, child cells are not always geometrically contained by their parent cells, so points within children may belong to other (adjacent) parent cells.

isaacbrodsky:
> Good point on "map". How about:
> 
> 
> 
> - H3 is a [hierarchical geospatial index](https://github.com/docs/highlights/indexing), which provides [exact logical containment but only approximate geometric containment](https://observablehq.com/@nrabinowitz/h3-hierarchical-non-containment) across the cell hierarchy. That is, child cells are not always geometrically contained by their parent cells, so points within children may belong to other (adjacent) parent cells.

I think that's a good way to put it.

benguild:
I think this is important enough that it **should probably be its own bulletpoint**. Placing it in the second sentence frames it as a minor detail, when in fact people keep hitting this issue.

Simply put, it’s unusual that the child cells don’t fit the parents. While there are advantages in this design, you should highlight this as a key difference aside from the fact that this offers a "hierarchical geospatial index" when it's not strictly so. Anybody who looks at this for the first time needs to know that, and I'm concerned about the pattern of folks not knowing it.

I'm just trying to be fairly direct here, since the purpose of the PR is to highlight this key detail to avoid others falling into the same difference in expectation again. We can work on the language, but we should agree on whether it's 1 or 2 bulletpoints first.

ajfriend:
Honestly, I think that's splitting hairs. In the context of an intro page, the language above makes it quite clear, especially if we bold the second sentence (and we should), which calls it out better than making a new bullet, IMO.

I'd propose the language above, with the second sentence in bold.

We can see how that goes, and make further changes if necessary. But my guess is that further changes on the intro page will have less effect than a new top-level docs page on "hierarchical containment" that we can link to, and that's where we should focus our time.

ajfriend:
And here's a ticket for what I think is the more important docs issue: https://github.com/uber/h3/issues/1057

benguild:
> the language above makes it quite clear, especially if we bold the second sentence (and we should)

I think with the bolding… @ajfriend, @isaacbrodsky, and I are in reasonable agreement here. I tweaked the language in the diff based on folks' thoughts, and would suggest that latest version. Can revise further if folks think it's less clear.

ajfriend:
Looks great. And thanks again for raising! We're always interested in suggestions to improve the docs (even if there's some back and forth on how to do it 😀).

benguild:
No problem. Can merge today if there are enough reviews completed.

Issue #1044, opened 2025-09-03
Add E_INDEX_INVALID error?


ajfriend:
Related to #1043, should we add an error indicating that the input to a function is not a valid H3 index of any mode? Maybe `E_INDEX_INVALID`?

This error could be used, for example, in `getIndexDigit`. (https://github.com/uber/h3/pull/1042)

Issue #1043, opened 2025-09-03
Add isValidIndex function


ajfriend:
As in #1024 and #1042, should we add a general function to test if the input is a valid H3 index of any mode (cell, directed edge, vertex, etc.)?

The code would be something like:

```c
bool isValid = H3_EXPORT(isValidCell)(cell) ||
               H3_EXPORT(isValidDirectedEdge)(cell) ||
               H3_EXPORT(isValidVertex)(cell);
```

cnaples79:
I'm working on this!

PR #1055, opened 2025-10-09
fix: valid index in cli

justinhwang:
Similar to #1042

### DIFF
--- a/src/apps/filters/h3.c
+++ b/src/apps/filters/h3.c
@@ -262,11 +262,11 @@ SUBCOMMAND(getResolution, "Extracts the resolution (0 - 15) from the H3 cell") {
     Arg *args[] = {&getResolutionArg, &helpArg, &cellArg};
     PARSE_SUBCOMMAND(argc, argv, args);
     // TODO: Should there be a general `isValidIndex`?
-    H3Error cellErr = H3_EXPORT(isValidCell)(cell);
-    H3Error edgeErr = H3_EXPORT(isValidDirectedEdge)(cell);
-    H3Error vertErr = H3_EXPORT(isValidVertex)(cell);
-    if (cellErr && edgeErr && vertErr) {
-        return cellErr;
+    bool isValid = H3_EXPORT(isValidCell)(cell) ||
+                   H3_EXPORT(isValidDirectedEdge)(cell) ||
+                   H3_EXPORT(isValidVertex)(cell);
+    if (!isValid) {
+        return E_DOMAIN;  // TODO: maybe create a new E_INDEX_INVALID error?
     }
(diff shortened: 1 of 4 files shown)


src/apps/filters/h3.c
```diff
@@ -262,11 +262,11 @@ SUBCOMMAND(getResolution, "Extracts the resolution (0 - 15) from the H3 cell") {
...
-    H3Error cellErr = H3_EXPORT(isValidCell)(cell);
-    H3Error edgeErr = H3_EXPORT(isValidDirectedEdge)(cell);
-    H3Error vertErr = H3_EXPORT(isValidVertex)(cell);
-    if (cellErr && edgeErr && vertErr) {
-        return cellErr;
+    bool isValid = H3_EXPORT(isValidCell)(cell) ||
+                   H3_EXPORT(isValidDirectedEdge)(cell) ||
+                   H3_EXPORT(isValidVertex)(cell);
+    if (!isValid) {
+        return E_DOMAIN;  // TODO: maybe create a new E_INDEX_INVALID error?
```
dfellis, line 269: [Looking at the set of errors](https://github.com/uber/h3/blob/master/src/h3lib/include/h3api.h.in#L81-L105) I would probably just go with `E_FAILED`. You can see in the new tests that the error message given when these sub-commands are called incorrectly is confusing with the current text it returns.

[Looking at the text it can select from](https://github.com/uber/h3/blob/master/src/h3lib/lib/h3Index.c#L37-L55) that would still be sub-optimal and `E_CELL_INVALID`'s error message would be more legible, but it also wouldn't be correct in some situations, so I'm not so sure about it.

justinhwang: Should we follow up with a method that wraps this logic for us (3x valid checks) and add a generic index invalid error?

dfellis: I would like some feedback from @isaacbrodsky @nrabinowitz and @ajfriend but I do think a general `isValid` function and a new error code would be better.

isaacbrodsky: I think it would make sense to add an E_INVALID_INDEX (#1044) and associated helper function. I am also fine with this PR as I understand it corrects the check that the index is valid.

ajfriend: I'd be in favor of https://github.com/uber/h3/issues/1043 and https://github.com/uber/h3/issues/1044. Up to you, @justinhwang if you want to do that now, or if we should follow up on it.

Although, we'd should probably make those changes before the next release to minimize the number of times we modify the error codes.

justinhwang: Yeah feel free to merge this one, I can follow up with #1043 and #1044

PR #1065, opened 2025-10-15
add H3_ERROR_END

ajfriend:
Define `H3_ERROR_CODE_MAX` in `h3api.h.in` to define the range of errors. Keep it updated as we add new errors.

Idea: Use this to address @dfellis's comment here: https://github.com/uber/h3/pull/1063#discussion_r2429916733

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The public API of this library consists of the functions declared in file
 ### Added
 - Added `getIndexDigit` function for inspecting specific index digits of an H3 index. (#1024, #1042)
 - Added `isValidIndex` function (#1056)
+- Added `H3_ERROR_END` sentinal value for iterating through errors (#1065)
 
 ## [4.3.0] - 2025-06-17
 ### Added


ajfriend:
Alternatively, AI suggested something like this:

```c
typedef enum {
    E_SUCCESS = 0,
    E_FAILED,
    E_DOMAIN,
    /* ... */
    E_DELETED_DIGIT,
    E_COUNT  // sentinel for iteration and sizing
} H3ErrorCodes;

#define H3_ERROR_MAX (E_COUNT - 1)
```

src/h3lib/lib/h3Index.c
```diff
@@ -66,7 +66,7 @@ static char *H3ErrorDescriptions[] = {
 const char *H3_EXPORT(describeH3Error)(H3Error err) {
     // err is always non-negative because it is an unsigned integer
     // TODO: Better way to bounds check here?
```
justinhwang: Could probably clean up this todo

justinhwang:
> Alternatively, AI suggested something like this:
> 
> ```c
> typedef enum {
>     E_SUCCESS = 0,
>     E_FAILED,
>     E_DOMAIN,
>     /* ... */
>     E_DELETED_DIGIT,
>     E_COUNT  // sentinel for iteration and sizing
> } H3ErrorCodes;
> 
> #define H3_ERROR_MAX (E_COUNT - 1)
> ```

We could use E_COUNT directly then no? (I also had a similar idea when adding E_INDEX_INVALID but wasn't sure as then it can be referenced externally - though maybe people would want a similar constant)

ajfriend:
@isaacbrodsky, any thoughts on which approach you'd prefer?

ajfriend:
> > Alternatively, AI suggested something like this:
> > ```c
> > typedef enum {
> >     E_SUCCESS = 0,
> >     E_FAILED,
> >     E_DOMAIN,
> >     /* ... */
> >     E_DELETED_DIGIT,
> >     E_COUNT  // sentinel for iteration and sizing
> > } H3ErrorCodes;
> > 
> > #define H3_ERROR_MAX (E_COUNT - 1)
> > ```
> 
> We could use E_COUNT directly then no? (I also had a similar idea when adding E_INDEX_INVALID but wasn't sure as then it can be referenced externally - though maybe people would want a similar constant)

Yeah, I suppose we could do something as simple as:

```c
typedef enum {
    E_SUCCESS = 0,
    E_FAILED = 1,
    E_DOMAIN = 2,
    /* ... */
    E_OPTION_INVALID = 15,  // Mode or flags argument was not valid
    E_INDEX_INVALID = 16,    // `H3Index` argument was not valid
    H3_ERROR_SENTINEL  // sentinel for iteration and sizing
} H3ErrorCodes;
```

isaacbrodsky:
> @isaacbrodsky, any thoughts on which approach you'd prefer?

We do something similar here: https://github.com/uber/h3/blob/b29d8207174883bdfe2d2714402e4a9d65a8aeb3/src/h3lib/include/coordijk.h#L82 I think whichever way you define the max sentinel is fine. I do prefer having the values explicitly set in the `enum` declaration, because I expect the code will be read much more than it will be written, and so I want to make reading it as easy as possible.

dfellis:
> > > Alternatively, AI suggested something like this:
> > > ```c
> > > typedef enum {
> > >     E_SUCCESS = 0,
> > >     E_FAILED,
> > >     E_DOMAIN,
> > >     /* ... */
> > >     E_DELETED_DIGIT,
> > >     E_COUNT  // sentinel for iteration and sizing
> > > } H3ErrorCodes;
> > > 
> > > #define H3_ERROR_MAX (E_COUNT - 1)
> > > ```
> > 
> > 
> > We could use E_COUNT directly then no? (I also had a similar idea when adding E_INDEX_INVALID but wasn't sure as then it can be referenced externally - though maybe people would want a similar constant)
> 
> Yeah, I suppose we could do something as simple as:
> 
> ```c
> typedef enum {
>     E_SUCCESS = 0,
>     E_FAILED = 1,
>     E_DOMAIN = 2,
>     /* ... */
>     E_OPTION_INVALID = 15,  // Mode or flags argument was not valid
>     E_INDEX_INVALID = 16,    // `H3Index` argument was not valid
>     H3_ERROR_SENTINEL  // sentinel for iteration and sizing
> } H3ErrorCodes;
> ```

This last one looks the "best" to me because (1) it's obviously different from the rest because it doesn't get a number assigned and it doesn't start with `E_` and (2) it will automatically be updated when new error codes are added without requiring that the developer remembers to do so (I doubt it would happen if anyone involved with H3 right now would do so, but in a couple of years a new contributor might not know and we might not remember if we don't explicitly see the sentinel definition in the Github diff view).

It could be made even more explicit by putting an inline multi-line comment *above* the `H3_ERROR_SENTINEL` stating something like:

```
/**
 * End of H3 Error Codes. Below is the sentinel value used for
 * iteration and sizing of the enum. Do not add error codes below it.
 **/
```

And if we want to be extra clever, the enum could be named `H3_ERROR_LENGTH` and then we still `#define H3_ERROR_MAX (H3_ERROR_LENGTH - 1)` below it so we don't have to have the subtraction in the code where the constant is used.

src/h3lib/lib/h3Index.c
ajfriend: Addressed. thanks

PR #1074, opened 2025-10-30
fuzzer for constructCell

### LINKED ISSUE #1067: fuzz constructCell

isaacbrodsky:
Fixes #1067

### DIFF
--- a/src/apps/fuzzers/fuzzerConstructCell.c
+++ b/src/apps/fuzzers/fuzzerConstructCell.c
@@ -35,7 +35,12 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     }
     const inputArgs *args = (const inputArgs *)data;
 
-    int resForAlloc = min(max(args->res, 0), MAX_H3_RES);
+    int resForAlloc = args->res;
+    if (resForAlloc < 0) {
+        resForAlloc = 0;
+    } else if (resForAlloc > MAX_H3_RES) {
+        resForAlloc = MAX_H3_RES;
+    }
     int *digits = calloc(resForAlloc, sizeof(int));
     memcpy(digits, args->digits, sizeof(int) * resForAlloc);
 


src/apps/fuzzers/fuzzerConstructCell.c
```diff
@@ -0,0 +1,49 @@
...
+    }
+    const inputArgs *args = (const inputArgs *)data;
+
+    H3Index out;
+    H3_EXPORT(constructCell)
+    (args->res, args->baseCellNumber, args->digits, &out);
+    if (args->res == 0) {
+        H3_EXPORT(constructCell)
+        (args->res, args->baseCellNumber, NULL, &out);
+    }
```
ajfriend, line 58: Just a (probably naive) question to improve my understanding of how to set up the fuzzers: I'm surprised we run an additional test if `res == 0`, as opposed to branching to a separate test on the value of `res`.

How does that work / why do we want to run both?

How would this compare to code like:

```c
if (args->res == 0) {
    H3_EXPORT(constructCell)
    (args->res, args->baseCellNumber, NULL, &out);
} else {
    H3_EXPORT(constructCell)
    (args->res, args->baseCellNumber, args->digits, &out);
}
```

Or, is the idea that the `res == 0` branch provides a more targeted test that's more likely to find issues quickly because it is reducing the search space?

ajfriend (approved):

src/apps/fuzzers/fuzzerConstructCell.c
isaacbrodsky: I think if res is 0 and a non-NULL pointer is passed for digits it should still be accepted. The inverse is not true however, since the caller must provide an array of length `res`.

isaacbrodsky: I'm updating this to be clearer about why we are testing different `digits` lengths.

ajfriend:
Makes sense, and this example helped me understand fuzzing a bit better :)

src/apps/fuzzers/fuzzerConstructCell.c
```diff
@@ -0,0 +1,65 @@
...
+        resForAlloc = 0;
+    } else if (resForAlloc > MAX_H3_RES) {
+        resForAlloc = MAX_H3_RES;
+    }
+    int *digits = calloc(resForAlloc, sizeof(int));
+    memcpy(digits, args->digits, sizeof(int) * resForAlloc);
+
+    H3Index out;
+    H3_EXPORT(constructCell)
+    (args->res, args->baseCellNumber, args->digits, &out);
```
nrabinowitz, line 49: Maybe this is just auto-formatting, but I find this hard to read compared with opening the parens on the previous line

isaacbrodsky: This is unfortunately how the auto-formatter wants to do it

dfellis (approved):

PR #1078, opened 2025-11-01
Add constructCell to CLI

LucaDiba:
- Add `constructCell` to CLI.
- Add tests.
- Update docs.

### DIFF
--- a/tests/cli/constructCell.txt
+++ b/tests/cli/constructCell.txt
@@ -2,6 +2,9 @@ add_h3_cli_test(
     testCliConstructCellRes0
     "constructCell -r 0 -b 0 -f newline" "8001fffffffffff")
 add_h3_cli_test(
+    testCliConstructCellRes0WithoutResolutionArg
+    "constructCell -b 0 -f newline" "8001fffffffffff")
+add_h3_cli_test(
     testCliConstructCellRes0Json
     "constructCell -r 0 -b 0 -f json" "\"8001fffffffffff\"")
 add_h3_cli_test(


src/apps/filters/h3.c
```diff
@@ -312,6 +312,89 @@ SUBCOMMAND(getIndexDigit,
...
+                 .value = &bc,
+                 .helpText = "Base cell number, 0-121 inclusive."};
+    Arg digitsArg = {.names = {"-d", "--digits"},
+                     .required = false,
+                     .scanFormat = "%1023c",
+                     .valueName = "d0,d1,...",
+                     .value = &digitsStr,
+                     .helpText =
+                         "Comma-separated list of child digits (0-6) of length "
+                         "resolution. Ignored if resolution is 0."};
```
LucaDiba: Is it fine for this to be CSV?

ajfriend: I think so. That matches what we do for compact and uncompact: https://h3geo.org/docs/api/hierarchy#compactcells

@dfellis ?

ajfriend:
This is great. Thanks @LucaDiba!

Would you also be able to add a short description for the CLI in the docs (which I skipped in https://github.com/uber/h3/pull/1077, because I wasn't sure what the final API would look like)?

LucaDiba:
> Would you also be able to add a short description for the CLI in the docs (which I skipped in #1077, because I wasn't sure what the final API would look like)?

Yup, updated.

src/apps/filters/h3.c
```diff
@@ -312,6 +312,89 @@ SUBCOMMAND(getIndexDigit,
...
+    int digits[MAX_H3_RES + 1] = {0};
+    if (res > 0) {
+        if (!digitsArg.found) {
+            return E_DIGIT_DOMAIN;
+        }
+
+        int count = 0;
+        char *p = digitsStr;
+        while (*p && count < MAX_H3_RES + 1) {
+            while (*p == ',') {
```
LucaDiba: Here `1,,,,2,3` would be valid. Do we want stricter validation and fail in this case?

ajfriend: Yeah, I think requiring that a digit be present makes sense.

ajfriend:
Here's a thought: What about making the resolution parameter optional? We can infer the res from the length of the digits.

I'm raising it here because I'm considering doing something like that for the Python bindings. 

Another solution would be to not have a resolution parameter at all, and just read from the length of the digits. That's not so different from what, say, `compactCells` does in the CLI; we don't pass in the number of cells. The res parameter is redundant if you have another way to know the length of the list.

dfellis:
> Here's a thought: What about making the resolution parameter optional? We can infer the res from the length of the digits.
> 
> I'm raising it here because I'm considering doing something like that for the Python bindings. 
> 
> Another solution would be to not have a resolution parameter at all, and just read from the length of the digits. That's not so different from what, say, `compactCells` does in the CLI; we don't pass in the number of cells. The res parameter is redundant if you have another way to know the length of the list.

I would be fine with optional, as long as an explicit mismatch results in an error. It can be useful to have it automatically inferred when using it manually, but including the resolution inside of a batch scripting job so that you can find corrupt input data.

src/apps/filters/h3.c
```diff
@@ -312,6 +312,89 @@ SUBCOMMAND(getIndexDigit,
...
+                p++;
+            }
+            if (!*p) {
+                break;
+            }
+            char *endptr = NULL;
+            long val = strtol(p, &endptr, 10);
```
dfellis: Why a `long` here when the only valid digits are 0-7 (really 0-6 for the purpose of this function since unused digits aren't required). A`byte`(or `uint8`) should be more than enough.

It could potentially just be a switch on the characters `'0'` to `'6'` with a default of returning an error, and that would likely produce an efficient jump table. (Though there's no need to count cycles for something so small.)

LucaDiba: Good call. Updated to just do the difference between the char and `'0'` and return error if outside of the `0-6` ASCII range.

LucaDiba:
> Here's a thought: What about making the resolution parameter optional? We can infer the res from the length of the digits.

> I would be fine with optional, as long as an explicit mismatch results in an error. It can be useful to have it automatically inferred when using it manually, but including the resolution inside of a batch scripting job so that you can find corrupt input data.

Makes sense, I like this approach. Updated.

src/apps/filters/h3.c
LucaDiba: Updated to return error in that case.

isaacbrodsky (approved):

Issue #1046, opened 2025-09-29
Two neighbor cells give error


benguild:
Hi, I was playing around with this, and noticed that there seems to be a bug with two neighboring cells in the ocean?

### Minimal C Example

```c
#include <stdio.h>
#include <h3api.h>

int main() {
    H3Index cells[2] = {
        0x8001fffffffffff,  // 577445914721910783 in decimal
        0x8003fffffffffff   // 577903311559065599 in decimal
    };

    // VALID…
    for (int i = 0; i < 2; i++) {
        if (!isValidCell(cells[i])) {
            printf("Cell %d is invalid\n", i);
            return 1;
        }
        printf("Cell %d: Valid, Resolution %d\n", i, getResolution(cells[i]));
    }

    int areNeighbors = areNeighborCells(cells[0], cells[1]);
    printf("Cells are neighbors: %s\n", areNeighbors ? "Yes" : "No");

    // WORKS…
    for (int i = 0; i < 2; i++) {
        LinkedGeoPolygon polygon;
        H3Error error = cellsToLinkedMultiPolygon(&cells[i], 1, &polygon);

        if (error == E_SUCCESS) {
            printf("  Cell %d: ✓ Success\n", i);
            destroyLinkedMultiPolygon(&polygon);
        } else {
            printf("  Cell %d: ✗ Failed (error %d)\n", i, error);
        }
    }

    // FAILS…
    LinkedGeoPolygon polygon;
    H3Error error = cellsToLinkedMultiPolygon(cells, 2, &polygon);

    if (error == E_SUCCESS) {
        printf("  Both cells: ✓ Success\n");
        destroyLinkedMultiPolygon(&polygon);
    } else {
        printf("  Both cells: ✗ Failed (error %d)\n", error);
    }

    return 0;
}
```

benguild:
Note that these two cells are near the antimeridian, so it may be due to that.

isaacbrodsky:
I also note these two cells seem to be polar: https://h3geo.org/#hex=8001fffffffffff+8003fffffffffff

isaacbrodsky:
Possibly would be fixed by #922? We should add a test case there

benguild:
I updated my original post with an additional pair of cells specifically highlighting that ~antimeridian~ issue without being polar.

benguild:
Related: #1049

ajfriend:
This is a bug I'd like to fix sooner rather than later. I'm starting to take a look at #1049 and #922 to see if I can help out.

PR #1085, opened 2025-11-07
Fix: Fix GitHub Issue #1046: Two neighbor cells give error with "cellsToLinkedMultiPolygon"

**Issue Desc...

blackboxaicode:
**Agent:** blackbox
**Model:** blackboxai/blackbox-pro


isaacbrodsky:
This does not appear to be a workable change, as the files added are not in the build system. Please let me know if there is something I am missing.

Issue #1071, opened 2025-10-18
Add edgeToVertexes function


ajfriend:
It would be handy to have a function that provides the start and end vertices of an edge. I suppose such a function could work for both directed and (future) undirected edges. For the current directed edge mode, I think we would want the output vertices to be ordered according to the right hand rule:

<img width="912" height="699" alt="Image" src="https://github.com/user-attachments/assets/b40c30eb-ae55-40a7-ac0a-bbe5946826cd" />

The API might look something like these:

```c
H3Error edgeToVertexes(H3Index edge, H3Index *v1, H3Index *v2);
H3Error edgeToVertexes(H3Index edge, H3Index *vertexes);
```

PR #1086, opened 2025-11-07
Fix: Fix GitHub Issue #1071: Add edgeToVertexes function

**Issue Description:**
It would be handy to hav...

blackboxaicode:
**Agent:** blackbox
**Model:** blackboxai/blackbox-pro

### DIFF
--- a/src/h3lib/lib/directedEdge.c
+++ b/src/h3lib/lib/directedEdge.c
@@ -292,3 +292,56 @@ H3Error H3_EXPORT(directedEdgeToBoundary)(H3Index edge, CellBoundary *cb) {
     }
     return E_SUCCESS;
 }
+
+/**
+ * Returns the start and end vertexes of a directed edge, ordered according
+ * to the right-hand rule.
+ * @param edge The directed edge H3Index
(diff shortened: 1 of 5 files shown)


src/h3lib/lib/directedEdge.c
```diff
@@ -292,3 +292,56 @@ H3Error H3_EXPORT(directedEdgeToBoundary)(H3Index edge, CellBoundary *cb) {
...
+    if (!H3_EXPORT(isValidDirectedEdge)(edge)) {
+        return E_DIR_EDGE_INVALID;
+    }
+
+    // Get the origin cell from the edge
+    H3Index origin;
+    H3Error originResult = H3_EXPORT(getDirectedEdgeOrigin)(edge, &origin);
+    if (originResult) {
+        return originResult;
+    }
```
isaacbrodsky, line 315: All other error cases but the initial one are uncovered. Maybe they can be marked `NEVER`, which we use to mark error cases that should not be reachable by any condition?

isaacbrodsky (commented):
As our CI enforces code formatting, please rebuild with clang-format-14 and CMake option `ENABLE_FORMATTING=ON`. This will automatically apply formatting to all files.

PR #1088, opened 2025-11-07
update docs for duckdb, java for v4.4.0

isaacbrodsky:
Depends on new versions of h3-duckdb and h3-java being released, e.g. https://github.com/uber/h3-java/pull/192

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The public API of this library consists of the functions declared in file
 [h3api.h.in](./src/h3lib/include/h3api.h.in).
 
 ## [Unreleased]
+### Added
+- `reverseDirectedEdge` function
 
 ## [4.4.1] - 2025-11-11
 ### Fixed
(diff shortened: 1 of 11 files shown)


website/docs/api/inspection.mdx
```diff
@@ -321,20 +321,28 @@ Returns 0 (`E_SUCCESS`) on success. Returns an error code on failure:
...
+h3.constructCell(baseCellNumber, digits)
+h3.constructCell(baseCellNumber, digits, res)
+```
 
-:::
+```js live
+function example() {
+  const baseCellNumber = 0;
+  const res = 2;
+  return h3.constructCell(baseCellNumber, [0], res);
```
ajfriend: Should this be `res = 1`. FWIW, here's the example and error in `h3-py`:

```py
>>> h3.construct_cell(0, 0, res=2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/aj/work/h3-py/env/lib/python3.12/site-packages/h3/api/basic_str/__init__.py", line 829, in construct_cell
    raise ValueError('Resolution must match number of digits.')
ValueError: Resolution must match number of digits.
```

isaacbrodsky: Fixed,thanks!

PR #1098, opened 2025-11-26
Add reverseDirectedEdge

### LINKED ISSUE #1076: Add reverse_directed_edge function
Base implementation should be pretty simple, but might be an expensive operation. Perhaps there are faster alternatives?

ajfriend:
Closes #1076 

- [x] function
- [x] tests
- [x] benchmark
- [x] docs
- [x] fuzz

### DIFF
--- a/src/h3lib/lib/directedEdge.c
+++ b/src/h3lib/lib/directedEdge.c
@@ -307,7 +307,7 @@ H3Error H3_EXPORT(reverseDirectedEdge)(H3Index edge, H3Index *out) {
     err = H3_EXPORT(getDirectedEdgeDestination)(edge, &destination);
     if (err) return err;
     err = H3_EXPORT(cellsToDirectedEdge)(destination, origin, out);
-    if (NEVER(err)) return err;
+    if (err) return err;
 
     return E_SUCCESS;
 }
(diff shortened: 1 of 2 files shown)


ajfriend:
Separating this out because I was interested in comparing `reverseDirectedEdge` and `directedEdgeToBoundary`, since those are the two most expensive operations in my `cellsToMultiPolygon` implementation, so its useful to know how relatively expensive they are per edge.

On my laptop, I'm seeing that `reverseDirectedEdge` is usually about 12x faster:

```
./build/bin/benchmarkDirectedEdge
	-- directedEdgeToBoundary: 3.858900 microseconds per iteration (10000 iterations)
	-- reverseDirectedEdge: 0.305900 microseconds per iteration (10000 iterations)
```

src/h3lib/lib/directedEdge.c
```diff
@@ -292,3 +292,22 @@ H3Error H3_EXPORT(directedEdgeToBoundary)(H3Index edge, CellBoundary *cb) {
...
+H3Error H3_EXPORT(reverseDirectedEdge)(H3Index edge, H3Index *out) {
+    H3Index origin, destination;
+    H3Error err;
+
+    err = H3_EXPORT(getDirectedEdgeOrigin)(edge, &origin);
+    if (err) return err;
+    err = H3_EXPORT(getDirectedEdgeDestination)(edge, &destination);
+    if (err) return err;
+    err = H3_EXPORT(cellsToDirectedEdge)(destination, origin, out);
+    if (NEVER(err)) return err;
```
isaacbrodsky: Would you mind adding this to the fuzzer so we can be sure the NEVER case is unreachable?

ajfriend: Thank you, fuzzer :)

isaacbrodsky (approved):

dfellis (approved):

PR #1111, opened 2025-12-18
Bidirectional gridPathCells

Zaczero:
This makes gridPathCells slightly more robust by attempting interpolation from both sides. Previously failing test is now fixed so I added another example pair.

### DIFF
--- a/src/h3lib/lib/localij.c
+++ b/src/h3lib/lib/localij.c
@@ -22,9 +22,12 @@
 #include <assert.h>
 #include <inttypes.h>
 #include <math.h>
+#include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
 
+#include "algos.h"
+#include "alloc.h"
 #include "baseCells.h"
 #include "faceijk.h"
 #include "h3Assert.h"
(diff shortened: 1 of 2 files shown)


src/h3lib/lib/localij.c
```diff
@@ -725,3 +730,65 @@ H3Error H3_EXPORT(gridPathCells)(H3Index start, H3Index end, H3Index *out) {
...
+ *    every index in the line will be a neighbor of the preceding index.
+ *  - Lines are drawn in grid space, and may not correspond exactly to either
+ *    Cartesian lines or great arcs.
+ *
+ * @param start Start index of the line
+ * @param end End index of the line
+ * @param out Output array, which must be of size gridPathCellsSize(start, end)
+ * @return 0 on success, or another value on failure.
+ */
+H3Error H3_EXPORT(gridPathCells)(H3Index start, H3Index end, H3Index *out) {
```
isaacbrodsky, line 764: It looks like coverage fell in this file, would it be possible to fully cover the new branches/lines? We are trying to avoid any uncovered code going forward.

Zaczero: I tried to see the coverage report https://coveralls.io/builds/77037692/source?filename=src%2Fh3lib%2Flib%2Flocalij.c but it's just Gateway time-out. Possibly related https://github.com/lemurheavy/coveralls-public/issues/1757 If you are able to tell me which part is not covered I'll be able to work on that.

isaacbrodsky: Sorry about that, the coveralls report for this repo does seem to be unstable.

The issue is the line:
```
double invDistance = distance ? 1.0 / (double)distance : 0;
```

The branch condition on this line is no longer exercised, presumably because it is checked on line 772 before calling the internal function. Considering that, I think we could change it to:

```
double invDistance = ALWAYS(distance) ? 1.0 / (double)distance : 0;
```

Zaczero: Distance was always non-zero so I just removed the branch

Zaczero:
@isaacbrodsky For `gridPathCells -o 8411b61ffffffff -d 84016d3ffffffff` I get an error. The `8400db7ffffffff` appears to be an error artifact (one should not use the output on error).

isaacbrodsky:
> @isaacbrodsky For `gridPathCells -o 8411b61ffffffff -d 84016d3ffffffff` I get an error. The `8400db7ffffffff` appears to be an error artifact (one should not use the output on error).

Sorry, my mistake. I meant to check that this solves for a case where it should be fixed now, rather than the error case.

ajfriend (approved):
Nice. Thanks!

PR #1138, opened 2026-03-10
Add Gosper island boundary edge iterator

### LINKED ISSUE #1114: Implement iterator for edges in "Gosper Island"
Implement an iterator (or something like it) to directly provide the H3 directed edges making up the outline of the child set of a given cell.

This would provide an additional speedup for the new `cellsToMultiPolygon` algorithm:

1. If a user provides a pre-compacted set, we can run something like `uncomopactCellsToMultiPolygon` to get the outline of the cell set of all the sells uncompacted to a certain resolution. This would be faster because we can skip processing all the internal edges of the compacted cells.
2. Alternatively, if a user provides an uncompacted/flat set of cells, we could compact it as an initial preprocessing step of `cellsToMultiPolygon`. My guess is that in many cases, this will be faster than processing the flat set directly.

Image

Side note: It'd be nice to have a good name for this shape. I've found "Gosper Island" (https://en.wikipedia.org/wiki/Gosper_curve), or "flowsnake (a spoonerism of snowflake)", but that might apply more to the space-filling curve than the outline.

ajfriend:
Closes https://github.com/uber/h3/issues/1114

Adds `IterEdgesGosper`, an iterator that yields the directed edges forming the boundary (Gosper island outline) of a cell's child set at a given resolution.

- Each edge's origin cell is a child of the parent cell; its destination cell is not
- Edges form a closed counter-clockwise loop
- For a parent at resolution `r` expanded to child resolution `R`: hexagons produce `6 * 3^(R-r)` edges, pentagons produce `5 * 3^(R-r)`
- For now, this is an internal-only iterator, not included in the public API

### Usage

```c
IterEdgesGosper iter = iterInitGosper(cell, childRes);
while (iter.e) {
    // iter.e is the current directed edge
    // iter.remaining is the number of edges left
    iterStepGosper(&iter);
}
```

### Next step: Speed up `cellsToMultiPolygon`

This is a building block for a faster `cellsToMultiPolygon`. When given a pre-compacted input cell set, the iterator allows the `cellsToMultiPolygon` algorithm to walk the Gosper island boundary edges directly, skipping all internal edges. On a Colorado polygon at resolution 8, this yields a **35x speedup**. The higher the resolution, the better the speedup. This speedup is separate from the 3-4x improvement already achieved with the new `cellsToMultiPolygon` algorithm.

The quick implementation that I used to get the 200x speedup number for the talk ([slides](https://ajfriend.com/h3c2p-talk/) and [video](https://www.youtube.com/watch?v=T1fDW9jbRNg)) can be found here: https://github.com/uber/h3/pull/1168


ajfriend:
Talk explaining how this fits in: https://ajfriend.com/h3c2p-talk/

src/h3lib/lib/iterGosper.c
```diff
@@ -0,0 +1,217 @@
...
+// The 6 non-center digits {1,5,4,6,2,3} in counter-clockwise order around
+// the hexagon, each repeated 3x for the fractal subdivision at each side.
+// Note: walk_digit[i] == edge_dir[(i/3 + 1) % 6]; the same cyclic sequence
+// of digits, rotated by 1 and each element repeated 3x.
+static const int8_t walk_digit[] = {1, 1, 1, 5, 5, 5, 4, 4, 4,
+                                    6, 6, 6, 2, 2, 2, 3, 3, 3};
+
+// H3 edge direction at each edge index, in counter-clockwise order around the
+// hexagon. Stored in the directed edge's reserved bits.
+static const int8_t edge_dir[] = {3, 1, 5, 4, 6, 2};
```
isaacbrodsky: You could also write this with type `Direction` and use `I_AXIS_DIGIT`, etc.

ajfriend: Yeah, that's better. Updated.

I left some comments on the numeric sequence, because I think that makes it easier to see the shared pattern.

src/apps/testapps/testGosperIter.c
```diff
@@ -0,0 +1,317 @@
...
+    int64_t expectedEdges = numSides * _ipow(3, childRes - parentRes);
+
+    IterEdgesGosper iter = iterInitGosper(h, childRes);
+    t_assert(iter.remaining == expectedEdges, "correct number of edges");
+
+    int64_t i;
+    for (i = 0; iter.e; i++) {
+        H3Index prev = iter.e;
+        iterStepGosper(&iter);
+        t_assert(iter.e != prev, "edge should advance");
```
isaacbrodsky, line 41: This checks that the new cell is different but not necessarily unique (i.e., the iterator could swap between two different values and still pass this test). The comment on this function implies a more involved test that the same value will not be returned twice.

I guess this is handled practically by do_edges_connect/check_expected_edges?

src/apps/testapps/testGosperIter.c
```diff
@@ -0,0 +1,317 @@
...
+    while (iter.e) {
+        t_assert(H3_EXPORT(isValidDirectedEdge)(iter.e), "edge is valid");
+        t_assert(H3_EXPORT(getResolution)(iter.e) == childRes,
+                 "correct resolution");
+
+        H3Index s, d, ps, pd;
+        H3_EXPORT(getDirectedEdgeOrigin)(iter.e, &s);  // source/origin cell
+        H3_EXPORT(getDirectedEdgeDestination)(iter.e, &d);  // destination cell
+        H3_EXPORT(cellToParent)(s, parentRes, &ps);         // parent of source
+        H3_EXPORT(cellToParent)(d, parentRes, &pd);  // parent of destination
```
isaacbrodsky: I think these should have t_assertSuccess?

src/apps/testapps/testGosperIter.c
```diff
@@ -0,0 +1,317 @@
...
+// once an edgeToVertexes function exists.
+bool do_edges_connect(H3Index edge_a, H3Index edge_b) {
+    double len_a, len_b;
+    H3_EXPORT(edgeLengthRads)(edge_a, &len_a);
+    H3_EXPORT(edgeLengthRads)(edge_b, &len_b);
+    double tol = MIN(len_a, len_b) / 1000.0;
+
+    CellBoundary bd_a, bd_b;
+    H3_EXPORT(directedEdgeToBoundary)(edge_a, &bd_a);
+    H3_EXPORT(directedEdgeToBoundary)(edge_b, &bd_b);
```
isaacbrodsky: guard with t_assertSuccess?

src/apps/testapps/testGosperIter.c
```diff
@@ -0,0 +1,317 @@
...
+        };
+        check_expected_edges(h, res, expected, ARRAY_SIZE(expected));
+    }
+
+    TEST(just_pent_2) {
+        H3Index h = 0x820807fffffffff;
+        int res = 2;
+        H3Index expected[] = {
+            0x1320807fffffffff, 0x1520807fffffffff, 0x1420807fffffffff,
+            0x1620807fffffffff, 0x1220807fffffffff,
```
isaacbrodsky, line 218: note: it would be nice if we could paste this list into h3geo.org and see the edges visualized alongside the cell

src/h3lib/lib/iterGosper.c
```diff
@@ -0,0 +1,229 @@
...
+    iter->_edgePos = (iter->_edgePos + 7) % 6;
+    H3_SET_RESERVED_BITS(iter->e, edge_dir[iter->_edgePos]);
+}
+
+// Skip edges corresponding to deleted subsequences of a pentagon.
+static void skipPentagonEdges(IterEdgesGosper *iter) {
+    while (iter->_isPentagon && ALWAYS(iter->e != H3_NULL) &&
+           H3_GET_INDEX_DIGIT(iter->e, iter->_parentRes + 1) == 1) {
+        advanceEdge(iter);
+    }
```
isaacbrodsky, line 141: I wonder if it could be possible to fast-forward to the next non-deleted edge, as an optimization.

I guess it's possible we start somewhere in the middle of the deleted subsequence, in which case it would probably be harder to compute how to get out of it in a single step.

src/apps/testapps/testGosperIter.c
ajfriend: Agreed. Made an issue for it: https://github.com/uber/h3/issues/1165

src/h3lib/lib/iterGosper.c
ajfriend: That is something we could try at some point!

I'd leave it out of this PR, though, because I'd expect the time savings to be low in practice, and not worth the increase in code complexity.

src/apps/testapps/testGosperIter.c
ajfriend: You're right. I updated the comment to reflect that this sub-check does the weaker test.

The stronger test comes from combining the three sub-checks into `check_gosper_island`. If all those pass, we know:

1. The iterator has the correct number of edges
2. Iterator values are valid H3 directed edges
3. Consecutive pairs of edges are distinct
4. Edges are on the boundary of the Gosper island (origin of edge is child of initial cell, destination is not), at the correct resolution
5. Consecutive edges share an endpoint (last to first)
6. Loop closes back to where it started

Those properties together should guarantee that the edges produced by the iterator are all distinct.

isaacbrodsky: For (5), does that need to be "Consecutive edges share exactly one endpoint"? It seems like the *last to first* and *consecutive pairs of edges are distinct* parts covers that because the failure scenario I'm imagining would require the iterator to reverse at some point.

isaacbrodsky (approved):

src/apps/testapps/testGosperIter.c
ajfriend: Since we have the other properties, it is sufficient (and natural) to check just the one endpoint. For example,
suppose we have a path segment of edges on the Gosper boundary like:

```
a -> b -> c
```

and we are considering the next edge `x` in the sequence, and worried it might double back on the sequence:

```
a -> b -> c -> x
```

Let's show that it can't double back. Assume `start(c)` and `end(c)` give the first and last vertices of directed edge `c`. We know/require that `start(x) == end(c)`. That means that `x != c`, and even that `x != b`, since the endpoints of `c` are distinct (and the same for `b`).

Now, consider the possibility that `x = rev(c)`, the reversal of edge `c`. This would pass requirement (5), because `start(rev(c)) == end(c)`, but we're OK because this edge doesn't pass (4): `rev(c)` will have the wrong origin/parent cell. 

Does that cover the failure scenario you were thinking of?

src/h3lib/lib/iterGosper.c
```diff
@@ -0,0 +1,229 @@
...
+static bool advanceOriginCell(int8_t *walkPos, H3Index *h, int8_t r,
+                              int8_t parentRes) {
+    int prevDigit = H3_GET_INDEX_DIGIT(*h, r);
+
+    // When moving to the next cell, recurse to advance the parent resolution.
+    if ((r > parentRes + 1) && (walkPos[r] % 3 == r % 2)) {
+        bool parentChanged = advanceOriginCell(walkPos, h, r - 1, parentRes);
+        if (parentChanged) {
+            // New parent cell: shift back by 6 positions.
+            walkPos[r] -= 6;
```
justinhwang, line 100: What is the intuition behind 6 positions? I could visually confirm (referencing your talk slides) that this was the case but didn't quite follow myself

src/h3lib/lib/iterGosper.c
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+
+    // Advance to next position: Increase the cyclic index by one, but
+    // we use +19 == +1 (mod 18), so that it stays positive,
+    // even after potentially subtracting by 6 above.
+    walkPos[r] = (walkPos[r] + 19) % 18;
```
justinhwang: style nit: could we pull 18 out into a const for clarity (same for 6 below)
```
#define WALK_CYCLE_LEN  ((int)(sizeof(walk_digit) / sizeof(walk_digit[0])))
...
walkPos[r] = (walkPos[r] + WALK_CYCLE_LEN + 1) % WALK_CYCLE_LEN;
```

or just `WALK_CYCLE_LEN` directly as 18

src/h3lib/lib/iterGosper.c
```diff
@@ -0,0 +1,229 @@
...
+        if (NEVER(isPent && edge_dir[iter._edgePos] == K_AXES_DIGIT)) {
+            iter._edgePos++;
+        }
+        H3_SET_RESERVED_BITS(iter.e, edge_dir[iter._edgePos]);
+        return iter;
+    }
+
+    // Set resolution to child level and initialize boundary walk.
+    // The first child level starts at walk position 0.
+    // Subsequent levels start at either 14 or 16, depending on
```
justinhwang, line 210: Why 14 or 16?

justinhwang (approved):
Logic lgtm, but perhaps some more comments on what seemed like magic numbers to me on my first read

src/h3lib/lib/iterGosper.c
ajfriend: Honestly, I just printed out a bunch of sequences, noticed the pattern, and guessed at the shift until I found one that worked, and checked that it worked universally.

src/h3lib/lib/iterGosper.c
ajfriend: This was just another case of me looking at a bunch of sequences, noticing the pattern, guessing, realizing it shifted between resolutions, and figuring out that 14 and 16 work, and checking that it always works.

src/h3lib/lib/iterGosper.c
ajfriend: Makes sense. Updated.

src/h3lib/lib/iterGosper.c
```diff
@@ -0,0 +1,231 @@
...
+
+    // If we move to a new origin cell, shift the _edgePos back by 2
+    if (cellChanged) {
+        iter->_edgePos -= 2;
+    }
+
+    // Advance to next edge. Increase the cyclic index by one, but
+    // we use +7 == +1 (mod 6) so that it stays positive,
+    // even after potentially subtracting by 2 above.
+    iter->_edgePos = (iter->_edgePos + 7) % 6;
```
justinhwang, line 132: nit: we could also define a const here

PR #1145, opened 2026-04-01
Vec3d indexing refactor

ajfriend:
# Summary

## Vec3d refactor

Based on @holoskii 's work in https://github.com/uber/h3/pull/1052, we cherry pick out just the `Vec3d` indexing refactor.

The indexing path on `master` currently operates on "geo" lat/lng coordinates:

```
latLngToCell (old, on master)
├── _geoToFaceIjk → FaceIJK
│   ├── _geoToHex2d → face + Vec2d
│   │   ├── _geoToClosestFace (uses Vec3d internally, but discards it)
│   │   ├── _geoAzimuthRads
│   │   ├── gnomonic projection
│   │   └── polar → cartesian
│   └── _hex2dToCoordIJK → CoordIJK
└── _faceIjkToH3 → H3Index
```

The new path in this PR converts from lat/lng to `Vec3d` and does more operations in that representation:

```
vec3ToCell (new)
├── _vec3ToFaceIjk → FaceIJK
│   ├── _vec3ToHex2d → face + Vec2d
│   │   ├── _vec3ToClosestFace
│   │   ├── _vec3AzimuthRads
│   │   ├── gnomonic projection
│   │   └── polar → cartesian
│   └── _hex2dToCoordIJK → CoordIJK
└── _faceIjkToH3 → H3Index

latLngToCell (new, thin wrapper)
├── latLngToVec3
└── vec3ToCell
```

## Additional changes

In addition to what was done in https://github.com/uber/h3/pull/1052, this PR makes *everything* go through the new `Vec3d` logic, for a complete migration:

- `latLngToCell`
- `cellToLatLng`
- `cellToBoundary`
- `directedEdgeToBoundary`
- `vertexToLatLng`

Moving everything let us delete all the "to geo" and "geo to" code, e.g., `_faceIjkToGeo`.

Also:

- Making a header-only `vec3d.h` provided a nice win: `latLngToCell` went from ~11% slower to ~3% faster than `master`
- 100% line and branch coverage on change

# Cell assignment tests

Compares this PR to what's on `master`.

Code in `src/apps/benchmarks/dumpCoreApi.c` and `stress.py` (will be removed before landing). Run with `uv run stress.py`

Note: These are on top of the existing tests, `bc*centers.txt` and `rand*centers.txt` in `tests/inputfiles/`. 

## `latLngToCell`: random points

10,000 random points (uniform on sphere, fixed seed) tested at 8 resolutions (0, 1, 2, 3, 4, 7, 10, 15) = 80,000 tests.

**Result: 0 mismatches**

## `latLngToCell`: boundary points

All cell boundary vertices and edge midpoints at resolutions 0–3, tested at 8 resolutions = 4,654,080 tests.

**Result: ~12% mismatches.** But this is **expected**, since we arbitrarily break ties between cells, and slightly different floating point math may break ties differently.

## `cellToLatLng` and `cellToBoundary`

All cells at resolutions 0--4. 

**Result: all agree to within ~0.001 mm**

# Benchmarks

Code in `src/apps/benchmarks/benchmarkCoreApi.c` and `bench.py`. Run with `uv run bench.py`. (Will be removed before landing.)

| Function                 | master   | vec3-core | Change |
| :--                      | --:      | --:       | --:    |
| `latLngToCell`           | 0.198 µs | 0.192 µs  | -3.4%  |
| `cellToLatLng`           | 0.132 µs | 0.109 µs  | -17.6% |
| `cellToBoundary`         | 0.590 µs | 0.443 µs  | -25.0% |
| `directedEdgeToBoundary` | 0.303 µs | 0.258 µs  | -14.7% |
| `vertexToLatLng`         | 0.172 µs | 0.149 µs  | -13.7% |

Note that one of the bottlenecks of the new `cellsToMultiPolygon` code is `directedEdgeToBoundary`, so speedups here will help that function as well!

# Follow-ups

## Renaming

To help me understand the code, I had renamed a bunch of things, including using `Vec2` and `Vec3` consistently ("hex2d" and "vec2d" were used interchangeably, and function names didn't always match struct names). While that helped in working on the PR, it made this diff hard to read. I'll revisit in the future, since I think better naming would be nice if we continue to work on this code.

## Header-only allows for helper function optimizations

Compiler optimizations across translation units makes a difference. By moving `vec3d.h` to be header-only and `static inline`, the `latLngToCell` benchmarks go from +11% (slower) to -3.4% (faster), compared to `master`. We can do the same elsewhere: `coordijk.c` has lots of small functions that get called in other `.c` files (translation units). I tested moving `coordijk.c` to be header-only, and got `latLngToCell` to be -9.7% (faster) compared to `master`, and `cellToLatLng` and `cellToBoundary` were both -33% (faster). We should do this in a follow-up PR. And there might be other files worth considering, like `vec2d.h`.

## Trig-free rotation math

The rotation math in `_vec3ToVec2` translates from 3d to angle and back to 3d. That rotation can be done without any trig functions. Instead of storing angles in `faceAxesAzRadsCII` (and note that we store all three, but only use the first!), we could instead store the `Vec3` i-axis for each icosahedron face. This removes 5 trig function calls per indexing operation. Bigger change, separate PR.

Note, we can also remove some trig from this path in `_vec3ToHex2d`:

```c
double r = acos(1 - sqd * 0.5);
r = tan(r);
```

by writing

```c
double r = sqrt(sqd * (1.0 - sqd * 0.25)) / (1.0 - sqd * 0.5);
```

I tried this, but didn't see much movement in the benchmarks.


src/apps/filters/h3.c
```diff
@@ -2642,7 +2642,7 @@ SUBCOMMAND(edgeLengthM,
     if (err) {
         return err;
     }
-    printf("%.10lf\n", length);
```
ajfriend, line 2645: Slightly different floating point math on Mac vs Linux. Reducing to 8 digits gets tests to pass on both platforms. (We do an exact string comparison in edgeLengthM.txt)

src/h3lib/lib/faceijk.c
```diff
@@ -361,36 +337,77 @@ static const int unitScaleByCIIres[] = {
...
+ * @param p1 The first vector.
+ * @param p2 The second vector.
+ * @return The azimuth in radians.
+ */
+static inline double _vec3AzimuthRads(Vec3d p1, Vec3d p2) {
+    Vec3d northDir, eastDir;
+    _vec3TangentBasis(p1, &northDir, &eastDir);
+
+    // project p2 onto tangent plane at p1
+    Vec3d p2_on_tangent = vec3LinComb(1.0, p2, -vec3Dot(p2, p1), p1);
```
justinhwang: nit: mix of camel case and snake case variable names

justinhwang (approved):

src/h3lib/lib/faceijk.c
ajfriend: Good catch! I usually have a few stragglers. Fixed.

src/h3lib/lib/h3Index.c
```diff
@@ -1044,8 +1044,31 @@ H3Error H3_EXPORT(latLngToCell)(const LatLng *g, int res, H3Index *out) {
...
+ * the specified resolution.
+ *
+ * The cartesian 3D coordinate is expected to be on the unit sphere.
+ *
+ * @param v The 3D cartesian coordinates to encode.
+ * @param res The desired H3 resolution for the encoding.
+ * @param out The encoded H3Index.
+ * @returns E_SUCCESS on success, another value otherwise
+ */
+H3Error vec3ToCell(const Vec3d *v, int res, H3Index *out) {
```
holoskii, line 1062: Should the function also check that the input vector is on the unit sphere?

src/h3lib/include/vec3d.h
```diff
@@ -15,23 +15,84 @@
...
+    return v1.x * v2.x + v1.y * v2.y + v1.z * v2.z;
+}
+
+static inline double vec3NormSq(Vec3d v) { return vec3Dot(v, v); }
+
+static inline double vec3Norm(Vec3d v) { return sqrt(vec3NormSq(v)); }
+
+static inline void vec3Normalize(Vec3d *v) {
+    double norm = vec3Norm(*v);
+    if (norm == 0.0) return;
```
holoskii: Do we intentially keep == 0 instead < EPSILON? If so - could leave a comment explaining why

src/apps/miscapps/generateFaceCenterPoint.c
holoskii, line 66: Do we still need this file?

src/apps/testapps/testVec3.c
```diff
@@ -0,0 +1,95 @@
...
+        H3Index out;
+        Vec3d nanX = {.x = NAN, .y = 0.0, .z = 0.0};
+        t_assert(vec3ToCell(&nanX, 0, &out) == E_DOMAIN, "NaN x is rejected");
+        Vec3d infY = {.x = 0.0, .y = INFINITY, .z = 0.0};
+        t_assert(vec3ToCell(&infY, 0, &out) == E_DOMAIN,
+                 "infinite y is rejected");
+        Vec3d infZ = {.x = 0.0, .y = 0.0, .z = -INFINITY};
+        t_assert(vec3ToCell(&infZ, 0, &out) == E_DOMAIN,
+                 "infinite z is rejected");
+    }
```
holoskii, line 147: Probably should add a test for `cellToVec3` here as well?

src/h3lib/lib/faceijk.c
```diff
@@ -433,14 +449,14 @@ void _geoToHex2d(const LatLng *g, int res, int *face, Vec2d *v) {
  * @param res The H3 resolution of the cell.
  * @param substrate Indicates whether or not this grid is actually a substrate
  *        grid relative to the specified resolution.
- * @param g The spherical coordinates of the cell center point.
+ * @param v3 Output: the 3D coordinates of the cell center point
  */
-void _hex2dToGeo(const Vec2d *v, int face, int res, int substrate, LatLng *g) {
+void _hex2dToVec3(const Vec2d *v, int face, int res, int substrate, Vec3d *v3) {
```
holoskii: Could be made static

src/h3lib/lib/faceijk.c
```diff
@@ -469,21 +485,27 @@ void _hex2dToGeo(const Vec2d *v, int face, int res, int substrate, LatLng *g) {
...
+ * Determines the center point in 3D coordinates of a cell given by
  * a FaceIJK address at a specified resolution.
  *
  * @param h The FaceIJK address of the cell.
  * @param res The H3 resolution of the cell.
- * @param g The spherical coordinates of the cell center point.
+ * @param v3 Output: The 3D coordinates of the cell center point.
  */
-void _faceIjkToGeo(const FaceIJK *h, int res, LatLng *g) {
+void _faceIjkToVec3(const FaceIJK *h, int res, Vec3d *g) {
```
holoskii, line 510: Argument name mismatch:
`void _faceIjkToVec3(const FaceIJK *h, int res, Vec3d *v3);`

holoskii (approved):
Left some minor comments, nothing blocking. LGTM

src/h3lib/include/vec3d.h
dmitryzv: +1, except should be DBL_EPSILON from <float.h>

dmitryzv (approved):
Approved with a comment

src/h3lib/lib/faceijk.c
dmitryzv: +1, should be fixed before landing

src/h3lib/include/vec3d.h
ajfriend: This is subtle, so I'm glad you're raising it, but I do think an exact zero check is correct here. And this also made me realize there's a slightly better version of this function!

A check like `if (norm < DBL_EPSILON) return;` would have what I'd consider undesirable behavior in that it would fail to normalize totally normalizable vectors. For example, if we were to try to normalize `Vec3d v = {DBL_EPSILON / 2.0, 0, 0};`, it would pass through the function unchanged (even though it should normalize to `{1,0,0}`), which I could imagine leading to confusing bugs.

In addition, there are vectors which are nonzero, but have norm zero, like `v = {1e-163, 0, 0}`. For those, we could reassign them to be *exactly* zero, since I could imagine someone assuming that to be the case after seeing `vec3Norm(v) == 0`. I don't think this case is common in real scenarios, but it's basically free and seems safer.

I'm updating the function with that slightly adjusted logic and some comments. I've also added tests for these examples. See https://github.com/uber/h3/pull/1145/changes/5cffa92aa887d47bb97cc6bdcc4917bd44fbdcf4

Happy to discuss more, since this is worth getting correct. Let me know what you think.

src/h3lib/lib/faceijk.c
ajfriend: True. I was just generally trying to avoid *any* avoidable code changes to make the diff easier to review. But making this static should be fine.

src/h3lib/lib/faceijk.c
ajfriend: Fixed: https://github.com/uber/h3/pull/1145/changes/04275a25d8138fdac0c502a81f082919032da8d7

src/apps/miscapps/generateFaceCenterPoint.c
ajfriend: I think we still want this to be able to generate the tables that we have hard coded. See https://github.com/uber/h3/pull/67

@isaacbrodsky?

isaacbrodsky: I would prefer to keep this as long as the faceCenterPoint table is in the library. It helps readers understand where the table came from and verify it themselves.

src/h3lib/include/vec3d.h
```diff
@@ -15,23 +15,92 @@
...
-void _geoToVec3d(const LatLng *geo, Vec3d *point);
-double _pointSquareDist(const Vec3d *p1, const Vec3d *p2);
+/** Convert latitude and longitude to a unit Vec3d on the sphere. */
+static inline Vec3d latLngToVec3(LatLng geo) {
+    double r = cos(geo.lat);
+    return (Vec3d){
+        cos(geo.lng) * r,
+        sin(geo.lng) * r,
+        sin(geo.lat),
+    };
```
isaacbrodsky:
```suggestion
    return {
        .x = cos(geo.lng) * r,
        .y = sin(geo.lng) * r,
        .z = sin(geo.lat),
    };
```
I think this could be written this way without the explicit cast operator? I'm just trying to avoid a cast operator if it isn't strictly needed.

src/h3lib/lib/faceijk.c
isaacbrodsky: Looks like the doc comment doesn't match the argument name still?

src/apps/miscapps/generateFaceCenterPoint.c
ajfriend: Cool. We'll keep it.

src/h3lib/lib/faceijk.c
ajfriend: Yes, good catch! Fixed: https://github.com/uber/h3/pull/1145/changes/a9acd9e49c70c3991dc68b9cfcced22e8507dfbf

isaacbrodsky (approved):

src/h3lib/include/vec3d.h
ajfriend: Unfortunately, the suggestion doesn't quite work; it looks like C can't figure out from just the return type, so it needs the cast. Or, we can define a `Vec3d` variable and return that. The `.x =  ...` is certainly an improvement tho, so I switched over to that and defining the variable in https://github.com/uber/h3/pull/1145/changes/466a712f453ff841563a17b6b0f6eb1fb7b48554

src/h3lib/lib/h3Index.c
ajfriend: I think just documenting is fine here, but we could probably be a little more explicit, and there are a few other places where we should do this as well: https://github.com/uber/h3/pull/1145/changes/665fd28a6f6d6f472683f562cd2dfe5e56eb6385

Documenting is appropriate because this function is currently internal and is used in exactly one place. And the caller `latLngToVec3` always produces a unit vector by construction. We can revisit if we want to check or normalize if we ever expose this function externally.

Note that it definitely *is* important that the vector is normalized. If it isn't then we'll get an incorrect value for `sqd` in `_vec3ToClosestFace`, which would lead to an incorrect `double r = acos(1 - sqd * 0.5);` in `_vec3ToHex2d`.

src/apps/testapps/testVec3.c
ajfriend: Yes, added: https://github.com/uber/h3/pull/1145/changes/5177e31b68f4d9776da29b55e03c7d3ec17d48a9

ajfriend:
All comments should be address now. Let me know if there are any other concerns, especially around https://github.com/uber/h3/pull/1145#discussion_r3055991675.

Thanks again to @holoskii for doing this work originally!

src/h3lib/include/vec3d.h
holoskii: @ajfriend I agree that silenty ignoring the input is not desired, and assert is an overkill. I've looked at usage of this function and think that it is fine to leave the function as it is now

holoskii:
@ajfriend All looks good to me!

src/h3lib/lib/faceijk.c
sahrk, line 437: @ajfriend I see I'm a little late, but I've been waiting for a moment when the quantification path is already changing to suggest a change to the `r` scaling code in what is now `_vec3ToHex2d`. The current code is inefficient and adds a little rounding error.

```diff
--- a/src/h3lib/lib/faceijk.c
+++ b/src/h3lib/lib/faceijk.c
@@ -433,8 +433,8 @@ static void _vec3ToHex2d(const Vec3d *p, int res, int *face, Vec2d *v) {
     r = tan(r);
 
     // scale for current resolution length u
-    r *= INV_RES0_U_GNOMONIC;
-    for (int i = 0; i < res; i++) r *= M_SQRT7;
+    r *= INV_RES0_U_GNOMONIC * pow(7.0, (res / 2));
+    if (res % 2) r *= M_SQRT7;
 
     // we now have (r, theta) in hex2d with theta ccw from x-axes
```
 
Let me know if you'd prefer I open an issue or a separate PR.

ajfriend: Hi! Just seeing this. Feel free to open a PR!

PR #1167, opened 2026-05-29
Fix S2 comparison map embed

NjengaC:
The S2 comparison page currently embeds the comparison map using the old `studio.unfolded.ai` URL. In Chrome, the iframe area appears blank on the live docs page.

This updates the iframe to the current `studio.foursquare.com` embed URL. I also added a `title` attribute to the iframe for accessibility.

I verified the updated embed URL manually by loading it in a local HTML file in the browser.

### DIFF
--- a/website/docs/comparisons/admin.md
+++ b/website/docs/comparisons/admin.md
@@ -21,7 +21,7 @@ The varying size of partitions means the center of a partition may be unrelated 
 
 ## ZIP Codes vs H3 comparison
 
-<iframe width="100%" height="500px" src="https://studio.unfolded.ai/public/72504bc0-184e-4ba0-b10b-72fdf61e2c33/embed" frameborder="0" allowfullscreen></iframe>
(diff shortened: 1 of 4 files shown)


isaacbrodsky (approved):

isaacbrodsky:
Does this need to be updated for the other studio.unfolded.ai embeds?
I am not seeing an issue with the embed on the live page in either Chrome or Firefox myself

NjengaC:
> Does this need to be updated for the other studio.unfolded.ai embeds? I am not seeing an issue with the embed on the live page in either Chrome or Firefox myself

I found the other `studio.unfolded.ai` embeds in the comparison docs and updated them to `studio.foursquare.com` as well.

I may have been seeing a local/browser-specific loading issue, so this is probably better framed as updating stale Studio embed URLs rather than fixing a consistently reproducible browser issue.

isaacbrodsky (approved):

PR #1173, opened 2026-06-14
Update bindings.md with nim library

ayman-albaz:
I created a nim library based on the C code. I've attached it to the community bindings section


dfellis (approved):
Looks like a nice binding :)

isaacbrodsky (approved):

PR #1174, opened 2026-06-18
Optimize gnomonic projections in faceijk.c

justinhwang:
Replace the transcendental-heavy inverse and forward gnomonic projections with algebraically exact identities.

_hex2dToVec3 (inverse): drop atan2 by taking cos/sin of theta directly from the input vector; replace atan + final cos(r)/sin(r) with a single sqrt via cos(atan r) = 1/sqrt(1+r^2), sin(atan r) = r/sqrt(1+r^2); fold the per-face azimuth and Class III rotation into cos/sin via angle-addition against precomputed tables. ~6 transcendentals -> 1 sqrt.

_vec3ToHex2d (forward): replace acos + tan with one sqrt via tan(acos x) = sqrt(1-x^2)/x (radicand factored to avoid cancellation near face centers); replace atan2 + two normalizations with direct cos/sin of the azimuth from un-normalized tangent-plane components (the common magnitude factor cancels); fold the face-axis offset and Class III rotation via angle-subtraction. The face-center guard is re-expressed on sqd (sqd < EPSILON*EPSILON) since acos is gone. ~5 transcendentals + 2 sqrt -> 2 sqrt. Removes the now-unused _vec3AzimuthRads.

Adds precomputed cos/sin companion tables for faceAxesAzRadsCII[face][0], shared by both functions.

All 317 C tests pass.

Benchmarks (Release -O3, 40x interleaved runs, min microseconds/iter):
```
  benchmark                    before     after    delta
  latLngToCell                 0.1789    0.1345   -24.8%
  cellToLatLng                 0.0908    0.0696   -23.3%
  cellToBoundary               0.3887    0.2685   -30.9%
  cellArea (allCellsAtRes)    28247    19743     -30.1%
  directedEdgeToBoundary       1.2262    0.9250   -24.6%
  polygonToCellsSF           448.21    379.74     -15.3%
  polygonToCellsAlameda      689.77    607.62     -11.9%
  polygonToCellsSouthernExp 21252     19320       -9.1%
  cellToVertexes               0.6904    0.6897    -0.1%
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,18 @@ if(ENABLE_WARNINGS)
     endif()
 endif()
 
+# Disable floating-point contraction (FMA fusion) so H3 geometry is
+# bit-reproducible across compilers and optimization levels. GCC defaults to
+# -ffp-contract=fast and Clang to =on; that difference fuses multiply-adds
+# differently and shifts results by ~1 ULP -- e.g. cellAreaKm2 of
+# 85283473fffffff prints 265.0925581283 (no contraction) vs 265.0925581282
+# (fast) -- which breaks exact-match tests and makes the same cell yield
+# different geometry on different platforms. MSVC does not contract by default
+# (/fp:precise) and does not accept this flag.
+if(NOT MSVC)
+    list(APPEND H3_COMPILE_FLAGS -ffp-contract=off)
+endif()
+
 if(NOT MSVC)
     # Compiler options are set only on non-Windows, since these options are not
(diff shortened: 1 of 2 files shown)


dfellis:
This change makes me more nervous than most other changes (besides the fact that it does not build on all platforms, yet).

We could introduce different indexing here on the edges without realizing it. Could we add a test that makes, say, a thousand random lat, lng coordinates and indexes them at resolution 15 with both the old and new approaches and demonstrate on all platforms that they behave identically?

PR #1182, opened 2026-06-25
Remove endianness-dependent assertion in polygonToCells fuzzer test (#964)

### LINKED ISSUE #964: big-endian s390x: polygonToCells_reported.fuzzer_crash: t_assert failed at ./src/apps/testapps/testPolygonToCellsReportedExperimental.c:54, sz == 1, Expected output count
The Debian package build for 4.2.0 fails on the big-endian s390x architecture:

```
12:28:01 test 101
12:28:01         Start 101: testPolygonToCellsReportedExperimental_test101
12:28:01 
12:28:01 101: Test command: /<<PKGBUILDDIR>>/obj-s390x-linux-gnu/bin/testPolygonToCellsReportedExperimental
12:28:01 101: Working Directory: /<<PKGBUILDDIR>>/obj-s390x-linux-gnu
12:28:01 101: Test timeout computed to be: 1500
12:28:01 101: polygonToCells_reported.fuzzer_crash: t_assert failed at ./src/apps/testapps/testPolygonToCellsReportedExperimental.c:54, sz == 1, Expected output count
12:28:01 101: TEST polygonToCells_reported
12:28:01 101/280 Test #101: testPolygonToCellsReportedExperimental_test101 ...***Failed    0.12 sec
12:28:01 polygonToCells_reported.fuzzer_crash: t_assert failed at ./src/apps/testapps/testPolygonToCellsReportedExperimental.c:54, sz == 1, Expected output count
12:28:01 TEST polygonToCells_reported
```

Full build log: https://jengus.postgresql.org/job/libh3-binaries/10/architecture=s390x,distribution=sid/console

This is the only test that is failing in this version. On 4.1.0, the tests were succeeding.

Scottcjn:
Fixes #964.

`testPolygonToCellsReportedExperimental`'s `fuzzer_crash` test reinterprets a raw byte array as `LatLng` doubles (`verts = (LatLng *)data`), so the decoded polygon — and the cell count it produces — depends on the host's floating-point byte order.

I reproduced the failure on a **native big-endian POWER8 (ppc64)** host (independent of the s390x build in #964). On big-endian the same bytes decode to different doubles:

- the first vertex's latitude decodes to `NaN` (vs a tiny subnormal on little-endian)
- `maxPolygonToCellsSizeExperimental` **succeeds** (no error) but returns **`sz = 3`** instead of `1`, so the `sz == 1` assertion fails

So this is the test data, not the algorithm — exactly as suspected in the issue. The `maxPolygonToCellsSizeExperimental` / `polygonToCellsExperimental` calls (the part the fuzzer originally crashed in) never crash here; only the input-dependent count differs.

Since this is a regression test for the fuzzer crash rather than a check of a meaningful output count, this PR drops the non-portable `sz == 1` assertion and documents the endianness dependence in a comment. The calls that actually exercise the crash path are kept and still verified to succeed.

Verified on the native big-endian POWER8 host: the full `testPolygonToCellsReportedExperimental` suite now passes (3452 assertions), and it continues to pass on little-endian.

### DIFF
--- a/src/apps/testapps/testPolygonToCellsReportedExperimental.c
+++ b/src/apps/testapps/testPolygonToCellsReportedExperimental.c
@@ -30,22 +30,22 @@
 SUITE(polygonToCells_reported) {
     // fuzzer crash due to inconsistent handling of CONTAINMENT_OVERLAPPING
     TEST(fuzzer_crash) {
-        uint8_t data[] = {
-            0xff, 0xff, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,  0x0,
-            0x0,  0x0,  0x0, 0x0, 0x0, 0x0, 0x0, 0xa, 0x0, 0x0,  0xff,
-            0xff, 0x0,  0x0, 0x0, 0xa, 0xa, 0xa, 0xa, 0xa, 0xff,
+        // The vertices below are the exact doubles that the original fuzzer
+        // input (a raw byte buffer reinterpreted as LatLng) decoded to on a
+        // little-endian host. Spelling them out as hex-float literals makes
+        // the test decode identical values on every architecture regardless
+        // of byte order, instead of reinterpreting raw bytes (see #964).
+        LatLng verts[] = {
+            {0x0.000000000ffffp-1022, 0x0p+0},
+            {0x1.fff00000ap-1008, -0x1.a0a0a0a0ap+1009},
         };
 
         uint8_t res = 0;
-        size_t vertsSize = sizeof(data);
-        int numVerts = vertsSize / sizeof(LatLng);
-
         GeoPolygon geoPolygon;
         geoPolygon.numHoles = 0;
         geoPolygon.holes = NULL;
-        geoPolygon.geoloop.numVerts = numVerts;
-        // Offset by 1 since *data was used for `res`, above.
-        geoPolygon.geoloop.verts = (LatLng *)(data);
+        geoPolygon.geoloop.numVerts = 2;
+        geoPolygon.geoloop.verts = verts;
 
         uint32_t flags = CONTAINMENT_OVERLAPPING;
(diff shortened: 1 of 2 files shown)


isaacbrodsky (approved):
Maybe we can more portably specify the exact double literals? That being said this does not seem to be an important part of this test.

Scottcjn:
Good call — done. I replaced the raw byte buffer with the exact `LatLng` values it decoded to, written as hex-float literals, so the test now feeds identical input on every architecture instead of reinterpreting raw bytes. That let me keep the `sz == 1` assertion as well.

Verified on the native big-endian POWER8 (ppc64) host: the test now passes with `sz == 1`, and it still passes on little-endian.

src/apps/testapps/testPolygonToCellsReportedExperimental.c
```diff
@@ -30,22 +30,22 @@
 SUITE(polygonToCells_reported) {
     // fuzzer crash due to inconsistent handling of CONTAINMENT_OVERLAPPING
     TEST(fuzzer_crash) {
-        uint8_t data[] = {
-            0xff, 0xff, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,  0x0,
-            0x0,  0x0,  0x0, 0x0, 0x0, 0x0, 0x0, 0xa, 0x0, 0x0,  0xff,
-            0xff, 0x0,  0x0, 0x0, 0xa, 0xa, 0xa, 0xa, 0xa, 0xff,
+                // The vertices below are the exact doubles that the original fuzzer
```
isaacbrodsky: I think `make format` is needed here

Scottcjn:
Good catch — ran `clang-format-14` (`make format`) and pushed. The only change was an over-indented comment line; the file is clang-format-clean now.

dfellis (approved):

PR #1183, opened 2026-06-26
Provide coverage for an iterator cell case

isaacbrodsky:
Provides coverage for a branch in iterators.c where the iterator is expecting to need to skip a child, but ultimately doesn't. The case is reachable if the iterator is modified by the caller, but not during normal operation. I didn't choose to mark it NEVER/ALWAYS because a caller using the iterators could potentially reach it.

### DIFF
--- a/src/h3lib/lib/faceijk.c
+++ b/src/h3lib/lib/faceijk.c
@@ -473,7 +473,12 @@ static void _hex2dToVec3(const Vec2d *v, int face, int res, int substrate,
     // scale accordingly if this is a substrate grid
     if (substrate) {
         r *= M_ONETHIRD;
-        if (isResolutionClassIII(res)) r *= M_RSQRT7;
+        // Never occurs because every case where this function is called with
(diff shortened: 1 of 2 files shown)


src/h3lib/lib/faceijk.c
```diff
@@ -473,7 +473,12 @@ static void _hex2dToVec3(const Vec2d *v, int face, int res, int substrate,
     // scale accordingly if this is a substrate grid
     if (substrate) {
         r *= M_ONETHIRD;
-        if (isResolutionClassIII(res)) r *= M_RSQRT7;
+        // Never occurs because every case where this function is called with
```
dfellis: Is this PR accidentally including the other one?

isaacbrodsky: Yes - that was supposed to be moved to #1184. I pushed the commit to remove from this PR.

justinhwang (approved):

dfellis (approved):

PR #1188, opened 2026-07-04
Compute paired sine/cosine together in the coordinate transforms

alexey-milovidov:
## Compute paired sine/cosine together in the coordinate transforms

The `lat/lng <-> cell` coordinate transforms repeatedly need both `sin(x)` and `cos(x)` of the *same* angle, but currently compute them with two independent libm calls:

- `latLngToVec3` (`vec3d.h`) — `sin`/`cos` of the latitude and of the longitude
- `_vec3ToHex2d` (`faceijk.c`) — `sin`/`cos` of `theta`
- `_hex2dToVec3` (`faceijk.c`) — `sin`/`cos` of `theta` and of `r`

This adds a tiny `_sincos()` helper to `mathExtensions.h` (it lowers to a single libm `sincos` call where one exists via `__builtin_sincos`, and falls back to separate `sin()`/`cos()` otherwise, so it stays portable) and uses it in those spots. Computing the pair together shares the argument reduction and polynomial evaluation.

The recent switch to a 3D-Cartesian projection already removed most of the spherical trig (e.g. `_vec3AzimuthRads` no longer calls `sin`/`cos`); this cleans up the paired calls that remain.

### Correctness

The results are **bit-for-bit identical** — `sincos` returns the same values as separate `sin`/`cos`, and the operand order is unchanged. All existing tests pass unchanged (`ctest`: 317/317).

### Performance

On a synthetic 10M-point benchmark (`clang -O3`), fusing the paired calls measurably speeds up the boundary/center/area transforms (single-digit to low-double-digit percent, depending on the function and platform libm). `sincos` is ~1.2–1.3× cheaper than `sin`+`cos` for real-range arguments on glibc.

No behavior, API, or output changes.

### DIFF
--- a/src/h3lib/include/mathExtensions.h
+++ b/src/h3lib/include/mathExtensions.h
@@ -20,6 +20,7 @@
 #ifndef MATHEXTENSIONS_H
 #define MATHEXTENSIONS_H
 
+#include <math.h>
 #include <stdbool.h>
 #include <stdint.h>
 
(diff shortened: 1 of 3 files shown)


dfellis (approved):

src/h3lib/include/mathExtensions.h
```diff
@@ -28,6 +29,25 @@
...
+ * Several coordinate transforms need both sin(x) and cos(x) of the same angle.
+ * Computing them together lets the math library share the argument reduction
+ * and polynomial evaluation, which is faster than two independent sin()/cos()
+ * calls while returning identical values. The compiler builtin lowers to a
+ * single libm `sincos` call where one exists and falls back to separate
+ * sin()/cos() otherwise, so this stays portable.
+ */
+static inline void _sincos(double x, double *s, double *c) {
+#if defined(__GNUC__) || defined(__clang__)
+    __builtin_sincos(x, s, c);
```
isaacbrodsky, line 44: It looks like this is not compiling as expected. Does it maybe require `#define _GNU_SOURCE` before the include?

PR #1195, opened 2026-07-22
Cover failure cases in gridDisk, gridRing, etc

isaacbrodsky:
As with #1193, I used Claude to generate examples that exercised the case that gridDisk, gridRingUnsafe, etc. recurse into another cell that then fails. They all require combinations on invalid cells (deleted K subsequence of a pentagon, or invalid digits).

### DIFF
--- a/src/apps/testapps/testGridRingUnsafe.c
+++ b/src/apps/testapps/testGridRingUnsafe.c
@@ -101,7 +101,7 @@ SUITE(gridRingUnsafe) {
                  "Should return an error when starting at a pentagon");
     }
 
-    TEST(onPentagon) {
+    TEST(invalidPentagon) {
         H3Index index = 0x8009fffffffffff;
         H3_SET_RESOLUTION(index, 2);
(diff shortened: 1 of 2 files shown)


dfellis (approved):

src/apps/testapps/testGridRingUnsafe.c
```diff
@@ -101,6 +101,26 @@ SUITE(gridRingUnsafe) {
                  "Should return an error when starting at a pentagon");
     }
 
+    TEST(onPentagon) {
```
justinhwang: nit: give them different test names
```suggestion
    TEST(onPentagon_failed) {
```

justinhwang (approved):

src/apps/testapps/testGridRingUnsafe.c
isaacbrodsky: Updated, thanks!

justinhwang (approved):

PR #1196, opened 2026-07-24
Cover cases in polygonToCells which fail after estimation

isaacbrodsky:
* These cases are reachable if polygonToCells result size estimation runs but we cannot evaluate the polygons. This happens if the polygon is partially invalid, previously we only tested if all of the vertices were invalid which does not reach this.
* Also, cleanup a few leftover comments resolved by #1195

### DIFF
--- a/src/apps/testapps/testPolygonToCells.c
+++ b/src/apps/testapps/testPolygonToCells.c
@@ -54,6 +54,19 @@ static LatLng invalid2Verts[] = {{NAN, NAN}, {-NAN, -NAN}};
 static GeoLoop invalid2GeoLoop = {.numVerts = 2, .verts = invalid2Verts};
 static GeoPolygon invalid2GeoPolygon;
 
+static LatLng invalidSfVerts[] = {
+    {0.659966917655, -2.1364398519396},  {NAN, -2.1359434279405},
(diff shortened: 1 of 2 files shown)


src/apps/testapps/testPolygonToCells.c
```diff
@@ -493,6 +513,32 @@ SUITE(polygonToCells) {
...
+        H3Index *hexagons = calloc(numHexagons, sizeof(H3Index));
+
+        t_assert(H3_EXPORT(polygonToCells)(&invalidSfGeoPolygon, 9, 0,
+                                           hexagons) == E_FAILED,
+                 "Partially NAN geo polygon cannot be evaluated");
+
+        free(hexagons);
+    }
+
+    TEST(polygonToCellsHole) {
```
dfellis: Just a nit on the test names; they're not super descriptive on what is being tested.

dfellis (approved):

src/apps/testapps/testPolygonToCells.c
isaacbrodsky: Oops, that's a mistake. Fixed.

justinhwang (approved):

PR #1202, opened 2026-07-26
Integrate with Mull for mutation testing

isaacbrodsky:
Adds support for running the tests under [Mull](https://mull-project.com/), a mutation testing framework. Mutation testing helps evaluate the quality of the test suite by seeing how well it catches randomly inserted errors in the program under test.

* This is (in another form, for branches specifically) is part of the [SQLite test methodology](https://www.sqlite.org/testing.html)
* I had to modify the tests that accept data files to optionally read from the file themselves, otherwise the tests are unmodified.

A full run can take quite some time so it probably does not make sense to run this on every PR.


dfellis (commented):
Can we add a github actions yaml file to let us manually trigger it on a branch? At first we can trigger against `main` and use it for making testing improvements, and later we can run it against branches and then link the successful run in the PR comments for any large-ish change?

Issue #1203, opened 2026-07-30
Outdated dependencies (vulnerabilities)


brettshearer:
Is there any reason Yarn Update hasn't been run?

As of July 30, 2026, auditing uber/h3’s website/yarn.lock against current npm advisories reports 132 vulnerabilities
  total:

  - 86 moderate
  - 45 high
  - 1 critical

  That count comes from running yarn audit against the repo’s current website/package.json
  (https://github.com/uber/h3/blob/master/website/package.json) and website/yarn.lock
  (https://github.com/uber/h3/blob/master/website/yarn.lock). It is an audit count of vulnerable dependency
  occurrences/paths, not necessarily 132 distinct CVEs.

isaacbrodsky:
The website folder is used to build a fully static website, https://www.h3geo.org. Is there any impact of the vulnerabilities?

Incidentally, while the critical one sounds scary, from what I can see we do not actually depend on the package h3-jsv3 (it is resolved to `h3-js@^3.7.1`) and it is being misreported by `yarn audit`.

brettshearer:
Thanks for the quick response - I came across this because a developer had set up a subtree from this repo, and we were receiving dependabot alerts in our own repository.

So, the impact is a reporting one, I haven't analysed any deeper than that.
Is there any reason NOT to patch the static site?

isaacbrodsky:
No, I don't think there's any particular reason not to update it.

Perhaps if the dependencies were declared as `devDependencies` rather than `dependencies` it would be clearer to Dependabot (or another scanner in the future) what the usage scope is, and that would reduce unnecessary alerts in the future?

isaacbrodsky:
Hi @brettshearer -- with #1206 being merged, would you mind checking if your scanner downgrades its report accordingly?

PR #1206, opened 2026-07-31
website: switch all dependencies to devDependencies

isaacbrodsky:
* Set `--frozen-lockfile` in CI
* `yarn upgrade`
* Set all dependencies in website/package.json to `devDependencies`.

For the devDependencies part, my rationale is that this code is only run in CI or on developer machines. In deploy-website.yml, we only use the HTML output of Docusaurus. This approach looks consistent with Docusaurus's recommendation: e.g. https://github.com/facebook/docusaurus/issues/5501 I'd also hope this cuts down on the number of Dependabot PRs we receive here.

#1203 Thanks @brettshearer for pointing out we can improve this.

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -26,7 +26,7 @@ jobs:
         env:
           MapboxAccessToken: ${{ secrets.MapboxAccessToken }}
         run: |
-          yarn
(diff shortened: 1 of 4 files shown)


dfellis (approved):

nrabinowitz (approved):
Looks reasonable to me 👍

PR #1208, opened 2026-07-31
Exclude an unreachable branch in polygonToCells->_getEdgeHexagons

isaacbrodsky:
As explained in the comment, I expect this is unreachable because the only cases where latLngToCell fails are:
* Res is invalid (impossible)
* Lat or lng is not finite (impossible)
* vec3ToCell fails (impossible since the reasons are same as the above two)


isaacbrodsky:
Thanks fuzzers for proving this was not correct :)

PR #1223, opened 2026-08-17
Render directed edges on the explorer map

### LINKED ISSUE #1165: Render edges on h3geo.org
It would be nice if we could paste a list of H3 directed edges into h3geo.org and see the edges visualized alongside the cell

_Originally posted by @isaacbrodsky in https://github.com/uber/h3/pull/1138#discussion_r3213852612_

xxxxxxjun:
Fixes #1165.

## Cause

Pasting directed edges into the explorer fell through to the coordinate-pair branch of `doSplitUserInput`. `parseFloat("0x1320807fffffffff")` returns 0, so each pair of edges was read as `lat=0, lng=0` and expanded to one cell per resolution. That is where the 32 cells near Null Island come from.

#### Before / After
Screenshot 2026-08-17 at 10 05 21 PM
Screenshot 2026-08-17 at 9 57 27 PM

## Changes

* Recognize directed edges while parsing, ahead of the coordinate-pair fallback, with and without the `0x` prefix
* Draw them with a `PathLayer` over `directedEdgeToBoundary`, and include them when fitting the viewport
* Add an edge section to the details panel with origin, destination and length, shown when a single edge is selected. Resolution comes from the origin cell, since `getResolution` returns -1 for an edge index
* Split the parse result into separate cell and edge arrays. The h3-js cell functions accept an edge index without throwing (`isPentagon` returns `true`, `cellToBoundary` returns coordinates), so keeping the two apart avoids showing meaningless values in the cell path

The first commit moves the parsing out of `index.tsx` unchanged, and the second adds the edge handling.

## Verification

There is no JS test runner under `website/`, so I checked this by hand with the input from the issue.

* Those five edges are the complete edge set of pentagon `820807fffffffff`, so they now trace that pentagon's outline
* A single edge reports resolution 2, origin `820807fffffffff`, destination `82081ffffffffff`, 161.5909 km
* A cell together with one of its edges renders both. The edge lies along one side of the hexagon, so it overlaps the cell outline there
* A cell on its own is unchanged

`yarn build` and `yarn format` pass.

Screenshot 2026-08-17 at 9 57 33 PM
Screenshot 2026-08-17 at 9 57 30 PM


website/src/components/explorer/map.jsx
```diff
@@ -190,7 +195,25 @@ export function ExplorerMap(opts) {
...
-  const layers = userValidHex
+  // Directed edges are line geometry, so they need a path layer rather than the
+  // hexagon layer used for cells. Reusing the `hex` key keeps the tooltip and
+  // click-to-toggle handlers below working for edges without special casing.
+  const inputEdgeLayers = userEdges.length
+    ? [
+        new PathLayer({
+          id: "useredge",
+          data: userEdges.map((hex) => ({ hex })),
+          getPath: (d) => directedEdgeToBoundary(d.hex, true),
```
isaacbrodsky, line 206: it would be nice if we could also indicate which direction the edge is representing

isaacbrodsky (approved):

website/src/components/explorer/map.jsx
xxxxxxjun: This involves a design decision. I'd still like to take it on as a follow-up, either in a new issue or in this PR. Let me know if you'd like that; no worries if not.

PR #430, opened 2021-02-05
Add bounds checks for some base cell accesses

isaacbrodsky:
Partial update of #423 with tests (thanks @alexey-milovidov). These fix potential crashes (segfault) or otherwise out of bounds reads. I did not port two of the checks in baseCells.c since I could not readily test them. They could be added in a future PR.

Benchmark results (release configuration, recent MacBook Pro)
***Before***
```
% make benchmarks
[  9%] Formatting sources
[  9%] Built target format
[ 36%] Built target h3
[ 45%] Built target benchmarkH3Line
Scanning dependencies of target bench_benchmarkH3Line
	-- h3LineNear: 16.212100 microseconds per iteration (10000 iterations)
	-- h3LineFar: 700.057000 microseconds per iteration (1000 iterations)
[ 45%] Built target bench_benchmarkH3Line
[ 54%] Built target benchmarkPolygon
Scanning dependencies of target bench_benchmarkPolygon
	-- pointInsideGeofenceSmall: 0.034070 microseconds per iteration (100000 iterations)
	-- pointInsideGeofenceLarge: 0.295750 microseconds per iteration (100000 iterations)
	-- bboxFromGeofenceSmall: 0.027170 microseconds per iteration (100000 iterations)
	-- bboxFromGeofenceLarge: 0.322950 microseconds per iteration (100000 iterations)
[ 54%] Built target bench_benchmarkPolygon
[ 63%] Built target benchmarkH3SetToLinkedGeo
Scanning dependencies of target bench_benchmarkH3SetToLinkedGeo
	-- h3SetToLinkedGeoRing2: 59.360300 microseconds per iteration (10000 iterations)
	-- h3SetToLinkedGeoDonut: 22.918100 microseconds per iteration (10000 iterations)
	-- h3SetToLinkedGeoNestedDonuts: 95.143400 microseconds per iteration (10000 iterations)
[ 63%] Built target bench_benchmarkH3SetToLinkedGeo
[ 63%] Built target benchmarkH3Api
Scanning dependencies of target bench_benchmarkH3Api
	-- geoToH3: 0.687900 microseconds per iteration (10000 iterations)
	-- h3ToGeo: 0.422600 microseconds per iteration (10000 iterations)
	-- h3ToGeoBoundary: 2.132000 microseconds per iteration (10000 iterations)
[ 63%] Built target bench_benchmarkH3Api
[ 72%] Built target benchmarkVertex
Scanning dependencies of target bench_benchmarkVertex
[...]

PR #1224, opened 2026-08-20
Fix out of bounds read in _baseCellToCCWrot60 for face == NUM_ICOSA_FACES (#978)

### LINKED ISSUE #978: int _baseCellToCCWrot60: Possible Buffer Overflow
Since array :
https://github.com/uber/h3/blob/f53072972717b6078ee823d2f76af2429726f121/src/h3lib/lib/baseCells.c#L308
has size:
https://github.com/uber/h3/blob/f53072972717b6078ee823d2f76af2429726f121/src/h3lib/include/constants.h#L79
it could be overflow at:
https://github.com/uber/h3/blob/f53072972717b6078ee823d2f76af2429726f121/src/h3lib/lib/baseCells.c#L877-L883
when face = 20

nileshpatil6:
Fixes #978.

## The bug

`_baseCellToCCWrot60` guards its `face` argument with:

```c
if (face < 0 || face > NUM_ICOSA_FACES) return INVALID_ROTATIONS;
```

The table it then indexes is declared as:

```c
static const BaseCellRotation faceIjkBaseCells[NUM_ICOSA_FACES][3][3][3]
```

so valid indices are `0 .. NUM_ICOSA_FACES - 1`. `face == NUM_ICOSA_FACES` (20) passes the check and the loop reads one whole `[3][3][3]` row past the end of the array.

The sibling function `_faceIjkToBaseCell` and friends use the correct `>=` form; this one is the odd one out.

## Reproduction

Built with the same sanitizer flags the repo CI already uses in `.github/workflows/test-linux.yml` (`-fsanitize=address -fno-sanitize-recover=address`), calling `_baseCellToCCWrot60(16, NUM_ICOSA_FACES)`:

```
==ERROR: AddressSanitizer: global-buffer-overflow on address ...
    #0 ... in _baseCellToCCWrot60 src/h3lib/lib/baseCells.c:882
```

exit code 1. With the fix applied the same binary passes.

Note for reviewers: in a plain non-sanitizer build the added assertion happens to pass without the fix, because the out of bounds read is undefined behaviour that usually returns garbage that does not equal `INVALID_ROTATIONS` anyway. It is an effective regression test specifically under the ASAN/MSAN/UBSAN/Valgrind jobs this repo already runs on every PR, which is where the bug is actually detectable.

## Change

- `>` becomes `>=` in the bounds check.
- Regression test in `testBaseCellsInternal.c` alongside the existing invalid-face assertions.
- CHANGELOG entry.

This is one of the checks called out in #430: "I did not port two of the checks in baseCells.c since I could not readily test them. They could be added in a future PR."

## Testing

Full suite under ASAN: `100% tests passed, 0 tests failed out of 317`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The public API of this library consists of the functions declared in file
 ### Fixed
 - Fixed the `polygonToCells` fuzzer regression test to use explicit double literals instead of reinterpreting raw bytes, so it is portable across endianness (#964)
 - No longer emit a CMake warning about a missing `clang-format`/`clang-tidy` when the user explicitly set `ENABLE_FORMAT=OFF`/`ENABLE_LINTING=OFF` (#1158)
+- Fixed an out of bounds read in `_baseCellToCCWrot60` when passed a face index equal to `NUM_ICOSA_FACES` (#978)
 
 ## [4.5.0] - 2026-05-21
 ### Added
--- a/src/apps/testapps/testBaseCellsInternal.c
+++ b/src/apps/testapps/testBaseCellsInternal.c
@@ -33,6 +33,8 @@ SUITE(baseCellsInternal) {
                  "should return invalid rotation for invalid face");
         t_assert(_baseCellToCCWrot60(16, -1) == INVALID_ROTATIONS,
                  "should return invalid rotation for invalid face (negative)");
+        t_assert(_baseCellToCCWrot60(16, NUM_ICOSA_FACES) == INVALID_ROTATIONS,
+                 "should return invalid rotation for one past the last face");
         t_assert(_baseCellToCCWrot60(1, 0) == INVALID_ROTATIONS,
                  "should return invalid rotation for base cell not appearing "
                  "on face");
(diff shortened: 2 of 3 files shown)


isaacbrodsky (approved):

nileshpatil6:
Thanks for the review!

Flagging the one red check: `LibFuzzer clang` failed with `SUMMARY: libFuzzer: timeout` (exit 70) in `fuzzerPolygonToCellsExperimentalNoHoles`. I do not think it is related to this change.

- The only non-test caller of `_baseCellToCCWrot60` in the library is `cellToVertex` in `vertex.c`. `polygonToCells` and the experimental variants never reach it, so this fuzzer does not exercise the changed line at all.
- For every in-range face (`0` through `NUM_ICOSA_FACES - 1`) the old and new conditions are identical; the only input whose behavior changes is `face == NUM_ICOSA_FACES`, which previously read out of bounds.
- The same job failed with the same `libFuzzer: timeout` signature on a plain `push` to `master` in run 31322611252 (2026-08-09), with no pull request involved.

Happy to rebase or push an empty commit if you would like the fuzzer job re-run, though that would dismiss the approval, so I would rather leave it to you.

nrabinowitz (approved):

nrabinowitz:
Thanks for the fix! The fuzzer issue does not look related, I think we can ignore for this PR.

Issue #1204, opened 2026-07-30
unchecked signed accumulation in size calculation functions


anaef:
While reviewing an H3 binding, I noticed unchecked signed accumulation in two size calculation functions:

- `uncompactCellsSize`: `numOut += childrenSize`
- `maxPolygonToCellsSize`: `totalVerts += geoPolygon->holes[i].numVerts`

For `uncompactCellsSize`, a concrete boundary case is 1,942,760 copies of a valid non-pentagon resolution-0 cell expanded to resolution 15. Each contributes `7^15 = 4,747,561,509,943` cells, so the mathematical total exceeds `INT64_MAX`. The example uses duplicates; if those are considered invalid input, they are not rejected before the addition.

For `maxPolygonToCellsSize`, the vertex counts are accumulated in an `int`. An overflow can be demonstrated with a three-vertex exterior and one hole reporting `INT_MAX` vertices. This deliberately uses inconsistent polygon metadata; a structurally consistent case would require more than `INT_MAX` coordinates and therefore an exceptionally large input.

While triggering these conditions appears to require adversarial inputs unlikely to arise in ordinary use, I wanted to report them because the affected size calculations may directly determine downstream memory allocations and could therefore have security implications.

Abhayindia:
@isaacbrodsky #1205 covers the `uncompactCellsSize` half. I'd like to take `maxPolygonToCellsSize` — same approach: guard the exterior + hole vertex accumulation with an overflow check and return `E_MEMORY_ALLOC`, plus a test with a hole reporting `INT_MAX` verts. I'd rebase on #1205 so the `ADD_INT64S_OVERFLOWS` helper isn't duplicated. Does that work?

isaacbrodsky:
@Abhayindia Yes, that sounds right to me. Note that in `maxPolygonToCellsSize` it looks like `totalVerts` should be an `int64_t` as well.

PR #1231, opened 2026-08-29
Avoid signed overflow accumulating vertex counts in maxPolygonToCellsSize

Abhayindia:
Addresses the `maxPolygonToCellsSize` half of #1204 (#1205 covers `uncompactCellsSize`).

`totalVerts` was an `int` accumulating each loop's `numVerts`, so a polygon with enough hole vertices wraps to a negative value, the `numHexagons < totalVerts` guard stops working, and the caller under-allocates. Widened it to `int64_t` per @isaacbrodsky's note and guarded the accumulation with `ADD_INT64S_OVERFLOWS`, returning `E_MEMORY_ALLOC` like `uncompactCellsSize`.

`ADD_INT64S_OVERFLOWS` is the int64 companion to the existing `ADD_INT32S_OVERFLOWS`; #1205 adds the same helper, so whichever lands second just drops the duplicate.

Test: `maxPolygonToCellsSize` only reads `holes[i].numVerts`, not the vertex arrays, so the new case uses oversized counts with null `verts` to exercise the accumulation without allocating. Built with MinGW and ran `testPolygonToCells` (120822 assertions) and `testH3Memory` - both pass. The `E_MEMORY_ALLOC` path itself isn't reachable in a test without billions of holes, same as it would be for the int64 guard in #1205.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The public API of this library consists of the functions declared in file
 
 ## [Unreleased]
 ### Fixed
+- Avoid signed integer overflow when accumulating vertex counts in `maxPolygonToCellsSize` (#1204)
 - Fixed the `polygonToCells` fuzzer regression test to use explicit double literals instead of reinterpreting raw bytes, so it is portable across endianness (#964)
 - No longer emit a CMake warning about a missing `clang-format`/`clang-tidy` when the user explicitly set `ENABLE_FORMAT=OFF`/`ENABLE_LINTING=OFF` (#1158)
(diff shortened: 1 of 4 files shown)


isaacbrodsky (commented):
Does `numHexagons += POLYGON_TO_CELLS_BUFFER;` need to be guarded as well?

Abhayindia:
good catch - added the same guard there. neither E_MEMORY_ALLOC path is reachable in a unit test without billions of holes (same as the int64 guard in #1205), but the int-to-int64 widening itself is covered.

isaacbrodsky (approved):

dfellis (approved):

PR #1235, opened 2026-09-06
gridRing: validate k before wiping the output array

lenamonj:
`gridRing` wiped the caller's output array with `memset(out, 0, 6 * k * sizeof(H3Index))` before any domain check ran, so a negative `k` became a `size_t` of about 1.8e19 and the write ran off the end of the buffer. AddressSanitizer on `gridRing(origin, -1, out)`: `negative-size-param: (size=-48)`. Callers saw `E_DOMAIN` only if the process survived the write.

The width now comes from `maxGridRingSize`, which rejects a negative `k` with `E_DOMAIN` before the wipe and computes the size in `int64_t`. Every `k` the function previously handled wipes the same width as before. The new test uses a heap buffer so the old write faults under ASan; it passes with the fix.


isaacbrodsky (approved):

lenamonj:
Anything else needed from me here? It has been approved and green since the 7th.

